PR #230, opened 2021-11-16
FIX: Implement emoji picker

markvanlan:
Dependent on this PR being merged first - https://github.com/discourse/discourse/pull/14976

Also I could not get `autocomplete` popup to go up on mobile - currently its cut off going down. Setting bottom value seems to work well! I didn't see any cases where it _doesn't_ work.

Screen Shot 2021-11-16 at 1 13 47 PM

Screen Shot 2021-11-16 at 1 14 20 PM

### DIFF
--- a/assets/stylesheets/mobile/mobile.scss
+++ b/assets/stylesheets/mobile/mobile.scss
@@ -30,10 +30,6 @@
   .tc-composer-input {
     padding-right: 75px;
   }
-  .autocomplete {
-    bottom: 40px;
-    top: unset !important;
-  }
   .open-toolbar-btn {
     right: 50px;
   }
--- a/assets/stylesheets/mobile/mobile.scss
+++ b/assets/stylesheets/mobile/mobile.scss
@@ -30,6 +30,10 @@
   .tc-composer-input {
     padding-right: 75px;
   }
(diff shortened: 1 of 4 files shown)


assets/stylesheets/mobile/mobile.scss
```diff
@@ -30,6 +30,10 @@
   .tc-composer-input {
     padding-right: 75px;
   }
+  .autocomplete {
+    bottom: 40px;
+    top: unset !important;
```
eviltrout, line 35: This is always fishy to me. Is this necessary?

markvanlan: Yeah, it is fishy. `autocomplete` uses `top` to set the position. By overriding `top`, and using a correct `bottom` value, the autocomplete box goes to the correct spot. I just realized that we own autocomplete.js so maybe I can fix a bug in there. 😢

markvanlan: @eviltrout I found the fix! `treatAsTextarea` fixes the position. Removed the css.

eviltrout: awesome

eviltrout (approved):

PR #241, opened 2021-11-23
DEV: Change to use uppy processor by default

martin-brennan:
In core PR https://github.com/discourse/discourse/pull/15058
we are changing to use uppy by default in the composer, so
there is no need for the experimental composer uploader check.

**Do not merge before core PR.**

### DIFF
--- a/assets/javascripts/discourse/components/chat-composer.js
+++ b/assets/javascripts/discourse/components/chat-composer.js
@@ -1,6 +1,5 @@
 import Component from "@ember/component";
 import UppyMediaOptimization from "discourse/lib/uppy-media-optimization-plugin";
-import ComposerUpload from "discourse/mixins/composer-upload";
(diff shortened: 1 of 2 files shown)


nattsw:
It's a bit unfortunate, but this seems to be quite a big change to git only from the removal of one processor?

Also, technically speaking would removing this processor affect acceptance tests had they been written, or none? Not trying to make your life more difficult, but it would've been nice if I could read a test documenting the change to know what to look out for.

martin-brennan:
@nattsw sorry yeah the huge diff is such a pain, it's because I removed a line from the component declaration so prettier decided all the code needed to move up a line :facepalm: . Go to `https://github.com/discourse/discourse-chat/pull/241/files?w=1` for the proper changes. As for the acceptance tests, there are none for chat uploads. I will take a look into writing one today, this is not urgent to get merged this PR

nattsw (approved):

PR #242, opened 2021-11-24
FEATURE: Open chat links from category/tag show

markvanlan:
Screen Shot 2021-11-24 at 12 26 50 PM

Screen Shot 2021-11-24 at 2 57 38 PM

### DIFF
--- a/app/controllers/chat_channels_controller.rb
+++ b/app/controllers/chat_channels_controller.rb
@@ -76,4 +76,34 @@ class DiscourseChat::ChatChannelsController < DiscourseChat::ChatBaseController
       render_json_error(membership)
     end
   end
+
+  def for_tag
+    params.require(:tag_name)
+
+    tag = Tag.find_by(name: params[:tag_name])
+    raise Discourse::NotFound unless tag
+
+    render_channel_for_chatable(
+      ChatChannel.find_by(chatable: tag)
+    )
+  end
+
+  def for_category
+    params.require(:category_id)
+
+    render_channel_for_chatable(
+      ChatChannel.find_by(chatable_id: params[:category_id], chatable_type: "Category")
+    )
+  end
+
+  private
+
+  def render_channel_for_chatable(channel)
+    if channel
+      guardian.ensure_can_see_chat_channel!(channel)
+      render_serialized(channel, ChatChannelSerializer)
+    else
+      render json: { chat_channel: nil }
+    end
+  end
 end
--- a/assets/javascripts/discourse/components/category-tag-chat-link.js
+++ b/assets/javascripts/discourse/components/category-tag-chat-link.js
@@ -1,0 +1,53 @@
+import Component from "@ember/component";
+import discourseComputed from "discourse-common/utils/decorators";
+import { ajax } from "discourse/lib/ajax";
+import { action } from "@ember/object";
+import { schedule } from "@ember/runloop";
+
+export default Component.extend({
+  category: null,
+  tag: null,
+  show: false,
+
+  init() {
+    this._super(...arguments);
+
+    if (!this.currentUser.has_chat_enabled) {
+      return;
+    }
+
+    if (this.category?.custom_fields.has_chat_enabled) {
+      this._fetchChannel("category", this.category.id);
+    }
+
+    if (this.tag) {
+      this._fetchChannel("tag", this.tag.id);
+    }
+  },
+
+  _fetchChannel(type, id) {
+    ajax(`/chat/chat_channels/for_${type}/${id}`).then((response) => {
+      if (response.chat_channel && !this.isDestroying && !this.isDestroyed) {
+        this.setProperties({
+          show: true,
+          channel: response.chat_channel,
+        });
+        schedule("afterRender", () => {
+          this.appEvents.trigger("sidebar:recalculate-button-width");
+        });
+      }
+    }).catch;
+  },
+
+  @action
+  openChat() {
+    this.appEvents.trigger("chat:open-channel-for-chatable", this.channel);
+
+    if (this.element && !this.isDestroying && !this.isDestroyed) {
+      const button = this.element.querySelector(".category-tag-chat-link");
+      if (button) {
+        button.blur();
+      }
+    }
+  },
+});
--- a/assets/javascripts/discourse/components/category-tag-chat-link.js
+++ b/assets/javascripts/discourse/components/category-tag-chat-link.js
@@ -35,7 +35,7 @@ export default Component.extend({
           this.appEvents.trigger("sidebar:recalculate-button-width");
         });
       }
-    }).catch;
+    });
   },
 
   @action
--- a/assets/javascripts/discourse/components/category-tag-chat-link.js
+++ b/assets/javascripts/discourse/components/category-tag-chat-link.js
@@ -12,7 +12,7 @@ export default Component.extend({
   init() {
     this._super(...arguments);
 
-    if (!this.currentUser.has_chat_enabled) {
+    if (!this.currentUser?.has_chat_enabled) {
       return;
     }
 
--- a/assets/javascripts/discourse/components/category-tag-chat-link.js
+++ b/assets/javascripts/discourse/components/category-tag-chat-link.js
@@ -1,5 +1,4 @@
 import Component from "@ember/component";
-import discourseComputed from "discourse-common/utils/decorators";
 import { ajax } from "discourse/lib/ajax";
 import { action } from "@ember/object";
 import { schedule } from "@ember/runloop";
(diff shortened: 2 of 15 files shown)


assets/javascripts/discourse/components/category-tag-chat-link.js
```diff
@@ -0,0 +1,52 @@
...
+      if (response.chat_channel && !this.isDestroying && !this.isDestroyed) {
+        this.setProperties({
+          show: true,
+          channel: response.chat_channel,
+        });
+        schedule("afterRender", () => {
+          this.appEvents.trigger("sidebar:recalculate-button-width");
+        });
+      }
+    }).catch;
```
eviltrout, line 38: I think this must be a mistake. `.catch` wouldn't do anything.

app/controllers/chat_channels_controller.rb
```diff
@@ -76,4 +76,34 @@ def notification_settings
...
     end
   end
+
+  def for_tag
+    params.require(:tag_name)
+
+    tag = Tag.find_by(name: params[:tag_name])
+    raise Discourse::NotFound unless tag
+
+    render_channel_for_chatable(
```
martin-brennan, line 86: I think perhaps you need to add a check around tag permissions for your `ensure_can_see_chat_channel` guardian extension, like we do in core for the `Tag#show` endpoint?

https://github.com/discourse/discourse/blob/7ce927eaa84bfaa9e72bd2f371625f2f95f6c42e/app/controllers/tags_controller.rb#L344-L346

assets/javascripts/discourse/components/category-tag-chat-link.js
```diff
@@ -0,0 +1,52 @@
...
+import { ajax } from "discourse/lib/ajax";
+import { action } from "@ember/object";
+import { schedule } from "@ember/runloop";
+
+export default Component.extend({
+  category: null,
+  tag: null,
+  show: false,
+
+  init() {
```
martin-brennan, line 11: I have no idea what we are _actually_ moving towards, but I thought we were supposed to use `@on("init")` for init functions and other ember things like willDestroyElement now?

https://github.com/discourse/discourse/blob/7ce927eaa84bfaa9e72bd2f371625f2f95f6c42e/app/assets/javascripts/discourse/app/models/category.js#L17-L17

It lets you avoid doing `this._super`

assets/javascripts/discourse/components/category-tag-chat-link.js
```diff
@@ -0,0 +1,52 @@
...
+    this._super(...arguments);
+
+    if (!this.currentUser?.has_chat_enabled) {
+      return;
+    }
+
+    if (this.category?.custom_fields.has_chat_enabled) {
```
martin-brennan, line 18: Is it possible `custom_fields` will be null?

martin-brennan (changes requested):
Just a few minor things. I think at some point it would make sense to cache which tags and categories have chat channels on first load, or add it into the serializer for tag or category so you don't have to make the additional AJAX request every time you open a tag or category. For now I think it should be OK.

assets/javascripts/discourse/components/category-tag-chat-link.js
markvanlan: It's possible for `this.category` to be null, but not `custom_fields` as long as I know. Guess I could add another `?` if that helps

assets/javascripts/discourse/components/category-tag-chat-link.js
markvanlan: I would be surprised... that line linked is 4 years old and I've done this style for every PR of chat that @eviltrout has been reviewing. I'm thinking it's the reverse, actually.

app/controllers/chat_channels_controller.rb
markvanlan: Good call, moved some logic to the guardian that was only in the `ChatChannelFetcher`, to secure this 👍

markvanlan:
@martin-brennan I talked with Robin about this approach... It's not simple to serialize the chat channel with categories, but it's realllly not simple to serialize them with tags (as only the tag name is passed around in some cases). The `has_chat_enabled` custom field does stop extra requests from being send for categories which is at least nice. I think this is fine for now though.

app/controllers/chat_channels_controller.rb
```diff
@@ -76,4 +76,34 @@ def notification_settings
...
+      ChatChannel.find_by(chatable: tag)
```
tgxworld, line 87: I don't have full context but don't we need to specify the `chatable_type` of `"tag"` here?

app/controllers/chat_channels_controller.rb
```diff
@@ -76,4 +76,34 @@ def notification_settings
...
+  end
+
+  private
+
+  def render_channel_for_chatable(channel)
+    if channel
+      guardian.ensure_can_see_chat_channel!(channel)
+      render_serialized(channel, ChatChannelSerializer)
+    else
+      render json: { chat_channel: nil }
```
tgxworld, line 106: I'm curious why we choose to return `nil` here instead of raising a `Discourse::NotFound` error when `channel` is `nil`.

spec/requests/chat_channel_controller_spec.rb
```diff
@@ -291,4 +291,91 @@
...
+        sign_in(user)
+        get "/chat/chat_channels/for_tag/#{hidden_tag_name}.json"
+        expect(response.status).to eq(403)
+      end
+
+      it "returns the channel for admin" do
+        sign_in(admin)
+        get "/chat/chat_channels/for_tag/#{hidden_tag_name}.json"
+        expect(response.status).to eq(200)
+        expect(response.parsed_body["chat_channel"]).not_to be_nil
```
tgxworld, line 336: I think it'll be better to assert for the right `chat_channel` instead of just `not_to be_nil`. The response can be returning else in the future and this test will still pass.

spec/requests/chat_channel_controller_spec.rb
```diff
@@ -291,4 +291,91 @@
...
+      get "/chat/chat_channels/for_tag/#{tag_without_channel.name}.json"
+      expect(response.status).to eq(200)
+      expect(response.parsed_body["chat_channel"]).to be_nil
+    end
+
+    describe "hidden tag" do
+      let(:hidden_tag_name) { "hidden1" }
+      before do
+        create_hidden_tags([hidden_tag_name])
+        ChatChannel.create(chatable: Tag.find_by(name: hidden_tag_name))
```
tgxworld, line 323:
```suggestion
        ChatChannel.create!(chatable: Tag.find_by(name: hidden_tag_name))
```

spec/requests/chat_channel_controller_spec.rb
```diff
@@ -291,4 +291,91 @@
...
+    end
+
+    it "returns the categories's chat channel" do
+      sign_in(user)
+      get "/chat/chat_channels/for_category/#{category.id}.json"
+      expect(response.status).to eq(200)
+      expect(response.parsed_body["chat_channel"]["id"]).to eq(category_channel.id)
+    end
+
+    it "errors when user trys to access staff channel" do
```
tgxworld, line 361:
```suggestion
    it "errors when user tries to access staff channel" do
```

assets/javascripts/discourse/components/category-tag-chat-link.js
martin-brennan: So weird cause I swear Joffrey has told me the opposite. No worries I will just ask about it internally.

assets/javascripts/discourse/components/category-tag-chat-link.js
martin-brennan: Nah just leave it

app/controllers/chat_channels_controller.rb
martin-brennan: Sweet!

martin-brennan:
@markvanlan thats fine about the whole serialiser thing was just thinking out loud

app/controllers/chat_channels_controller.rb
markvanlan: No, it's a polymorphic relationship and it picks up chatable_type automagically

assets/javascripts/discourse/components/category-tag-chat-link.js
eviltrout: I will rarely reject a PR for using them but I prefer not using decorators for this. See our internal discussion.

app/controllers/chat_channels_controller.rb
markvanlan: No need to error if a category/tag doesn't have a chat_channel. I'd rather just return nil and handle it properly on the client.

tgxworld: I understand but I’ve always found it weird from an API perspective. A successful 200 response yet the payload does not contain what I requested for.

markvanlan: I don't agree; a nil chat channel is expected behavior, and I don't want to serve a 404 error on expected behavior

martin-brennan (approved):

PR #243, opened 2021-11-29
FIX: Navigate to full-page in mobile view

markvanlan:


assets/javascripts/discourse/components/topic-chat-float.js
```diff
@@ -129,7 +129,15 @@ export default Component.extend({
       ajax(`/chat/chat_channels/${channel.id}/follow`, { method: "POST" })
         .then(() => {
           this.chat.startTrackingChannel(channel);
-          this.switchChannel(channel);
+          if (this.site.mobileView) {
+            return this.router.transitionTo(
```
eviltrout, line 133: Could it make to move this logic into `this.switchChannel`?

eviltrout (approved):

assets/javascripts/discourse/components/topic-chat-float.js
markvanlan: yeah we probably should, good idea.

eviltrout (approved):

PR #244, opened 2021-11-29
Workaround for Safari scrollbar position

pmusaraj:
Safari/Webkit appears to have a bug with `flex-direction: column-reverse`, on first load, it doesn't update the location of the scrollbar. The scrollbar is stuck at the top of the element, even though the element contents is scrolled to the bottom. 

This workaround triggers a 1-pixel scroll, to force recalculating the position of the scrollbar.

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -139,7 +139,7 @@ export default Component.extend({
           }
           this.setMessageProps(data.chat_view);
           this.decorateMessages();
-          this.onScroll();
+          this._stickScrollToBottom();
         })
         .catch((err) => {
           throw err;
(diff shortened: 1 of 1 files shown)


assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -804,7 +808,6 @@ export default Component.extend({
 
   @action
   composerValueChanged(composerValue) {
-    this.reStickScrollIfNeeded();
```
pmusaraj, line 807: It looks to me like this recalculation is unnecessary on every keystroke. With the workaround above, it also causes lots of jitteriness...

markvanlan: The composer height can change on composer value change, and if the composer height changes we might need to scroll..

pmusaraj: I'm sure I am missing context, but why do we need to scroll?

markvanlan: previously, when the composer changed height it would cause the chat-live-pane to scroll... can't replicate right now if I comment out the line `this.reStickScrollIfNeeded();`, so maybe we are good to remove now

markvanlan (approved):

PR #252, opened 2021-12-01
FEATURE: ChatUploads table with visuals in composer (no markdown)

markvanlan:
Supporting core PR - https://github.com/discourse/discourse/pull/15153, frontend tests will fail until they are being run on top of main with that PR merged.

### Single upload example in widget

Screen Shot 2021-12-01 at 9 43 33 AM

### Uploads container horizontally scrolls when there is overflow

Screen Shot 2021-12-01 at 9 44 38 AM

### Editing a message will populate it's uploads in the composer so uploads can be edited

Screen Shot 2021-12-01 at 9 45 25 AM


CvX (commented):
Looks like tests need to be updated

markvanlan:
> Looks like tests need to be updated

@CvX Do you mean the failing tests? Those are b/c things break with core not having the related PR. Maybe you just meant updated with correct spelling :)

assets/javascripts/discourse/components/chat-composer-upload.js
```diff
@@ -0,0 +1,19 @@
+import Component from "@ember/component";
+import discourseComputed from "discourse-common/utils/decorators";
+import { IMAGES_EXTENSIONS_REGEX } from "discourse/lib/uploads";
+
+export default Component.extend({
+  IMAGE_TYPE: "image",
+
+  tagName: "",
```
pmusaraj, line 8: I would use `span.chat-upload` here, and remove the wrapping span from the `hbs` template.

assets/stylesheets/common/common.scss
```diff
@@ -1015,16 +1023,76 @@ $float-height: 530px;
...
+  white-space: nowrap;
+  overflow-x: auto;
 
-  .uploading-indicator {
-    width: calc(100% - 80px);
-  }
-  .cancel-btn {
-    width: 70px;
+  .chat-upload {
+    display: inline-block;
```
pmusaraj, line 1034: Duplicate `display:` property.

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -670,9 +673,12 @@ export default Component.extend({
   },
 
   @action
-  editMessage(chatMessage, newContent) {
+  editMessage(chatMessage, newContent, uploads) {
     this.set("sendingloading", true);
-    let data = { new_message: newContent };
+    let data = {
+      newMessage: newContent,
```
eviltrout, line 679: We should be using `snake_case` for parameters to ruby controllers, so `new_message` and `upload_ids` etc.

pmusaraj (approved):

assets/javascripts/discourse/components/chat-live-pane.js
markvanlan: Okay, I was avoiding this so I could reuse the same JS object for settings attrs, but I'll do that :)

PR #254, opened 2021-12-01
DEV: Use adapter for chat messages

markvanlan:


CvX (commented):
👍 but the specs need to be updated 😃

markvanlan:
> 👍 but the specs need to be updated 😃

@CvX okay you got me on this one :)

assets/javascripts/discourse/models/chat-message.js
```diff
@@ -0,0 +1,2 @@
+import RestModel from "discourse/models/rest";
+export default RestModel.extend({});
```
eviltrout, line 2: I'm surprised this is required - does it not create a record automatically?

eviltrout (approved):

PR #256, opened 2021-12-02
FEATURE: Reactions!

markvanlan:
Screen Shot 2021-12-02 at 2 39 13 PM

Screen Shot 2021-12-02 at 2 39 22 PM

Screen Shot 2021-12-02 at 2 39 30 PM

Screen Shot 2021-12-02 at 2 39 41 PM

Screen Shot 2021-12-02 at 2 39 48 PM

Screen Shot 2021-12-02 at 2 39 54 PM


app/serializers/chat_base_message_serializer.rb
```diff
@@ -11,13 +11,35 @@ class ChatBaseMessageSerializer < ApplicationSerializer
...
+        count: reactions.count,
+        users: ActiveModel::ArraySerializer.new(users, each_serializer: BasicUserSerializer).as_json,
+        reacted: users_reactions.include?(emoji)
+      }
+    end
+    reactions_hash
+  end
+
+  def include_reactions?
+    object.reactions.any?
```
CvX, line 36: Could you double check this doesn't cause an extra `SELECT 1 FROM…` query?

markvanlan: I did, yes!

assets/javascripts/discourse/components/chat-message-reaction.js
```diff
@@ -0,0 +1,47 @@
+import Component from "@ember/component";
+import discourseComputed from "discourse-common/utils/decorators";
+import { bind } from "discourse-common/utils/decorators";
```
CvX, line 3:
```suggestion
import discourseComputed, { bind } from "discourse-common/utils/decorators";
```

😛

CvX (approved):

assets/javascripts/discourse/templates/components/tc-message.hbs
```diff
@@ -21,6 +27,12 @@
         {{#unless message.staged}}
           <div class="tc-msgactions-hover">
             <div class="tc-msgactions">
+              {{d-button
+                class="btn-flat react-btn"
+                action=(action "startReaction" ".react-btn" SHOW_LEFT)
```
pmusaraj, line 32: Feels a little unusual to pass `.react-btn`, both here, and in the `.chat-message-react-btn` below.

assets/javascripts/discourse/components/tc-message.js
```diff
@@ -158,6 +199,170 @@ export default Component.extend({
...
+  hasReactions(reactions) {
+    return Object.values(reactions).some((r) => r.count > 0);
+  },
+
+  @action
+  showUsersList(reaction) {
+    let list;
+    let usernames = reaction.users.map((u) => u.username).join(", ");
+    if (reaction.reacted) {
+      if (reaction.count === 1) {
```
pmusaraj, line 212: I would find a `switch case` statement more readable here.

assets/javascripts/discourse/components/tc-message.js
```diff
@@ -158,6 +199,170 @@ export default Component.extend({
...
+      "chat-message:reaction-picker-opened",
+      this.message.id
+    );
+
+    schedule("afterRender", () => {
+      this._repositionEmojiPicker(btnSelector, position);
+    });
+  },
+
+  _repositionEmojiPicker(btnSelector, position) {
```
pmusaraj, line 261: This repositioning of the emoji picker feels a little fragile. Could we use Popper instead? 

Also, not sure how this currently works on mobile, you probably went something full width there.

assets/javascripts/discourse/templates/components/tc-message.hbs
markvanlan: those are selectors to know which button was pressed, while passing in the `SHOW_LEFT` arg. I think it makes the JS much nicer.

assets/javascripts/discourse/components/tc-message.js
pmusaraj: Nevermind (<> not as fast with switches)

assets/javascripts/discourse/components/tc-message.js
pmusaraj: Ignore this too, pulled the repo locally, works quite well.

PR #262, opened 2021-12-03
FIX: Remove links around chat channel title

markvanlan:
This does a few things:
* Sets max-height on chat channel settings modal
* Removes links around chat-channel-title in favor of passed in action
  *  Huge benefit of this is that we can make the title expand the widget when it is hidden on click.
* Expands widget when it was collapsed and sidebar link is clicked on
* Cleans up a ton of css

### DIFF
--- a/assets/javascripts/discourse/components/chat-channel-title.js
+++ b/assets/javascripts/discourse/components/chat-channel-title.js
@@ -14,8 +14,6 @@ export default Component.extend({
   },
 
   click() {
-    if (this.onClick) {
-      return this.onClick();
-    }
+    return this.onClick?.();
   },
 });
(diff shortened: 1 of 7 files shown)


assets/javascripts/discourse/components/chat-channel-title.js
```diff
@@ -12,4 +12,10 @@ export default Component.extend({
   usernames(users) {
     return users.map((user) => user.username).join(", ");
   },
+
+  click() {
+    if (this.onClick) {
+      return this.onClick();
+    }
```
CvX, line 19:
```suggestion
    return this.onClick?.();
```

😏

CvX (approved):

PR #268, opened 2021-12-06
FIX: Drop --full-page-chat-height to fix app height issue

CvX:
No more listening to "resize" event or manual height var calculation. The issue in the Desktop app was that the content was taller than the app window (when you had a lot of channels/personal messages)


CvX:
@markvanlan Updated with fixes for the sidebar!

markvanlan (approved):
Let's give it a shot!

PR #288, opened 2021-12-07
FIX: If in PWA with chat isolated, no new tab

markvanlan:
https://github.com/discourse/discourse/pull/15218


assets/javascripts/discourse/widgets/chat-header-icon.js
```diff
@@ -74,6 +78,14 @@ export default createWidget("header-chat-link", {
     }
   },
 
+  isInPwa() {
```
eviltrout, line 81: Should we move this to capabilities?

markvanlan: sure!

eviltrout (approved):

PR #293, opened 2021-12-08
FIX: payload of incoming webhooks can be a JSON string

OsamaSayegh:


app/controllers/incoming_chat_webhooks_controller.rb
```diff
@@ -26,7 +26,12 @@ def create_message_slack_compatable
 
     # See note in validate_payload on why this is needed
     attachments = if params[:payload].present?
-      params[:payload][:attachments]
+      payload = params[:payload]
+      if String === payload
+        payload = JSON.parse(payload)
```
SamSaffron, line 31: maybe allow for the parse to fail? do we need some protection here?

SamSaffron (approved):

app/controllers/incoming_chat_webhooks_controller.rb
OsamaSayegh: added a handler for `JSON::ParserError`  exceptions which will respond with a 400 if the JSON is malformed.

PR #294, opened 2021-12-08
UI: implements link to message button

jjaffeux:
Kapture 2021-12-08 at 15 28 42

### DIFF
--- a/assets/javascripts/discourse/components/tc-message.js
+++ b/assets/javascripts/discourse/components/tc-message.js
@@ -1,3 +1,4 @@
+import getURL from "discourse-common/lib/get-url";
 import Component from "@ember/component";
 import discourseComputed, { bind } from "discourse-common/utils/decorators";
 import I18n from "I18n";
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse/components/tc-message.js
```diff
@@ -444,4 +445,31 @@ export default Component.extend({
...
+
+  @action
+  copyLinkToMessage() {
+    this.element
+      .querySelector(".link-to-message-btn")
+      ?.classList?.add("copied");
+
+    const { protocol, host } = window.location;
+    const endpoint = `/chat/channel/${this.details.chat_channel_id}/chat?messageId=${this.message.id}`;
+    const url = `${protocol}//${host}${getURL(endpoint)}`;
```
markvanlan, line 457: Do you really need protocol and host here? I thought `getURL(endpoint)` would be enough

markvanlan (approved):
Thank you!!! I have been meaning to do this.

assets/javascripts/discourse/components/tc-message.js
jjaffeux: yes right, it needs better handling in some cases

PR #296, opened 2021-12-08
FIX: Properly highlight chat messages

markvanlan:
I know important is bad, but it's IMPORTANT! We have the background color set to various things (in widget or in full-screen), and then we have hover background color, but we waant to always ignore all of that and show the highlight when it should be there.


CvX (approved):
It won't be the end of the world if we introduce a single `!important` though I suppose the "correct" solution would be to make sure that when `.chat-message` elements have the `highlighted` class, they don't have other background-changing classes.

PR #300, opened 2021-12-08
FEATURE: Client-side per-channel drafts

markvanlan:
Very simple implementation: store text/uploads in a `drafts` object in the chat service, with the channel ID as the key. When a new chat channel is opened, load the draft into the composer, and keep it updated while the user inputs text/uploads/sends messages.

Coolest thing about this is if you are typing in the widget and expand into full page, your uploads/text will still be there :)

Oh also, this make the composer textarea **not-** disabled when uploads are processing. That was old logic when the uploads went into markdown


CvX (approved):
🎉

eviltrout (approved):

PR #302, opened 2021-12-08
UX: Don't do the ugly text bounce on edit

markvanlan:
There was also a double check for if the `messageIsValid` for both `internalEditMessage` and `internalSendMessage` (they're both called after the verifications)

### DIFF
--- a/assets/javascripts/discourse/components/chat-composer.js
+++ b/assets/javascripts/discourse/components/chat-composer.js
@@ -493,16 +493,12 @@ export default Component.extend(TextareaTextManipulation, ComposerUploadUppy, {
 
   @action
   internalSendMessage() {
-    if (this._messageIsValid()) {
-      return this.sendMessage(this.value, this.uploads).then(() =>
-        this.reset()
-      );
-    }
+    return this.sendMessage(this.value, this.uploads).then(this.reset);
   },
 
   @action
   internalEditMessage() {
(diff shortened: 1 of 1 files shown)


assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -493,22 +493,14 @@ export default Component.extend(TextareaTextManipulation, ComposerUploadUppy, {
...
   internalEditMessage() {
-    if (this._messageIsValid()) {
-      return this.editMessage(
-        this.editingMessage,
-        this.value,
-        this.uploads
-      ).then(() => this.reset());
-    }
+    return this.editMessage(this.editingMessage, this.value, this.uploads).then(
+      this.reset
```
eviltrout, line 502: Note doing it this way can mess up the `this` context. If it works that's great but it's generally safer to do it via `() => this.reset()`

eviltrout (approved):

CvX (approved):

assets/javascripts/discourse/components/chat-composer.js
markvanlan: Since it's an action, `this` is bound properly

markvanlan: I mean `reset` is an action so it's bound. Do you prefer the other style? I can swap

PR #305, opened 2021-12-09
FEATURE: hold back on emoji autocomplete

SamSaffron:
In chat since "ENTER" submits a message we need to be a bit more protective
of autocomplete dialog.

Hold off for at least 2 chars and avoid autocomplete unless there are 2 non
symbol letters, this catches most commonly typed word emojis

### DIFF
--- a/assets/javascripts/discourse/components/chat-composer.js
+++ b/assets/javascripts/discourse/components/chat-composer.js
@@ -341,7 +341,16 @@ export default Component.extend(TextareaTextManipulation, ComposerUploadUppy, {
           const full = `:${term}`;
           term = term.toLowerCase();
 
-          if (term.length < this.siteSettings.emoji_autocomplete_min_chars) {
+          // We need to avoid quick emoji autocomplete cause it can interfere with quick
+          // typing, set minimal length to 2
+          let minLength = Math.max(this.siteSettings.emoji_autocomplete_min_chars, 2);
+
+          if (term.length < minLength) {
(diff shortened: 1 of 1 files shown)


martin-brennan (approved):
approved, but prettier is not happy with you :)

SamSaffron:
lol, not configured here ... will fix ....

assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -341,7 +341,23 @@ export default Component.extend(TextareaTextManipulation, ComposerUploadUppy, {
...
+          );
+
+          if (term.length < minLength) {
+            return resolve(SKIP);
+          }
+
+          // bypass :-p and other common typed smileys
+          if (
+            !term.match(
+              /[^-\{\}\[\]\(\)\*_\<\>\\\/].*[^-\{\}\[\]\(\)\*_\<\>\\\/]/
```
tgxworld, line 358: Sorry just a follow up comment but I think having test cases for this regexp is important. There is a comment left but common typed smileys is loosely defined and figuring out what the regexp is trying to do is not straight forward.

SamSaffron: testing autocomplete is very difficult at the moment, very flaky, we need to fix that.

tgxworld: I don’t mean we should test autocomplete but the matching of the regexp.

PR #306, opened 2021-12-09
FEATURE: Highlight my mentions and @here/@all

OsamaSayegh:
image

### DIFF
--- a/assets/javascripts/discourse/components/tc-message.js
+++ b/assets/javascripts/discourse/components/tc-message.js
@@ -32,6 +32,23 @@ export default Component.extend({
     );
   },
 
+  didInsertElement() {
+    this._super(...arguments);
+    if (!this.currentUser) {
+      return;
+    }
+    this.element
+      .querySelector(".tc-text")
+      .querySelectorAll(".mention")
+      .forEach((node) => {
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse/components/tc-message.js
```diff
@@ -32,6 +32,23 @@ export default Component.extend({
...
+    this._super(...arguments);
+    if (!this.currentUser) {
+      return;
+    }
+    this.element
+      .querySelector(".tc-text")
+      .querySelectorAll(".mention")
+      .forEach((node) => {
+        const mention = node.textContent.trim().substr(1);
+        const highlightable = [this.currentUser.username, "here", "all"];
```
jjaffeux, line 45: maybe have here and all in some constant?

OsamaSayegh: sure, done!

pmusaraj (approved):

pmusaraj:
Unrelated to this PR: in your screenshot that `@here` and `@all` don't get rounded corners.

OsamaSayegh:
Yes, this is because `@here` and `@all` are not styled the same as regular mentions (no background color at all, they're just bold). Not sure if this is intentional or not.

PR #310, opened 2021-12-09
FIX: Attempt to fix the sporadic defocusing problem

davidtaylorhq:
- Remove the `.blur()` immediately before `.focus()`
- Move the `_focusTextArea()` call from `didRender` to `didInsertElement`
- Remove the `_messageIsEmpty` check from this call. IMO we still want to focus, even if there's a draft message in the textarea.


markvanlan (approved):

markvanlan:
let's give it a shot! Not sure about the blur.. can't remember.

PR #311, opened 2021-12-09
Fix textarea input throttling

davidtaylorhq:
`throttle` doesn't work with anonymous functions. If you do this, throttle is being handed a brand new function instance each time, and it has no way to know they are identical.


CvX (approved):

markvanlan:
nice, thanks david

PR #320, opened 2021-12-10
FIX: Input blur from user tracking state sync

martin-brennan:
In be2db2dd723d1bf4a0d33cf1b78e94c3f4dfb007 we introduced
this method of detecting that the chat window should have
its channels resynced if we fall too far behind from dropped
MessageBus messages. This full resync causes a route transition
and the input to blur from cleanDOM in core. However we did
not update `lastUserTrackingMessageId` if this reload happened,
leaving the state in a permanently bad state.

This commit always updates `lastUserTrackingMessageId` if the
new chat messageId is greater than the `lastUserTrackingMessageId`
regardless of whether all the channels are refreshed.


SamSaffron (approved):

SamSaffron:
lets give this a shot!

PR #322, opened 2021-12-10
FEATURE: optionally handles discourse-spoiler-alert

jjaffeux:


markvanlan (approved):
While you're at it if you want to solve the issue of not erroring when local-dates is not installed (tobias ran into this issue), by removing this import and using `requirejs` I would not be mad https://github.com/discourse/discourse-chat/blob/66a59d1260f14eccd85cf7cdb566d80b186093fa/assets/javascripts/discourse/components/chat-live-pane.js#L13

jjaffeux:
> While you're at it if you want to solve the issue of not erroring when local-dates is not installed (tobias ran into this issue), by removing this import and using `requirejs` I would not be mad
> 
> https://github.com/discourse/discourse-chat/blob/66a59d1260f14eccd85cf7cdb566d80b186093fa/assets/javascripts/discourse/components/chat-live-pane.js#L13

how would that be possible? this is a core plugin

PR #328, opened 2021-12-10
FEATURE: Mark unread chat mentions as read when messages are read

markvanlan:
I'm sorry @eviltrout I have to do multiple queries here for the AR callbacks to run and send message_bus data to the client to the count actually clears.

Also lower time to enqueue notify job to 3 seconds. 5 is so long.

### DIFF
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -278,6 +278,15 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
       .where(user: current_user, chat_channel: @chat_channel)
       .update_all(last_read_message_id: params[:message_id])
 
+    mention_notifications = Notification
+      .where(user: current_user)
+      .where(notification_type: Notification.types[:chat_mention])
+      .where("data LIKE ?", "%\"chat_channel_id\":#{@chat_channel.id}%")
+
+    mention_notifications.each do |mention|
+      mention.update(read: true)
+    end
+
     ChatPublisher.publish_user_tracking_state(
       current_user,
       @chat_channel.id,
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -281,6 +281,7 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
     mention_notifications = Notification
       .where(user: current_user)
       .where(notification_type: Notification.types[:chat_mention])
+      .where(read: false)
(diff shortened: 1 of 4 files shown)


app/controllers/chat_controller.rb
```diff
@@ -278,6 +278,15 @@ def set_user_last_read
       .where(user: current_user, chat_channel: @chat_channel)
       .update_all(last_read_message_id: params[:message_id])
 
+    mention_notifications = Notification
+      .where(user: current_user)
+      .where(notification_type: Notification.types[:chat_mention])
+      .where("data LIKE ?", "%\"chat_channel_id\":#{@chat_channel.id}%")
```
CvX, line 284: Will this be in any way performant on sites with loads of `Notification` records? 🤔

markvanlan: The plan is to create a `chat_mentions` table that stores the notification_id and go off that in the future. B/c yes I bet this query is slow AF

markvanlan: probably not too bad though. It's scoped to the user AND `chat_mention` notification type

markvanlan: Oh I should add `unread: true`

markvanlan: thereeeee we go

app/controllers/chat_controller.rb
```diff
@@ -278,6 +278,16 @@ def set_user_last_read
...
       .update_all(last_read_message_id: params[:message_id])
 
+    mention_notifications = Notification
+      .where(user: current_user)
+      .where(notification_type: Notification.types[:chat_mention])
+      .where(read: false)
+      .where("data LIKE ?", "%\"chat_channel_id\":#{@chat_channel.id}%")
+
+    mention_notifications.each do |mention|
+      mention.update(read: true)
```
CvX, line 288:
```suggestion
      mention.update!(read: true)
```

(unless there's a reason to use the non-exploding call 😃)

CvX (approved):

app/controllers/chat_controller.rb
CvX: …actually, if there's a chance of many records in `mention_notifications` maybe doing the update in a single query would be better?

markvanlan: I actually do want this to be a non-exploding call, and a failure there would break all of chat state tracking (well it just wouldn't publish so you'd have to refresh)

PR #329, opened 2021-12-10
FIX: Don't scroll the entire chat pane to GitHub oneboxes

davidtaylorhq:
When a GitHub onebox renders off-screen, calling `scrollIntoView` would scroll the entire chat window up to the onebox. This is not what we want.

All we want to do is scroll the `.onebox .onebox-body` div so that the selected line is in view

Unfortunately calculating the y-position of that line relative to the `.onebox-body` is quite complicated, but I think this should be pretty robust.

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -942,7 +942,31 @@ export default Component.extend({
     this.element
       .querySelectorAll(".onebox.githubblob li.selected")
       .forEach((line) => {
-        line.scrollIntoView({ block: "center" });
+        const scrollingElementSelector = ".onebox .onebox-body";
+
+        let linePosition = line.offsetTop + line.offsetHeight / 2;
+
+        // offsetTop is relative to the offsetParent (i.e. a position: relative element)
+        // Keep iterating up them until we reach the topmost one within the onebox
+        // Sum up the offsetTop values as we go
+        let offsetParent = line.offsetParent;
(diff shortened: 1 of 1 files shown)


jjaffeux:
Hmm weird I tested this though... Will look Monday in front of my computer, what about we go intersection observer route instead? That seems overly complicated.

davidtaylorhq:
> what about we go intersection observer route instead

Intersection observer gives us callbacks when DOM element intersections change... I don't see how we can use that to pre-emptively scroll a Div to show a specific child element?

But yes I agree this is very complicated. Unfortunately I couldn't find any simpler way. 

As you mentioned earlier today, the 'correct' solution is for us to restyle the GitHub onebox to work better in a chat UI without needing this scroll hack.

jjaffeux:
This being said Im ok to merge this to avoid issues meanwhile. thanks 👍

jjaffeux (approved):

jjaffeux:
> > what about we go intersection observer route instead
> 
> 
> 
> Intersection observer gives us callbacks when DOM element intersections change... I don't see how we can use that to pre-emptively scroll a Div to show a specific child element?
> 
> 
> 
> But yes I agree this is very complicated. Unfortunately I couldn't find any simpler way. 
> 
> 
> 
> As you mentioned earlier today, the 'correct' solution is for us to restyle the GitHub onebox to work better in a chat UI without needing this scroll hack.

We would scroll only when entering viewport ?

PR #341, opened 2021-12-14
FIX: Revert change to push notification excerpt

markvanlan:
I thought using `excerpt` would be much cleaner, but I forgot it's HTML.... woops

### DIFF
--- a/app/jobs/regular/create_chat_mention_notifications.rb
+++ b/app/jobs/regular/create_chat_mention_notifications.rb
@@ -48,7 +48,7 @@ module Jobs
                                  username: @creator.username
                                 ),
         tag: DiscourseChat::ChatNotifier.push_notification_tag(:mention, @chat_channel.id),
-        excerpt: @chat_message.message[0...400],
+        excerpt: @chat_message.push_notification_excerpt,
         post_url: "/chat/channel/#{@chat_channel.id}/#{@chat_channel.title(membership.user)}?messageId=#{@chat_message.id}"
       }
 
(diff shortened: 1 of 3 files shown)


app/jobs/regular/create_chat_mention_notifications.rb
```diff
@@ -48,7 +48,7 @@ def send_os_notifications(membership)
                                  username: @creator.username
                                 ),
         tag: DiscourseChat::ChatNotifier.push_notification_tag(:mention, @chat_channel.id),
-        excerpt: @chat_message.excerpt,
+        excerpt: @chat_message.message[0..399],
```
nattsw, line 51: I know this is a revert, but why 399?

ZogStriP: Same question 🤔 That number seems quite magical

markvanlan: don't question the magic

markvanlan: IDK why I had this before 🤔

nattsw: I think maybe defining why the number or what it means is nice.

Separately, would also be nice if `send_notifications` could be extracted to test?

markvanlan: There are some integration tests for chat message creation that run jobs immediately and test that push notifications are sent 🤷

nattsw: That's great!

app/models/chat_message.rb
```diff
@@ -27,6 +27,10 @@ def excerpt
     PrettyText.excerpt(cooked, 50, {})
   end
 
+  def push_notification_excerpt
+    message[0...400]
```
nattsw, line 31: I think this is slightly better, we have still no idea why 400 tho. Given you've tried your best to explain but can't yet find anything better, let's let this go and please revisit if you suddenly 💡 

(Is it a char LIMIT?)

nattsw (approved):

PR #342, opened 2021-12-14
FIX: Cancel messages promise when channel changes

markvanlan:


eviltrout (approved):

CvX:
I don't think it does what you think it would! 😃

`cancel()` from `@ember/runloop` "cancels a scheduled item" identified by a timer id, which is returned by e.g. `later()` (from the same module) It does not cancel promises.

And, even if it did 😛 `fetchMessages()` does not seem to actually return anything. 😄

CvX (approved):

PR #343, opened 2021-12-14
FIX: Track all DM channels in frontend even if hidden

markvanlan:
I keep seeing this issue where someone personal chats me, and I don't see it until I do a page refresh. I thought the issue was with personal chat channel creation, but then I realized it's the limit on the DM channel query.

Currently the client is only tracking the latest 10 channels... so if Robin is 11th most recently updated and sends me a new message, I will not get it.

This PR does 2 things:
* Serializer _all_ your DM channels, and hides channels 11-on. This allows tracking of all channels.
* Keeps unread DM channels at the top of the list. First sort by unread messages count, then sort by `updated_at` on the channel

### DIFF
--- a/assets/javascripts/discourse/components/channel-list.js
+++ b/assets/javascripts/discourse/components/channel-list.js
@@ -29,9 +29,25 @@ export default Component.extend({
   sortedDirectMessageChannels: computed(
     "directMessageChannels.@each.updated_at",
     function () {
+      if (!this.directMessageChannels?.length) {
+        return [];
+      }
+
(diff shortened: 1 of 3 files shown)


eviltrout (approved):

CvX:
> First sort by unread messages count, then sort by updated_at on the channel

Wouldn't sorting by updated_at **then** by unread count be better? 🤔 (and more of a standard?)

markvanlan:
CvX and I chatted privately about this and he agreed to give this a shot :)

PR #347, opened 2021-12-15
DEV: initial block-kit implementation

jjaffeux:


assets/stylesheets/common/common.scss
```diff
@@ -1536,3 +1536,79 @@ $float-height: 530px;
     margin-left: 0.5em;
   }
 }
+
+.d-wrap[data-wrap="blocks"] {
```
markvanlan, line 1540: This css should probably be scoped inside chat message, in case core implements a similar `d-wrap`, right?

jjaffeux: qin any case that would cause issues but Im thinking of something more scoped for the named: bk-xxx

assets/javascripts/discourse/lib/block-kit.js
```diff
@@ -0,0 +1,70 @@
+function stringToDOMNode(string) {
```
martin-brennan, line 1: I think we should add this to the core discourse/lib/utilities, we have added the exact same function in a core test and we've also made similar functionality in other places in core (though not inside a dedicated function)

https://github.com/discourse/discourse/blob/1c9a0fe18e19831277fd1c1be6d28559b3bad108/app/assets/javascripts/discourse/tests/unit/lib/formatter-test.js#L13-L18

https://github.com/discourse/discourse/blob/1c9a0fe18e19831277fd1c1be6d28559b3bad108/app/assets/javascripts/pretty-text/addon/oneboxer.js#L78-L80

https://github.com/discourse/discourse/blob/1c9a0fe18e19831277fd1c1be6d28559b3bad108/app/assets/javascripts/discourse/tests/unit/lib/link-mentions-test.js#L44-L47

assets/javascripts/discourse/lib/block-kit.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  init() {
+    this.element
+      .querySelectorAll(".d-wrap[data-wrap=blocks]")
+      .forEach((blocks) => {
+        blocks.querySelectorAll(".d-wrap[data-wrap=block]").forEach((block) => {
+          const { type } = block.dataset;
+
+          if (type === "context") {
```
martin-brennan, line 24: Do we have all these as separate conditions even though they call the same function for future flexibility? Just curious

martin-brennan (commented):
Just a couple of comments, I know you are still working on this so no need to change straight away.

assets/javascripts/discourse/lib/block-kit.js
jjaffeux: Yes I know I already have a PR awaiting merge on this: https://github.com/discourse/discourse/pull/15310 ;)

assets/javascripts/discourse/lib/block-kit.js
jjaffeux: Yes for this demo I just didn't  want to think about this.

assets/javascripts/discourse/lib/block-kit.js
martin-brennan: Oh nice! I saw you couldn't add to discourse-common either, a pain I know too well :joy:

jjaffeux:
Closing for now.

PR #348, opened 2021-12-15
FEATURE?: Mobile chat header icon opens channel list

markvanlan:
This overloads the chat index route a bit to handle 2 cases - 
1. for when the user has not joined any channels
2. channel selector index

The reason for this: We need a channel selector index for mobile chat. Can't use`chat.channel` route, as we need a channel ID and title for that route, and we really just want this channel index to live at `/chat`. This solution seems fine for now, but I could see us moving the blank page to `/chat/get-started` if it needs more complex logic or we add more rich features.


davidtaylorhq (approved):

SamSaffron:
This is so good, thanks heaps.

PR #351, opened 2021-12-15
FEATURE: Open non-chat routes in new tab with chat_isolated

markvanlan:


SamSaffron (approved):

SamSaffron:
Oh ... broken test @markvanlan ...

markvanlan:
@SamSaffron Initializer wasn't re-evaluated between tests.. had to add another `if` ... don't love that

PR #356, opened 2021-12-16
FEATURE: Open non-chat routes in new tab with chat isolated

markvanlan:
Followed the ember guide on this one - https://guides.emberjs.com/release/routing/preventing-and-retrying-transitions/

The only quirk is that the URL will transition even after a route is aborted... There is an ugly hack to fix this using `next` and `replaceState` to fix the URL. It _works_.

EDIT:
There are so many little issues that keep popping up. IDK if this approach will work.


markvanlan:
It does lol

PR #359, opened 2021-12-17
UX: adds support for expanding oneboxes excerpts

jjaffeux:


nattsw (approved):
Would be nice to see what it was and what it became visually

PR #365, opened 2021-12-19
FIX: A proper fix for calculating unread counts

CvX:
A follow-up to #364.

The change in slack_compatibility_spec.rb is unrelated but fixes specs.


markvanlan (approved):
Thanks, much better

PR #367, opened 2021-12-21
UX: Prevent message overflow; circular indicators

CvX:
Mostly message sizing, to prevent them from expanding beyond parent element bounds.
Also: circular (i.e. not oval) presence indicators. 🟢 

Chance of regressing something: pretty high.
But I tried it out in full-chat mode, floating, and mobile. Short messages, long, replies, images.


markvanlan (approved):
Just stick around for regressions ;)

PR #369, opened 2021-12-21
UX: fast access to favorite reactions

jjaffeux:
Note that this commit also:
- implements a ... button to limit the width of the whole tc-message action bar
- fixes few bugs like the emoji picker not closing when re-cliking on button and the button keeping focus


jjaffeux:
Screenshot 2021-12-21 at 20 13 02
Screenshot 2021-12-21 at 20 12 52

assets/javascripts/discourse/components/tc-message.js
```diff
@@ -549,4 +616,26 @@ export default Component.extend({
         ?.classList?.remove("copied");
     }, 250);
   },
+
+  _computeFavoriteEmojis() {
+    next(() => {
```
jjaffeux, line 621: This is needed otherwise emojiStore.favorites might not be updated

markvanlan (approved):
Look fine assuming the JS tests get fixed

PR #373, opened 2021-12-22
FEATURE: Let me talk to myself

nattsw:
And also put my own DM at the top

Chat screen showing a DM to myself and the sidebar
.

### DIFF
--- a/app/controllers/direct_messages_controller.rb
+++ b/app/controllers/direct_messages_controller.rb
@@ -5,21 +5,19 @@ class DiscourseChat::DirectMessagesController < DiscourseChat::ChatBaseControlle
     guardian.ensure_can_chat!(current_user)
     params.require(:usernames)
 
-    users = User
-      .where(username: params[:usernames].split(","))
-      .to_a
-      .concat([current_user])
-    user_ids = users.map(&:id).uniq
-    raise Discourse::InvalidParameters if user_ids.count < 2
+    ActiveRecord::Base.transaction do
+      users = [current_user]
+      users += User.where(username: params[:usernames].split(",")).to_a if current_user.username != params[:usernames]
+      user_ids = users.map(&:id).uniq
 
-    direct_messages_channel = DirectMessageChannel.for_user_ids(user_ids)
-    if direct_messages_channel
-      chat_channel = ChatChannel.find_by(chatable: direct_messages_channel)
-    else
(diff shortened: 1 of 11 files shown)


nattsw:
@markvanlan I feel like you would know this better: My most recent commit was to remove the transaction I added when creating a DM. I found the transaction useful when I was dev-ing so errors along the way would be properly reverted. I don't know if we want it as a thing, so I've removed it and we can add it back if it sounds useful.

app/controllers/direct_messages_controller.rb
```diff
@@ -5,12 +5,9 @@ def create
     guardian.ensure_can_chat!(current_user)
     params.require(:usernames)
 
-    users = User
-      .where(username: params[:usernames].split(","))
-      .to_a
-      .concat([current_user])
+    users = [current_user]
+    users += User.where(username: params[:usernames].split(",")).to_a if current_user.username != params[:usernames]
```
markvanlan, line 9: We generally prefer `<<` as it's a much faster operation

assets/javascripts/discourse/services/chat.js
```diff
@@ -319,6 +319,14 @@ export default Service.extend({
         this.currentUser.chat_channel_tracking_state[a.id]?.unread_count || 0;
       const unreadCountB =
         this.currentUser.chat_channel_tracking_state[b.id]?.unread_count || 0;
+
+      const userAIsSelf = a.chatable.users[0].id === this.currentUser.id;
+      const userBIsSelf = b.chatable.users[0].id === this.currentUser.id;
```
markvanlan, line 324: I don't think your personal DM channel should be at the top of the list always, other chat apps don't do this that I've seen. Was this just something you thought would be good, or have you seen this elsewhere?

nattsw: Slack does that! 🤔

markvanlan (approved):

assets/javascripts/discourse/services/chat.js
nattsw: Leaving this as-is, since we talked about it in chat. :p

CvX: I think I'll avoid messaging myself until there's the ability to close a conversation 😅 I don't want to be stuck with least relevant chat at the top! 😛

PR #378, opened 2021-12-22
UX: Fix another case of messages overflowing the viewport

CvX:


jjaffeux:
Yep seen this one too :+1:

jjaffeux (approved):

PR #381, opened 2021-12-22
FEATURE: Link to chat from browse page

markvanlan:
This is not a sexy change, but it's needed and I likely won't be working much in the next week so I wanted to get this in. Only rendered if on mobile or don't have sidebar.

Screen Shot 2021-12-22 at 4 13 03 PM
Screen Shot 2021-12-22 at 4 12 29 PM


SamSaffron:
a bit odd, I would just have the "Chat" icon in the header take you back, as it acts as a home page.

PR #382, opened 2021-12-23
UX: Chat header icon takes to channel list of mobile and browse

markvanlan:
It's getting more and more magical up in here :)


CvX (approved):

CvX:
Is there still a plan to make the chat icon in the header do anything on mobile? (when the chat is already open)

PR #402, opened 2021-12-30
FEATURE: allows to unfollow direct message channel

jjaffeux:
Screenshot 2021-12-30 at 11 21 20

Kapture 2021-12-30 at 11 24 47

### DIFF
--- a/lib/direct_message_channel_creator.rb
+++ b/lib/direct_message_channel_creator.rb
@@ -3,21 +3,35 @@
 module DiscourseChat::DirectMessageChannelCreator
   attr_reader :chat_channel, :users
 
-  def self.create(users)
-    direct_messages_channel = DirectMessageChannel.create!(users: users)
-    chat_channel = ChatChannel.create!(chatable: direct_messages_channel)
-    users.each do |user|
-      UserChatChannelMembership.create(
-        user_id: user.id,
-        chat_channel_id: chat_channel.id,
-        last_read_message_id: nil,
-        following: true,
-        muted: false,
-        desktop_notification_level: UserChatChannelMembership::NOTIFICATION_LEVELS[:always],
-        mobile_notification_level: UserChatChannelMembership::NOTIFICATION_LEVELS[:always],
-      )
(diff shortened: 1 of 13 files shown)


markvanlan:
So if you unfollow a channel because you don't want it in the list, when that user messages you, you'll never see it until you manually refollow the channel? 

I think we should add a check in message creation, that checks if it's a DM channel, and ensures all members are following

jjaffeux:
> So if you unfollow a channel because you don't want it in the list, when that user messages you, you'll never see it until you manually refollow the channel?
> 
> I think we should add a check in message creation, that checks if it's a DM channel, and ensures all members are following

Indeed good point thanks :+1:

jjaffeux:
@markvanlan ok that should do it: https://github.com/discourse/discourse-chat/pull/402/commits/c171be04732dc0255590a0b69625e1993e73b2f0

jjaffeux:
test failure is fixed in main

lib/direct_message_channel_creator.rb
```diff
@@ -3,21 +3,35 @@
...
-        muted: false,
-        desktop_notification_level: UserChatChannelMembership::NOTIFICATION_LEVELS[:always],
-        mobile_notification_level: UserChatChannelMembership::NOTIFICATION_LEVELS[:always],
-      )
+  def self.create!(users)
+    direct_messages_channel = DirectMessageChannel.for_user_ids(users.map(&:id).uniq)
+    unless direct_messages_channel
+      direct_messages_channel = DirectMessageChannel.create!(users: users)
+      chat_channel = ChatChannel.create!(chatable: direct_messages_channel)
+    else
```
CvX, line 11: v. minor thing: could you swap these around and make it and if-else?

CvX (approved):
Looks good! 🚢

lib/direct_message_channel_creator.rb
jjaffeux: yep sure

PR #405, opened 2021-12-30
UX: simplify date formatting for last week

jjaffeux:


jjaffeux:
spec is fixed in https://github.com/discourse/discourse-chat/pull/406

romanrizzi (approved):

PR #412, opened 2022-01-02
UX: Make the full-page chat more core-like

CvX:
(and a very minor tweak to .first-daily-message)

Before / After
Screen Shot 2022-01-03 at 00 15 59 Screen Shot 2022-01-03 at 00 23 46


jjaffeux (approved):

jjaffeux:
Yes I'm also not a fan of this unique look this page has, nothing has this look in Discourse core.

PR #416, opened 2022-01-03
FEATURE: Collapse youtube previews in chat

nattsw:
Introduce slack-esque folding for youtube onebox. /t/57501

(I actually spent some time looking at core's `lazyYT` to see if I could introduce a "collapsible" setting, but it had proved to be more difficult than just doing just this because of how it's padding the bottom manually.)

Here it is expanded:
expanded youtube preview in chat

Here it is collapsed:
collapsed youtube preview in chat

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -966,7 +966,7 @@ export default Component.extend({
     }
 
     this.element.querySelectorAll(".lazyYT").forEach((iframe) => {
-      $(iframe).lazyYT({ collapsible: true });
+      $(iframe).lazyYT();
       collapseLazyYT(iframe);
     });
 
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -965,7 +966,8 @@ export default Component.extend({
     }
 
     this.element.querySelectorAll(".lazyYT").forEach((iframe) => {
-      $(iframe).lazyYT();
+      $(iframe).lazyYT({ collapsible: true });
```
nattsw, line 969: oh wait this shouldn't be here

markvanlan: I'm also thinking the caret when it's open should point up, to show that you can collapse it, and then when collapsed, the caret should point down

markvanlan (approved):

PR #418, opened 2022-01-03
UX: Unify sidebar row sizes

CvX:


CvX:
@markvanlan would you like to try it our locally or are we doing it live? 😄

markvanlan (approved):
It's already borked. Lets go for it!!!

PR #421, opened 2022-01-03
UX: Fix channel list sizes in floatie and mobile

CvX:


SamSaffron (approved):

SamSaffron:
Nice, I like this a lot, overall the the less magic in mobile css the better.

PR #424, opened 2022-01-04
DEV: Add model annotations

martin-brennan:
We should be doing this for plugins that have more than
one table I think...at least periodically, if not every
commit having it enforced like Discourse Core.


nattsw (approved):

nattsw:
Is this automated?

techAPJ:
@nattsw the annotations are added via [annotate gem](https://github.com/ctran/annotate_models).

```
gem install annotate
bundle exec annotate --models --exclude tests,fixtures,factories,serializers
```

PR #425, opened 2022-01-04
FIX: ensures list of displayed channels is uniq by id

jjaffeux:


CvX (commented):
Is this a temporary solution? Sort* functions should just sort the data, not remove invalid entries. 😃

jjaffeux:
yes so far I haven't been able to repro this issue of duplicates in channels list.

jjaffeux:
I will close it for now, as we can't repro.

PR #426, opened 2022-01-04
UX: smoother state when unfollowing a direct message channel

jjaffeux:


jjaffeux:
I'm trying to write a test ATM, but I guess the core logic shouldn't change

assets/javascripts/discourse/components/chat-channel-row.js
```diff
@@ -14,13 +12,13 @@ export default Component.extend({
   router: service(),
   chat: service(),
 
-  @discourseComputed("active", "channel.muted")
-  rowClassNames(active, muted) {
-    const classes = ["chat-channel-row"];
+  @discourseComputed("active", "channel.{id,muted}")
```
nattsw, line 15: oh wow what. does this mean 

```
@discourseComputed("active", "channel.id", "channel.muted")
```

? Asking because it's new to me.

jjaffeux: yes indeed

jjaffeux: but that's probably not going to be useful in future

assets/javascripts/discourse/components/chat-channel-row.js
```diff
@@ -68,13 +66,7 @@ export default Component.extend({
...
   @action
-  leaveChatChannel() {
-    return ajax(`/chat/chat_channels/${this.channel.id}/unfollow`, {
-      method: "POST",
-    })
-      .then(() => {
-        this.chat.stopTrackingChannel(this.channel);
-      })
-      .catch(popupAjaxError);
+  async leaveChatChannel() {
```
nattsw, line 69: What does it mean when you make the action async? I see it's used in a button in `chat-channel-row` but I'm not sure what impact adding the modifier would have.

jjaffeux: it means that you can do : `await leaveChatChannel`

assets/javascripts/discourse/services/chat.js
```diff
@@ -470,15 +471,42 @@ export default Service.extend({
     });
   },
 
+  unfollowDirectMessageChannel(channel) {
```
nattsw, line 474: Related to above, then does this need `async`?

jjaffeux: yes I can do it :+1:

nattsw: It's turtles all the way down..........

nattsw (approved):

PR #427, opened 2022-01-04
FIX: Use ember component for collapsing instead of yanking into a container

nattsw:
A few bugs were occurring:
- When editing the youtube URL to something else, the lazyYT invocation
was causing Ember to error out when clearing child nodes, because of the
way collapsing was done. This caused the edit to not show up properly
- The title kept getting added when a new message appeared in the chat
because we were redecorating every message very often.

Using the ember component will make sure it renders nicely with the
lifecycle of the `message` and maybe also we can reuse it for images?


nattsw:
Also, not creating a component test because it involves multiple components and the integration test tests all of them.

assets/javascripts/discourse/components/tc-message-collapser.js
```diff
@@ -0,0 +1,34 @@
...
+  title: null,
+
+  init() {
+    this._super(...arguments);
+
+    const cookedElement = new DOMParser().parseFromString(
+      this.cooked,
+      "text/xml"
+    ).firstChild;
+    const title = cookedElement.getAttribute("data-youtube-title");
```
jjaffeux, line 18:
```suggestion
    const title = cookedElement.dataset.youtubeTitle;
```

assets/javascripts/discourse/components/tc-message-collapser.js
```diff
@@ -0,0 +1,34 @@
...
+    const id = cookedElement.getAttribute("data-youtube-id");
```
jjaffeux, line 19:
```suggestion
    const id = cookedElement.dataset.youtubeId;
```

assets/javascripts/discourse/templates/components/tc-message-collapser.hbs
```diff
@@ -0,0 +1,25 @@
+<div class="tc-message-collapsible">
+  <span>
```
jjaffeux, line 2: is this span necessary?

nattsw: Not any more. It was a remnant of the past~

assets/javascripts/discourse/components/tc-message-collapser.js
nattsw: Unfortunately this isn't possible as `firstChild` is a `Node` not a `HTMLElement`. Even `firstElementChild` doesn't return an element but a document T_T

You can try it out

```
const s = '<div data-youtube-id="WaT_rLGuUr8" data-youtube-title="Japanese Katsu Curry (Pork Cutlet)"></div>';
const t = new DOMParser().parseFromString(s, "text/xml");
t.firstChild.getAttribute("data-youtube-title") // 'Japanese Katsu Curry..
t.firstChild.dataset.youtubeTitle // error
t.firstElementChild.dataset.youtubeTitle // error
```

assets/javascripts/discourse/components/tc-message-collapser.js
nattsw: Same as above.

assets/javascripts/discourse/components/tc-message-collapser.js
jjaffeux: and what about using discourse helper `domFromString`: https://github.com/discourse/discourse/blob/main/app/assets/javascripts/discourse-common/addon/lib/dom-from-string.js


I'm actually not sure what would be the difference between the two solutions here 🤔

jjaffeux (approved):

PR #434, opened 2022-01-04
FIX: ensures loading state is removed

jjaffeux:
Before this fix we could enter the previous condition:

```javascript
if (this.chatChannel.id !== channelId) {
  return;
}
```

And would never reach the point where we set to false the loading property.


pmusaraj (approved):

jjaffeux:
Note that I think it's probably hiding bigger issues, but in any case we shouldn't prevent loading state to be removed if loading is done.There's an error to handle or a state to re-conciliate  higher in the chain most probably.

PR #436, opened 2022-01-04
FIX: reconciliate state of live-pane

jjaffeux:
This is not the root issue, we should investigate more on why in some cases, we are processing a request to fetch a list of messages we don't care about anymore.

This commit will ensure that if it ever happens, we try to force transition to the channel the chat-live-pane is currently in, instead of showing a blank page.


CvX (approved):

pmusaraj (approved):
Yeah, it seems odd that we can end up in this state.

PR #438, opened 2022-01-05
DEV: Extract tc-text so that tc-message isn't too large

nattsw:
Also to set up collapsing for images. `tc-message` is also growing so this feels like it's a useful complexity reduction in that file.

I may have a look at extracting reactions (`chat-message-reaction-list`) too to make `tc-message` less overwhelming. On first look it does look quite trivial to extract but there are a lot of minor details which need to be tested to prevent regressions.


tgxworld (approved):

nattsw:
Component test is in upcoming PR

PR #439, opened 2022-01-05
FIX: user is not available on AnonymousUser

jjaffeux:


CvX (approved):

jjaffeux:
> Looks good but I'm wondering if we should start having serializer tests

Yes we should, there's logic in serializers so I think we should

PR #440, opened 2022-01-05
FIX: Fix margins for multiline text

nattsw:
Before:
Screenshot 2022-01-05 at 6 47 28 PM

After:
Screenshot 2022-01-05 at 6 48 06 PM


assets/stylesheets/common/common.scss
```diff
@@ -1884,3 +1875,17 @@ body.composer-open .topic-chat-float-container {
     }
   }
 }
+
+.tc-text {
```
nattsw, line 1879: Extracting this out as I'm using this for image uploads too.

lis2 (approved):

PR #441, opened 2022-01-05
FEATURE: Collapse images

nattsw:
Introduces a slack-esque image collapsing ability

One file with no accompanying text
Showing one image with file name

Folded
Screenshot 2022-01-05 at 7 10 14 PM

One file with accompanying text
Screenshot 2022-01-05 at 7 06 49 PM

Folded
Screenshot 2022-01-05 at 7 10 25 PM

Multiple files with multiple lines of accompanying text
Screenshot 2022-01-05 at 7 07 01 PM

Sanity check for youtube collapsing
Screenshot 2022-01-05 at 7 07 17 PM

### DIFF
--- a/assets/javascripts/discourse/templates/components/tc-text.hbs
+++ b/assets/javascripts/discourse/templates/components/tc-text.hbs
@@ -1,8 +1,7 @@
 <div class="tc-text">
-
   {{#if isYoutubeCollapsible}}
     {{tc-message-collapser cooked=cooked}}
-  {{else if isImageCollapsible }}
+  {{else if isImageCollapsible}}
     {{tc-message-collapser cooked=cooked uploads=uploads}}
   {{else}}
     {{html-safe cooked}}
@@ -15,5 +14,4 @@
   {{/if}}
 
   {{yield}}
-
 </div>
--- a/assets/javascripts/discourse/templates/components/tc-text.hbs
+++ b/assets/javascripts/discourse/templates/components/tc-text.hbs
@@ -2,17 +2,10 @@
 
   {{#if isYoutubeCollapsible}}
     {{tc-message-collapser cooked=cooked}}
-
   {{else if isImageCollapsible }}
-    <div class="tc-uploads">
-      {{#each uploads as |upload|}}
-        {{tc-message-collapser cooked=cooked upload=upload}}
-      {{/each}}
-    </div>
-
+    {{tc-message-collapser cooked=cooked uploads=uploads}}
   {{else}}
     {{html-safe cooked}}
-
   {{/if}}
 
   {{#if edited}}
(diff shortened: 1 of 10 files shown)


test/javascripts/components/tc-message-collapser-test.js
```diff
@@ -0,0 +1,145 @@
...
+discourseModule(
+  "Discourse Chat | Component | tc message collapser",
+  function (hooks) {
+    setupRenderingTest(hooks);
+    const youtubeCooked =
+      '<div class="onebox lazyYT lazyYT-container" data-youtube-id="WaT_rLGuUr8" data-youtube-title="Japanese Katsu Curry (Pork Cutlet)"/>';
+    const imageCooked = "<p>A picture of Tomtom</p>";
+
+    // youtube
+    componentTest("shows youtube link in header", {
```
jjaffeux, line 20: I might be wrong on this, but my feeling is this text is supposed to define context not the expected output.

Expected output can be written down in last parameter of the assertions:

```
assert.ok(true, "it shows the truth")
```

jjaffeux:
You have a legit test failure I think

nattsw:
Yeah, there's an extra phantom space in the filename

test/javascripts/components/tc-message-collapser-test.js
nattsw: qunit makes it weird to define context since it doesn't have the nice `context` block to wrap around a set of component tests. Ideally the ones after `//youtube` are all for youtube and the ones after `//images` are all for images.

nattsw: that said maybe I can extract them into two separate files and have the `discourseModule` description say so

jjaffeux: You can create multiple acceptance() blocks for this, agreed this is more complicated than rspec but 🤷‍♂️

jjaffeux:
> Will remove `tc-` prefixes to files in favour of just `chat` in a next PR

I feel like chat-text is not a great name too, as actually... it doesn't actually hold "text" but more "html". What about `chat-message-content` or something like this?

nattsw:
> `chat-message-`

Yup I was thinking of that

assets/javascripts/discourse/templates/components/tc-text.hbs
```diff
@@ -1,20 +1,15 @@
...
-    <div class="tc-message-edited">
-      {{i18n "chat.edited"}}
-    </div>
-  {{/if}}
-  {{yield}}
-</div>
+{{#if isYoutubeCollapsible}}
+  {{tc-message-collapser cooked=cooked}}
+{{else if isImageCollapsible}}
+  {{tc-message-collapser cooked=cooked uploads=uploads}}
```
jjaffeux, line 4: one part of me thinks it's strange that we need to have knowledge of uploads in tc-message-collapser

nattsw: I did think hard about it, but if we think about responsibilities, I don't think I want `tc-text` to have knowledge on how to define collapsed titles or text. Even putting `isYoutubeCollapsible` is bleeding collapsing knowledge to `tc-text` already.

Ideally what I wanted is another component 😅 `text -> collapsible -> youtube-collapsible OR image-collapsible` but I decided against premature optimisation

jjaffeux (approved):

PR #442, opened 2022-01-05
REFACTOR: implements unread indicator as a component

jjaffeux:


jjaffeux:
You can go heavy with the nitpicks on this one :+1:

jjaffeux:
Out for some time, will look at test failure later

assets/javascripts/discourse/components/chat-channel-unread-indicator.js
```diff
@@ -0,0 +1,39 @@
...
+  },
+
+  @discourseComputed("currentUserTrackingState")
+  unreadCount(trackingState) {
+    if (!this.channel) {
+      return;
+    }
+
+    return trackingState?.[this.channel.id]?.unread_count || 0;
+  },
```
CvX, line 38: For completion sake, I'd add `channel` (and `isDirectMessage`) to computed dependencies

assets/javascripts/discourse/models/chat-channel.js
```diff
@@ -0,0 +1,5 @@
+export const CHATABLE_TYPES = {
+  directMessageChannel: "DirectMessageChannel",
+  topicChannel: "Topic",
+  categoryChannel: "Category",
+};
```
CvX, line 5: I'd expect `models/chat-channel` to export a `ChatChannel` model, not an enum 😉 It should probably live in `lib` or some such.

assets/javascripts/discourse/components/chat-channel-unread-indicator.js
jjaffeux: I was thinking of this, and this can't actually happen no? the upper component would totally be replaced if this was to happen, but I guess it doesn’t harm and at least you are covered if you change context.. 🤔 

Ok will do :p

assets/javascripts/discourse/models/chat-channel.js
jjaffeux: Not sure about this, I agree I would expect the default to export a ChatChannel, but having the enum defined in this file does seem legit?

CvX: If there was a default export, sure, an additional export would be fine.
But in general we want filenames/paths to fit contents. If you look at the list of models and see `chat-channel` file, what would you expect to see inside? 😃

jjaffeux: no sorry I still don't agree. It's not because we haven't created a chat-channel model yet that I should put this code in lib. Its properties of chat-channel model so it makes sens in this file.

eviltrout (approved):

PR #445, opened 2022-01-05
FIX: Simplify the layout to fix an overflow issue

CvX:


markvanlan (approved):
Looks simpler! I hope it works hehe

PR #446, opened 2022-01-05
FEATURE: Store reply-to in draft

markvanlan:
@CvX I know you'll have the async in the test... but the chat float component waits on a promise to check the channels so... idk how else to do this shiz.

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -752,6 +752,7 @@ export default Component.extend({
     this.messages.pushObject(stagedMessage);
     this._resetAfterSend();
     this._stickScrollToBottom();
+    this.appEvents.trigger("chat-composer:reply-to-set", null);
     return Promise.resolve();
   },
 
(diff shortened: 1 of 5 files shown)


CvX:
UX question: should it maybe save reply-to info in a message draft instead of just clearing it?

markvanlan:
Yeah, that's a better idea isn't it

pmusaraj:
I ran into something adjacent to this just now: when you hit arrow up to edit your last message but then hit Esc to cancel it, it cancels the reply state, but doesn't clear the input. It feels like it should?

markvanlan:
Yeah I can clear the input there!

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -908,14 +911,28 @@ export default Component.extend({
 
   @action
   _setDraftForChannel(draft) {
+    if (draft?.replyToMsg) {
+      draft.replyToMsg = {
+        id: draft.replyToMsg.id,
+        excerpt: draft.replyToMsg.excerpt,
+        user: draft.replyToMsg.user,
+      };
```
markvanlan, line 919: Here we don't care to store ALL the chat message data, just what we need to render the reply-to.

CvX (approved):

PR #447, opened 2022-01-05
UX: Remove reply-to line if message is directly above

markvanlan:
We can leave the reply_to information going to the server and simple not render the in_reply_to on the client.


ZogStriP (approved):

ZogStriP:
Should this have a test as well?

markvanlan:
@ZogStriP It does, see it broke :trollface:

ZogStriP (approved):

PR #454, opened 2022-01-06
UX: Show better copy when creating channels

nattsw:
Before:
Screenshot 2022-01-06 at 3 22 39 PM

Shows a link to categories when no category is selected:
Screenshot 2022-01-06 at 3 34 19 PM

Shows a link to /c/category-slug/edit/security if a category is selected
Screenshot 2022-01-06 at 3 32 06 PM

### DIFF
--- a/assets/javascripts/discourse/controllers/create-channel-modal.js
+++ b/assets/javascripts/discourse/controllers/create-channel-modal.js
@@ -33,6 +33,20 @@ export default Controller.extend(ModalFunctionality, {
     return !this.entitySelected || !name?.length > 0;
   },
 
+  @discourseComputed("category")
+  categoryHint(category) {
+    if (category) {
+      return {
+        link: `/c/${category.slug}/edit/security`,
+        category: category.name,
+      };
+    }
+    return {
+      link: "/categories",
+      category: "category",
+    };
+  },
+
   @action
   onCategoryChange(categoryId) {
     let category = categoryId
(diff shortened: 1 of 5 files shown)


config/locales/client.en.yml
```diff
@@ -87,12 +87,12 @@ en:
 
       create_channel:
         choose_category:
-          label: "Choose a category to manage who can access the channel"
+          label: 'Choose a category<br>Manage access by visiting <a href=%{link} target="_blank">%{category} security settings</a>'
```
tgxworld, line 90: I can't remember but are category names escaped by default? Otherwise, this may result in an XSS.

assets/javascripts/discourse/controllers/create-channel-modal.js
```diff
@@ -33,6 +33,20 @@ export default Controller.extend(ModalFunctionality, {
     return !this.entitySelected || !name?.length > 0;
   },
 
+  @discourseComputed("category")
+  categoryHint(category) {
+    if (category) {
+      return {
+        link: `/c/${category.slug}/edit/security`,
+        category: category.name,
```
tgxworld, line 41: I can't remember but are category names escaped by default? Otherwise, this may result in an XSS since we use it in a translation that is marked `html-safe`. Maybe we should escape this just to be safe?

nattsw: You're absolutely right

tgxworld (approved):

tgxworld:
Looks good to me. Thank you for making the changes :+1:

PR #456, opened 2022-01-06
PERFORMANCE: preloads category and topic custom fields

jjaffeux:
Before this change, each public chat channel would trigger a `%{CHATABLE_TYPE}_custom_fields` query and cause a N+1

### DIFF
--- a/lib/chat_channel_fetcher.rb
+++ b/lib/chat_channel_fetcher.rb
@@ -26,7 +26,15 @@ module DiscourseChat::ChatChannelFetcher
         .where(user_chat_channel_memberships: { user_id: guardian.user.id, following: true })
     end
 
-    filter_public_channels(channels, memberships, guardian)
+    channels = filter_public_channels(channels, memberships, guardian).to_a
+
+    preload_fields = Category.instance_variable_get(:@custom_field_types).keys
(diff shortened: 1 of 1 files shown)


lib/chat_channel_fetcher.rb
```diff
@@ -26,7 +26,15 @@ def self.secured_public_channels(guardian, memberships, scope_with_membership: t
         .where(user_chat_channel_memberships: { user_id: guardian.user.id, following: true })
     end
 
-    filter_public_channels(channels, memberships, guardian)
+    channels = filter_public_channels(channels, memberships, guardian).to_a
+
+    preload_fields = Category.instance_variable_get(:@custom_field_types).keys
```
jjaffeux, line 31: There's not way to access `custom_field_types` more directly AFAICT. I could add something in core, but we could still not use it right now as it would force people to update right away to use chat.

But let me know if you think I should add it to core so we can use it in the future.

CvX: Yeah, a proper api would be good. But rather than exposing `custom_field_types` maybe it would be better to change `#preload_custom_fields` to allow preloading all custom fields. Because that's the use case, right?

CvX (approved):

PR #465, opened 2022-01-06
FIX: Pluralize i18n strings

gschlager:


config/locales/client.en.yml
```diff
@@ -185,7 +187,9 @@ en:
       replying_indicator:
         single_user: "%{username} is typing"
         multiple_users: "%{commaSeparatedUsernames} and %{lastUsername} are typing"
-        many_users: "%{commaSeparatedUsernames} and %{count} others are typing"
+        many_users:
+          one: "%{commaSeparatedUsernames} and %{count} other are typing"
```
CvX, line 191: This case is never used, is it? (see: `multiple_users`)

gschlager: I didn't check, but in the end it doesn't matter. In 99% of all cases we need to split it into `one` and `other` when there's a `%{count}`, otherwise it can't be translated correctly in languages with different pluralization rules than English.

CvX: Ok, just saying that this particular string won't be used. (`single_user` is for one typing user, `multiple_users` is for two, `many_users` is for 3+, so count here is 2 or more)

CvX (approved):

PR #466, opened 2022-01-06
FIX: "All loaded" message was shown too early

CvX:
Wrong variable was being used.


jjaffeux (approved):

CvX:
Ah, way to bury the lede: turns out this fixes the bug that prevented loading previous messages in many cases.

PR #469, opened 2022-01-07
FIX: Prevent audio from playing on top of each other CAUSING IT TO BE LOUD

nattsw:
/t/56762

Debounces the playing of audio notifications.

We also have to think about what we want to do when audio playing fails because the user has not interacted with the app due to [an autoplay policy browsers have](https://developer.chrome.com/blog/autoplay/#audiovideo-elements).

I don't have tests for this because I don't think it's possible to unit test this without sufficiently mocking the plugin API and we seem to be against mocking.

### DIFF
--- a/assets/javascripts/discourse/initializers/chat-notification-sounds.js
+++ b/assets/javascripts/discourse/initializers/chat-notification-sounds.js
@@ -5,6 +5,12 @@ export const CHAT_SOUNDS = {
   ding: "/plugins/discourse-chat/audio/ding.mp3",
 };
 
+const MENTION = 29;
+const MESSAGE = 30;
+const CHAT_NOTIFICATION_TYPES = [MENTION, MESSAGE];
+
+const AUDIO_DEBOUNCE_TIMEOUT = 3000;
+
 export default {
   name: "chat-notification-sounds",
   initialize(container) {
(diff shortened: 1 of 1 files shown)


martin-brennan (approved):
Very nice, love the `const`s as well!

assets/javascripts/discourse/initializers/chat-notification-sounds.js
```diff
@@ -13,13 +19,33 @@ export default {
...
+          "User needs to interact with DOM before we can play notification sounds"
+        );
+      });
+    }
+
+    function playAudioWithDebounce(user) {
+      if (canPlay) {
+        canPlay = false;
+
+        setTimeout(() => {
```
markvanlan, line 38: Can you use [debounce](https://api.emberjs.com/ember/release/functions/@ember%2Frunloop/debounce) for this?

martin-brennan: Should be able to, this is just what Mattermost is doing in their source, I linked that code in our internal chat :)

nattsw: https://github.com/discourse/discourse-chat/pull/470

PR #473, opened 2022-01-07
REFACTOR: chat-channel-title

jjaffeux:
- tagLess
- fully tested

### DIFF
--- a/assets/javascripts/discourse/models/chat-channel.js
+++ b/assets/javascripts/discourse/models/chat-channel.js
@@ -1,5 +1,6 @@
 export const CHATABLE_TYPES = {
   directMessageChannel: "DirectMessageChannel",
   topicChannel: "Topic",
+  tagChannel: "Tag",
   categoryChannel: "Category",
 };
--- a/assets/javascripts/discourse/components/chat-channel-title.js
+++ b/assets/javascripts/discourse/components/chat-channel-title.js
@@ -1,18 +1,21 @@
 import discourseComputed from "discourse-common/utils/decorators";
 import Component from "@ember/component";
-import { gt } from "@ember/object/computed";
+import { action } from "@ember/object";
+import { gt, reads } from "@ember/object/computed";
 
 export default Component.extend({
-  classNameBindings: [":chat-channel-title"],
   channel: null,
-  multiDm: gt("channel.chatable.users.length", 1),
+  multiDm: gt("users.length", 1),
+  onClick: null,
+  users: reads("channel.chatable.users.[]"),
 
-  @discourseComputed("channel.chatable.users")
+  @discourseComputed("users")
   usernames(users) {
     return users.mapBy("username").join(", ");
   },
 
-  click() {
-    return this.onClick?.();
+  @action
+  handleOnClick(event) {
+    return this?.onClick(event);
   },
 });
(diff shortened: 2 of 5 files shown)


nattsw:
i feel slightly torn about fabricators. on one hand it feels nice to just be able to create an object like that, but the alternative is that i have to look here and there to see what the defaults are, and sometimes it gets hard to modify one when it grows too big. also, if it’s some kind of unit / component test, i’d rather create the object in the ”unit” itself. if we’d have to reuse an object a lot in many places, it feels like a smell that maybe the component doesn’t exactly just do one thing and one thing well

assets/javascripts/discourse/models/chat-channel.js
```diff
@@ -1,5 +1,6 @@
 export const CHATABLE_TYPES = {
   directMessageChannel: "DirectMessageChannel",
   topicChannel: "Topic",
+  tagChannel: "Tag",
```
nattsw, line 4: I recall @markvanlan saying that tagChannels will be gone soon.

jjaffeux:
> alternative is that i have to look here and there to see what the defaults are

No silver bullet on this, but if you don't do this, you would also have to look what are the expected properties

> and sometimes it gets hard to modify one when it grows too big

yes maybe? that's speculation, and I would rather have one slightly complicated object to change, than having to change dozens of custom objects everywhere in my tests when I change one thing in my model

> if it’s some kind of unit / component test, i’d rather create the object in the ”unit” itself.  if we’d have to reuse an object a lot in many places, it feels like a smell that maybe the component doesn’t exactly just do one thing and one thing well

The alternative to this, is passing only exactly what you need, I feel like it often ends up in gigantic list of properties, is hard to maintain, generates lot of code and make changes in nested components complicated as you have a lot of things to change in the hierarchy, passing the channel property to a component showing the title of a channel doesn't feel like a smell to me

assets/javascripts/discourse/models/chat-channel.js
jjaffeux: Yes I think you are rigjt. I guess we can remove this when they are actually gone.

nattsw:
> that’s speculation 
I don’t think it’s speculation. Take our chat-fixtures file for example. It has grown too big and clunky that when I try to modify one thing, something else breaks.

> gigantic list of properties

If a component is well segregated (aka SRP), we should not have to pass in that many attributes. As an example, when I looked at the chat channel title component, I expect it to be invoked like this `{{chat-channel-title title=channel.title}}` because why should the title component take in an entire channel be aware of it?

But I understand, the refactor is okay. What we’re talking about is something bigger than this PR. Though I would still prefer not to have the fabricator.

nattsw (approved):
I think overall it's okay. I don't want differing *opinions* to get in the way.

jjaffeux:
> If a component is well segregated (aka SRP), we should not have to pass in that many attributes. As an example, when I looked at the chat channel title component, I expect it to be invoked like this `{{chat-channel-title title=channel.title}}` because why should the title component take in an entire channel be aware of it?
> 
> But I understand, the refactor is okay. What we’re talking about is something bigger than this PR. Though I would still prefer not to have the fabricator.

I really don't agree on this, it's not because it's called `{{chat-channel-title}}` that it's only property should be a title... if you go this way you will need A LOT of components. While I don't think this is currently perfect I think this is totally ok, a simple proof of this is that it's very easy to test.

Having only to create one single chat-channel object to test a component is rather lightweight IMO.

jjaffeux:
> I don’t think it’s speculation. Take our chat-fixtures file for example. It has grown too big and clunky that when I try to modify one thing, something else breaks. There’s also evidence of this in core.

I'm sorry... but that doesn’t prove anything. First it's fixtures not fabricators, if you want evidence of fabricators in core, we have theme everywhere and they work quite well. And then I stand by my point that's it's speculation because, I refused to keep creating small channel objects with always the same copy-pasted params and created this small helper.

Wether it will become some gigantic mess is not related to fabricators, but to future developments of this. My point being it has more chance to become a gigantic mess if we keep creating lots of small objects everywhere than trying to DRY it now.

PR #478, opened 2022-01-08
UX: Fix compatibility with custom themes

CvX:
These changes should be no-op with the default themes, but fix some inconsistencies when used with custom ones.


tgxworld (approved):
I didn’t test the changes here locally but it seems safe to me and Low risk.

PR #479, opened 2022-01-08
FIX: Loading past messages

CvX:
A follow-up to #466. Turns out the bug I was hunting was caused by `allPastMessagesLoaded` being persisted between channel changes, which meant that once you saw an "end" of one channel you couldn't scroll-to-load in any channels.


markvanlan:
Nice and simple!

markvanlan (approved):

PR #483, opened 2022-01-10
DEV: Rename chat-message to prepare for tc-message rename

nattsw:
Since most of us don't know what the `tc-` prefix stands for, we are favouring the `chat-` prefix as a replacement. 

This PR is part 1 of the rename, since `tc-message` is to be renamed to `chat-message`, we have to replace the existing `chat-message` usages first.


tgxworld (approved):

tgxworld:
Did not test out locally but the changes are straight forward and this change is low risk IMO.

PR #486, opened 2022-01-10
DEV: Rename tc-text to chat-message-text

nattsw:
Related: https://github.com/discourse/discourse-chat/pull/483

A simple rename.


jjaffeux (approved):
I still think -text is not a great name for something which will contain rich content, but the rename is already better :p

nattsw:
Let's see if we can think of something better in the meantime!

PR #490, opened 2022-01-10
FEATURE: Also collapse gifs

nattsw:
and make sure they don't stretch past the chat area's width

a

### DIFF
--- a/assets/javascripts/discourse/components/chat-message-collapser.js
+++ b/assets/javascripts/discourse/components/chat-message-collapser.js
@@ -8,14 +8,10 @@ export default Component.extend({
   collapsed: false,
   uploads: null,
   cooked: null,
-
-  @computed("cooked")
-  get title() {
-    return domFromString(this.cooked).dataset.youtubeTitle;
-  },
+  message: null,
 
   @computed("uploads")
-  get filename() {
+  get imageFilename() {
     if (this.uploads) {
       if (this.uploads.length === 1) {
         return this.uploads[0].original_filename;
@@ -26,7 +22,12 @@ export default Component.extend({
   },
 
   @computed("cooked")
-  get link() {
+  get youtubeTitle() {
+    return domFromString(this.cooked).dataset.youtubeTitle;
+  },
+
+  @computed("cooked")
+  get youtubeLink() {
     const id = domFromString(this.cooked).dataset.youtubeId;
     return `https://www.youtube.com/watch?v=${escape(id)}`;
   },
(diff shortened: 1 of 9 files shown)


jjaffeux (approved):

assets/javascripts/discourse/components/chat-message-collapser.js
```diff
@@ -40,4 +41,38 @@ export default Component.extend({
...
+  get isAnimatedImage() {
+    return isAnimatedImage(this.cooked, this.message);
+  },
 });
+
+function isYoutube(cooked) {
+  return /^<div class="onebox lazyYT lazyYT-container"/.test(cooked);
+}
+
+function isAnimatedImage(cooked, message) {
```
pmusaraj, line 65: This is slightly misnamed, you're targeting animated onebox images here. I think it might be good to change this to `isOneboxedImage` and target all `img` elements with a onebox class, and let them be collapsible.

assets/javascripts/discourse/components/chat-message-collapser.js
```diff
@@ -40,4 +41,38 @@ export default Component.extend({
...
+  const onebox = `<p><img src="${escape(message)}" class="animated onebox"`;
+  return cooked.startsWith(onebox);
+}
+
+function isImage(uploads) {
+  return uploads?.length > 0;
```
pmusaraj, line 71: What happens if the uploads here are PDFs or other non-image documents?

assets/javascripts/discourse/components/chat-message-collapser.js
pmusaraj: Also, nitpick, but you likely don't need to pass the message, instead, like with YouTube above you can search for an img tag and a onebox class.

assets/javascripts/discourse/components/chat-message-collapser.js
nattsw: As of before-this-commit, they show up like that and are available for download:

Screenshot 2022-01-10 at 10 19 17 PM

assets/javascripts/discourse/components/chat-message-collapser.js
nattsw: Actually for this commit, I contemplated using `message` for the youtube link instead of deriving it from the onebox :|

assets/javascripts/discourse/components/chat-message-collapser.js
pmusaraj: Cool! So the only tiny comment then is that `isImage` is a misnomer.

assets/javascripts/discourse/components/chat-message-collapser.js
nattsw: > target all `img`

I think that's good for the next PR. I just tried with https://i.guim.co.uk/img/media/c5e73ed8e8325d7e79babf8f1ebbd9adc0d95409/2_5_1754_1053/master/1754.jpg?width=465&quality=45&auto=format&fit=max&dpr=2&s=065f279099ded1062688e357b155dc29 and it doesn't show up properly

assets/javascripts/discourse/components/chat-message-collapser.js
nattsw: Your comments have been addressed here https://github.com/discourse/discourse-chat/pull/504

PR #491, opened 2022-01-10
REFACTOR: tc-avatar

jjaffeux:
- chat-user-avatar, with tests (previously tc-avatar)
- chat-emoji-avatar, with test
- chat-user-presence-flair, tested in isolation (previously tc-user-presence-flair)
- fixes an issue where html for tc-user-presence-flair was also present for each avatar even if not online
- displays presence indicator for reply messages
- removes a lot duplicated css

### DIFF
--- a/assets/javascripts/discourse/components/chat-user-avatar.js
+++ b/assets/javascripts/discourse/components/chat-user-avatar.js
@@ -3,13 +3,16 @@ import discourseComputed from "discourse-common/utils/decorators";
 import { inject as service } from "@ember/service";
 
 export default Component.extend({
-  tagName: "div",
-  classNames: ["tc-presence-flair"],
-  classNameBindings: ["online"],
+  tagName: "",
+
   chat: service(),
 
+  user: null,
+
+  avatarSize: "tiny",
+
   @discourseComputed("chat.presenceChannel.users.[]", "user.id")
-  online(users, userId) {
-    return !!users?.find((u) => u.id === userId);
+  isOnline(users, userId) {
+    return !!users?.findBy("id", userId);
   },
 });
--- a/assets/javascripts/discourse/templates/components/chat-emoji-avatar.hbs
+++ b/assets/javascripts/discourse/templates/components/chat-emoji-avatar.hbs
@@ -1,0 +1,5 @@
+<div class="chat-emoji-avatar">
+  <div class="chat-emoji-avatar-container">
+    {{replace-emoji emoji}}
+  </div>
+</div>
--- a/assets/javascripts/discourse/templates/components/chat-message.hbs
+++ b/assets/javascripts/discourse/templates/components/chat-message.hbs
@@ -81,13 +81,11 @@
         {{#if message.in_reply_to}}
           <div role="button" onclick={{action "viewReply"}} class="tc-reply-display">
             {{d-icon "share" title="chat.in_reply_to"}}
-            <span class="tc-reply-av">
-              {{#if message.in_reply_to.chat_webhook_event.emoji}}
-                {{replace-emoji message.in_reply_to.chat_webhook_event.emoji}}
-              {{else}}
-                {{avatar message.in_reply_to.user imageSize="tiny"}}
-              {{/if}}
-            </span>
+            {{#if message.in_reply_to.chat_webhook_event.emoji}}
+              {{chat-emoji-avatar emoji=message.in_reply_to.chat_webhook_event.emoji}}
+            {{else}}
+              {{chat-user-avatar user=message.in_reply_to.user}}
+            {{/if}}
             <span class="tc-reply-msg">
               {{replace-emoji message.in_reply_to.excerpt}}
             </span>
@@ -98,18 +96,9 @@
           {{format-chat-date message details "tiny"}}
         {{else}}
           {{#if message.chat_webhook_event.emoji}}
-            <div class="tc-avatar">
-              <div class="tc-avatar-container">
-                {{replace-emoji message.chat_webhook_event.emoji}}
-              </div>
-            </div>
+            {{chat-emoji-avatar emoji=message.chat_webhook_event.emoji}}
           {{else}}
-            <div class="tc-avatar">
-              <div class="tc-avatar-container" data-user-card={{message.user.username}}>
-                {{avatar message.user imageSize="medium"}}
-                {{tc-user-presence-flair user=message.user}}
-              </div>
-            </div>
+            {{chat-user-avatar user=message.user avatarSize="medium"}}
           {{/if}}
         {{/if}}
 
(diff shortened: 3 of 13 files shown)


assets/javascripts/discourse/components/chat-user-avatar.js
```diff
@@ -3,13 +3,16 @@ import discourseComputed from "discourse-common/utils/decorators";
 import { inject as service } from "@ember/service";
 
 export default Component.extend({
-  tagName: "div",
-  classNames: ["tc-presence-flair"],
-  classNameBindings: ["online"],
+  tagName: "",
```
nattsw, line 6: If we skip setting `tagName` entirely and just insert classNames: "chat-user-avatar" and remove the first div in the template, it's the same thing right?

assets/javascripts/discourse/components/chat-user-avatar.js
```diff
@@ -3,13 +3,16 @@ import discourseComputed from "discourse-common/utils/decorators";
...
 
+  user: null,
+
+  avatarSize: "tiny",
+
   @discourseComputed("chat.presenceChannel.users.[]", "user.id")
-  online(users, userId) {
-    return !!users?.find((u) => u.id === userId);
+  isOnline(users, userId) {
+    return !!users?.findBy("id", userId);
```
nattsw, line 16: Ah interesting, did not know about [this](https://api.emberjs.com/ember/release/classes/EmberArray/methods/mapBy?anchor=findBy)

assets/javascripts/discourse/templates/components/chat-emoji-avatar.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="chat-emoji-avatar">
```
nattsw, line 1: What's a "chat-emoji-avatar"? I mean, the words make sense separately to me but is it an avatar that's actually an emoji? Being a little pedantic here, since we're renaming things

assets/javascripts/discourse/templates/components/chat-message.hbs
```diff
@@ -81,13 +81,11 @@
         {{#if message.in_reply_to}}
           <div role="button" onclick={{action "viewReply"}} class="tc-reply-display">
```
nattsw, line 82: Will you be removing this `tc-` too?

assets/javascripts/discourse/templates/components/chat-message.hbs
```diff
@@ -81,13 +81,11 @@
...
-              {{else}}
-                {{avatar message.in_reply_to.user imageSize="tiny"}}
-              {{/if}}
-            </span>
+            {{#if message.in_reply_to.chat_webhook_event.emoji}}
+              {{chat-emoji-avatar emoji=message.in_reply_to.chat_webhook_event.emoji}}
+            {{else}}
+              {{chat-user-avatar user=message.in_reply_to.user}}
+            {{/if}}
             <span class="tc-reply-msg">
```
nattsw, line 89: And this (`tc-`)

test/javascripts/components/chat-emoji-avatar-test.js
```diff
@@ -0,0 +1,28 @@
...
+} from "discourse/tests/helpers/component-test";
+import { discourseModule, exists } from "discourse/tests/helpers/qunit-helpers";
+import hbs from "htmlbars-inline-precompile";
+
+discourseModule(
+  "Discourse Chat | Component | chat-emoji-avatar",
+  function (hooks) {
+    setupRenderingTest(hooks);
+
+    componentTest("uses an emoji as avatar", {
```
nattsw, line 12: Ah the question is answered. I didn't know we can use an emoji as avatar.

assets/javascripts/discourse/templates/components/chat-emoji-avatar.hbs
nattsw: Self-answered below

nattsw (changes requested):
Hmm maybe we can do away with the style reduction in another PR:

Before:
Screenshot 2022-01-11 at 5 37 16 PM

After:
Screenshot 2022-01-11 at 5 37 53 PM

assets/javascripts/discourse/templates/components/chat-message.hbs
jjaffeux: this is out of scope of this PR yes

assets/javascripts/discourse/templates/components/chat-message.hbs
jjaffeux: this is out of scope of this PR yes

jjaffeux:
> Hmm maybe we can do away with the style reduction in another PR:
> 
> Before: Screenshot 2022-01-11 at 5 37 16 PM
> 
> After: Screenshot 2022-01-11 at 5 37 53 PM

Should be good now
Screenshot 2022-01-11 at 11 10 42

jjaffeux:
> > Hmm maybe we can do away with the style reduction in another PR:
> > Before: Screenshot 2022-01-11 at 5 37 16 PM
> > After: Screenshot 2022-01-11 at 5 37 53 PM
> 
> Should be good now Screenshot 2022-01-11 at 11 10 42

Al

> > Hmm maybe we can do away with the style reduction in another PR:
> > Before: Screenshot 2022-01-11 at 5 37 16 PM
> > After: Screenshot 2022-01-11 at 5 37 53 PM
> 
> Should be good now Screenshot 2022-01-11 at 11 10 42
> 
> Although Im trying to think of something better here, problem is we want to have a fixed with for this, but image size is not fixed, so final visible margin will vary...

Ok I think last commit is better:

Screenshot 2022-01-11 at 11 21 26

assets/javascripts/discourse/components/chat-user-avatar.js
jjaffeux: BTW this also means no `click(event) {}`, attribute bindings, classname bindings...

assets/stylesheets/common/chat-message.scss
```diff
@@ -115,9 +110,8 @@
       }
     }
 
-    .tc-avatar {
+    .chat-user-avatar {
       grid-area: avatar;
-      width: 100%;
     }
 
     .chat-message-content {
```
nattsw, line 117: Testing locally, I removed these two  (.chat-user-avatar and .chat-message-content) styles and nothing changed 🤔

jjaffeux:
> How much do you care about this difference? (ignore the up and down movement because of different scroll heights, what I mean is the width between the avatar and the username)
> 
>  Screen.Recording.2022-01-11.at.8.39.51.PM.mov

Yep I think it's fine, I mostly want this to be easier to change for designers. And at least it's consistent now.

nattsw (approved):
woohoo thanks for doing this

assets/stylesheets/common/chat-message.scss
jjaffeux: Oooops I saw this after merging, feel free to try to slim down the css if you want to

PR #495, opened 2022-01-11
UI: attempts to simplify css for chat channel rows

jjaffeux:
- ensures we don't have a background under topic bad icon
- removes background under channel leave button to make it easier to work with different height rows
- ensures icon color on hover is consistent
- moves most sidebar extensions to a specific css file

### DIFF
--- a/assets/stylesheets/common/common.scss
+++ b/assets/stylesheets/common/common.scss
@@ -72,14 +72,6 @@ $float-height: 530px;
   }
 }
 
-.has-sidebar {
-  .header-dropdown-toggle.open-chat {
-    .chat-channel-unread-indicator {
-      border-color: var(--primary-very-low);
-    }
-  }
-}
-
 body.composer-open .topic-chat-float-container {
   bottom: 11px; // prevent height of grippie from obscuring ...is typing indicator
 }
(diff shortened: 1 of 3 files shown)


jjaffeux:
Before (sidebar/without sidebar):
Screenshot 2022-01-11 at 10 38 37 
Screenshot 2022-01-11 at 10 39 32

After (sidebar/without sidebar):
Screenshot 2022-01-11 at 10 22 35
Screenshot 2022-01-11 at 10 29 45

ZogStriP (approved):

CvX:
> ensures icon color on hover is consistent

@jjaffeux in dark mode (no sidebar) "x" icon next to personal channels and "+" next to "personal chat" header no longer change on hover

jjaffeux:
Yes, is that an issue ?

CvX:
Yeah. We usually have on-hover elements on all interactive elements. And especially here, there's no indication that "x" icon does anything since the whole row is interactive and highlights on hover. And it is inconsistent with the pencil icon next to "channels" header which does have on-hover effect.

jjaffeux:
Yes agreed, I thought after that we had actually two actions on this row, and that we needed this visual difference.

PR #496, opened 2022-01-11
FIX: attempts to avoid duplicated channels tracking

jjaffeux:


jjaffeux:
I sadly couldn't repro this locally, it's an attempt to:
- fix one possible issue
- add a safety net

assets/javascripts/discourse/services/chat.js
```diff
@@ -172,12 +172,12 @@ export default Service.extend({
       }
 
       if (!this._fetchingChannels) {
```
jjaffeux, line 174: This whole code block was suspicious, even though it's very probably that refreshChannels won't be done before we reach next line, it's still not a clean flow.

davidtaylorhq (approved):

PR #498, opened 2022-01-11
FIX: DOM operations should be in afterRender not next

jjaffeux:


assets/javascripts/discourse/components/full-page-chat.js
```diff
@@ -57,12 +60,14 @@ export default Component.extend({
...
-      sidebarScroll.scrollTop = sidebarScroll.scrollHeight;
-    }
+    schedule("afterRender", () => {
+      const sidebarScroll = document.querySelector(
+        ".sidebar-container .scroll-wrapper"
+      );
+      if (sidebarScroll) {
+        sidebarScroll.scrollTop = sidebarScroll.scrollHeight;
+      }
+    });
```
CvX, line 70: This one is called from `didInsertElement`, so maybe there's no need to delay it?

jjaffeux: yep I'm actually doing few tests on this, agreed it's probably overkill

jjaffeux: @CvX look at last changes please, I think that should actually be enough

CvX (approved):
Looks good! The PR title and commit message need to be updated though 😃

PR #502, opened 2022-01-11
FEATURE: Sync drafts with server

nbianca:


plugin.rb
```diff
@@ -186,6 +186,21 @@ def self.allowed_group_ids
     include_has_chat_enabled? && object.user_option.chat_sound
   end
 
+  add_to_serializer(:current_user, :chat_drafts) do
+    Draft
+      .select(:draft_key, :data)
```
eviltrout, line 191: `pluck` would be slightly faster here. Since this query happens on the current_user serializer we should consider it.

assets/javascripts/discourse/services/chat.js
```diff
@@ -59,7 +60,10 @@ export default Service.extend({
       this._subscribeToChannelEdits();
       this.appEvents.on("page:changed", this, "_storeLastNonChatRouteInfo");
       this.presenceChannel = this.presence.getChannel("/chat/online");
-      this._draftStore = new KeyValueStore(DRAFT_STORE_NAMESPACE);
```
eviltrout, line 62: There's a lint failure due to `KeyValueStore` still being imported.

eviltrout (approved):

SamSaffron:
Can we clear the draft in ChatMessageCreator?

PR #510, opened 2022-01-13
FIX: correctly computes channels max height in float panel

jjaffeux:
Before this change it would be possible to have last channel unreachable as the container would not scroll.

### DIFF
--- a/assets/stylesheets/common/common.scss
+++ b/assets/stylesheets/common/common.scss
@@ -237,7 +237,7 @@ body.composer-open .topic-chat-float-container {
 
 .topic-chat-drawer-content {
   box-sizing: border-box;
-  height: 100%;
+  height: calc(100% - #{$header-height});
   padding-bottom: 0.25em;
 
   .chat-channels .chat-channel-divider {


assets/stylesheets/common/common.scss
```diff
@@ -237,7 +237,7 @@ body.composer-open .topic-chat-float-container {
 
 .topic-chat-drawer-content {
   box-sizing: border-box;
-  height: 100%;
+  height: calc(100% - #{$header-height});
```
davidtaylorhq, line 240: I think `var(--header-offset)` might be the more modern version of this?

jjaffeux: > header-offset

no in this case $header-height is a local variable of the file, related to height of the drawer's header and not the discourse core header.

davidtaylorhq: Ohhhh I see, sounds good :+1:

davidtaylorhq (approved):

assets/stylesheets/common/common.scss
jjaffeux: https://github.com/discourse/discourse-chat/blob/47ceb26434cd00e35c34aca564360083afe7ba8a/assets/stylesheets/common/common.scss#L1

PR #512, opened 2022-01-13
FIX: ensures we don't keep dots in last part of channel URL

jjaffeux:


jjaffeux:
encoreURIComponent wouldn't convert it too, so AFAIK this is the only solution for this specific issue

PR #519, opened 2022-01-14
FIX: ensures sidebar doesn’t override leave-btn direct hover style

jjaffeux:


jjaffeux:
Before:

Screenshot 2022-01-14 at 15 05 39

After:
Screenshot 2022-01-14 at 15 05 45

nbianca (approved):

PR #521, opened 2022-01-16
UX: Various mobile improvements

CvX:
* UX: Remove mobile background color
* UX: Remove hover style from mobile chat list
* UX: Add safe-area padding
* UX: Remove alert-info spacing also on mobile
* DEV: Remove an unused grid-area


jjaffeux:
Hard to review so many small changes from code but I see nothing wrong. The important is most likely needed, but maybe we could try to follow up to avoid having to do this?

jjaffeux (approved):

CvX:
Just to clarify, this PR didn't introduce that `!important` line, just merged two blocks, hence the diff. As to it being required – it's here to override core styling, so it's probably needed.

PR #522, opened 2022-01-16
FIX: attempts to make scrolling github more resilient

jjaffeux:
It now tries to find the first parent scrollable element from a selected line.

It works nicely in my tests:
Screenshot 2022-01-16 at 22 16 00

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -1030,34 +1030,38 @@ export default Component.extend({
     decorateGithubOneboxBody(this.element);
   },
 
+  _getScrollParent(node, maxParentSelector) {
+    if (node === null || node.classList.contains(maxParentSelector)) {
+      return null;
+    }
+
+    if (node.scrollHeight > node.clientHeight) {
+      return node;
+    } else {
+      return this._getScrollParent(node.parentNode, maxParentSelector);
+    }
+  },
+
   _scrollGithubOneboxes() {
     this.element
       .querySelectorAll(".onebox.githubblob li.selected")
       .forEach((line) => {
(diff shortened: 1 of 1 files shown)


jjaffeux:
@davidtaylorhq can you think of any issue with this solution?

CvX (approved):

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1030,34 +1030,38 @@ export default Component.extend({
...
-        // Keep iterating up them until we reach the topmost one within the onebox
-        // Sum up the offsetTop values as we go
-        let offsetParent = line.offsetParent;
-        while (true) {
-          linePosition += offsetParent.offsetTop;
-          if (!offsetParent.offsetParent?.closest(scrollingElementSelector)) {
-            // The next offsetParent would be outside the onebox
-            break;
-          }
-        }
```
davidtaylorhq, line 1051: This logic handles the case where you have nested `position: relative` elements. For example

```
<div style='overflow: scroll'>
  <h1>Some element which takes up height</h1>
  <div style='position: relative'>
    <code>...</code>
  </div>
</div>
```
In this scenario, `querySelector('code').offsetHeight` would be its position within the `position: relative` div, **not** its position within the scroll wrapper.

If the GitHub onebox doesn't have any `position: relative` wrappers inside it, then I guess it's fine to drop this logic. But it might come back to bite us in future if core/themes adjust the CSS of the Onebox.

jjaffeux: Note as this was discussed on another channel:
- code is now the scrollable element this is why it got broken in the first place
- agreed it's not going to be resilient to any big change anyways

PR #523, opened 2022-01-17
FIX: Use Page Visibility API for determining whether to read

markvanlan:
Using this api https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API solves the issue of switching tabs and messages not being marked as read, because `document.hasFocus` is still false.

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -699,7 +699,12 @@ export default Component.extend({
   },
 
   _floatOpenAndFocused() {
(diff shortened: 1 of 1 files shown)


davidtaylorhq:
I *think* the `visibilityState` is still 'visible', even when a **window** (not a tab) is in the background... and i don't think we want to mark messages read if they arrive in a background window.

I'd suggest using the `userPresent` API from core's `lib/user-presence` library here. It has a few smarts to take care of these kinds of issues.

davidtaylorhq (approved):

PR #524, opened 2022-01-17
FIX: ensures textarea is resized upon entering a new channel

jjaffeux:
Before this change the textarea was only resized on initial insert of the element or typing, which would mean that typing multiple lines in one channel would increase height of composer, and switching to another channel would keep this height.

Note: this behavior was only visible with the https://meta.discourse.org/t/experiment-on-meta-horizontal-loading-slider/177939 theme component


markvanlan (approved):

jjaffeux:
Note that at the moment, the focus/resizing logic is kinda mixed up and we might want to refactor this in only one place, especially given we also have different behavior for mobile.

PR #525, opened 2022-01-17
FIX: selecting messages alignment was broken

jjaffeux:


jjaffeux:
That should fix it.

Before: 
Screenshot 2022-01-17 at 16 47 48


After:
Screenshot 2022-01-17 at 16 47 40


Note that there's a huge opportunity for refactoring here, as it's a deep nesting of class-less divs which are adding a lot of content to the page and bringing no value.

markvanlan (approved):

PR #527, opened 2022-01-17
FEATURE: Ability to mention groups

markvanlan:
Currently group mentions do not work. This PR changes that :)

Relies on core PR for notification type to be added https://github.com/discourse/discourse/pull/15616

Screen Shot 2022-01-17 at 3 13 21 PM

### DIFF
--- a/lib/chat_notifier.rb
+++ b/lib/chat_notifier.rb
@@ -80,36 +80,67 @@ class DiscourseChat::ChatNotifier
     Jobs.enqueue_in(3.seconds, :create_chat_mention_notifications, {
       chat_message_id: @chat_message.id,
       user_ids: user_ids,
+      user_ids_to_group_mention_map: user_ids_to_group_mention_map,
       timestamp: @timestamp
     })
   end
 
-  def cooked_mentions
-    @cooked_mentions ||= Nokogiri::HTML5.fragment(@chat_message.cooked).css(".mention").map(&:text)
+  def direct_mentions_from_cooked
+    @direct_mentions_from_cooked ||= Nokogiri::HTML5.fragment(@chat_message.cooked).css(".mention").map(&:text)
+  end
+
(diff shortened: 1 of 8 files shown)


lib/chat_notifier.rb
```diff
@@ -80,36 +80,67 @@ def enqueue_mentioned_job(user_ids)
...
+                               end
+  end
+
+  def user_ids_to_group_mention_map
+    map = {}
+    group_mentioned_users.each do |user|
+      group_name = (user.groups.map(&:name) & group_name_mentions).first
+      map[user.id] = group_name
+    end
+    map
```
markvanlan, line 113: The method looks funky, but it avoids doing 1 query for each group mentioned; groups are preloaded in the user query so we can loop through users from 1 query, and get the matching group name mentioned.

spec/components/chat_message_creator_spec.rb
```diff
@@ -208,6 +210,80 @@
...
+      end
+
+      it "creates chat mentions for group mentions and direct mentions" do
+        expect {
+          DiscourseChat::ChatMessageCreator.create(
+            chat_channel: public_chat_channel,
+            user: user1,
+            content: "hello @#{admin_group.name} @#{user_group.name}"
+          )
+        }.to change { admin1.chat_mentions.count }.by(1)
```
eviltrout, line 263: Count tests are not my favorite, although I understand at times they are necessary. If there's any way to peek in the notifications themselves and looking for a particular identifier that would be preferred.

eviltrout (approved):

spec/components/chat_message_creator_spec.rb
markvanlan: I'm not sure what else I'd look for. The important part is that X user gets a chat mention 🤔

PR #528, opened 2022-01-18
FIX: Show lightbox after collapse and expand

nattsw:
The lightboxing function is applied once when [messages are loaded](https://github.com/discourse/discourse-chat/blob/3eabe03afa73418565d7f336f207ca3b2702ed07/assets/javascripts/discourse/components/chat-live-pane.js#L477). Since collapsing an image removed it from the document entirely, the lightboxing feature wasn't working after a user collapses and expands the image. 

To prevent this we could either A: reapply the lightboxing function every time after expand, or B: just skip removing the image when it's collapsed and just hide it. I chose B.

### DIFF
--- a/test/javascripts/acceptance/chat-live-pane-collapse-test.js
+++ b/test/javascripts/acceptance/chat-live-pane-collapse-test.js
@@ -95,7 +95,7 @@ acceptance("Discourse Chat - Chat live pane", function (needs) {
   });
 
   test("can collapse and expand youtube chat", async function (assert) {
-    const youtubeContainerSelector = ".chat-message-container-1 .lazyYT";
+    const youtubeContainer = ".chat-message-container-1 .lazyYT";
     const expandImage =
       ".chat-message-container-1 .chat-message-collapser-closed";
     const collapseImage =
@@ -103,19 +103,19 @@ acceptance("Discourse Chat - Chat live pane", function (needs) {
 
     await visit("/chat/channel/1/cat");
 
-    assert.ok(visible(youtubeContainerSelector));
+    assert.ok(visible(youtubeContainer));
     assert.ok(visible(collapseImage), "the open arrow is shown");
     assert.notOk(exists(expandImage), "the close arrow is hidden");
 
     await click(collapseImage);
 
-    assert.notOk(visible(youtubeContainerSelector));
+    assert.notOk(visible(youtubeContainer));
     assert.ok(visible(expandImage), "the close arrow is shown");
     assert.notOk(exists(collapseImage), "the open arrow is hidden");
 
     await click(expandImage);
 
-    assert.ok(visible(youtubeContainerSelector));
+    assert.ok(visible(youtubeContainer));
     assert.ok(visible(collapseImage), "the open arrow is shown again");
     assert.notOk(exists(expandImage), "the close arrow is hidden again");
   });
--- a/test/javascripts/acceptance/chat-live-pane-collapse-test.js
+++ b/test/javascripts/acceptance/chat-live-pane-collapse-test.js
@@ -38,6 +38,34 @@ acceptance("Discourse Chat - Chat live pane", function (needs) {
               username: "tomtom",
             },
           },
+          {
+            id: 2,
+            message: "",
+            cooked: "",
+            excerpt: "",
+            uploads: [
+              {
+                id: 4,
+                url: "/images/avatar.png",
+                original_filename: "tomtom.jpeg",
+                filesize: 93815,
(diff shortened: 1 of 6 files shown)


test/javascripts/acceptance/chat-live-pane-collapse-test.js
```diff
@@ -52,29 +80,72 @@ acceptance("Discourse Chat - Chat live pane", function (needs) {
...
+      ])
+    );
   });
 
   test("can collapse and expand youtube chat", async function (assert) {
-    const youtubeContainerSelector = ".lazyYT";
-    const close = ".chat-message-collapser-closed";
-    const open = ".chat-message-collapser-opened";
+    const youtubeContainerSelector = ".chat-message-container-1 .lazyYT";
+    const expandImage =
```
tgxworld, line 99:
```suggestion
    const expandImageSelector =
```

I'm wondering if this is clearer and will be more consistent with the variable above.

test/javascripts/acceptance/chat-live-pane-collapse-test.js
```diff
@@ -52,29 +80,72 @@ acceptance("Discourse Chat - Chat live pane", function (needs) {
...
+    assert.notOk(exists(expandImage), "the close arrow is hidden again");
+  });
+
+  test("lightbox shows up before and after expand and collapse", async function (assert) {
+    const lightboxImage = ".mfp-img";
+    const image = ".chat-message-container-2 .chat-img-upload";
+    const expandImage =
+      ".chat-message-container-2 .chat-message-collapser-closed";
+    const collapseImage =
+      ".chat-message-container-2 .chat-message-collapser-opened";
```
tgxworld, line 129: Same comment as https://github.com/discourse/discourse-chat/pull/528/files#r786421329.

test/javascripts/acceptance/chat-live-pane-collapse-test.js
```diff
@@ -52,29 +80,72 @@ acceptance("Discourse Chat - Chat live pane", function (needs) {
...
+
+    await visit("/chat/channel/1/cat");
+
+    await click(image);
+
+    assert.ok(
+      exists(document.querySelector(lightboxImage)),
```
tgxworld, line 136:
```
import { query } from "discourse/tests/helpers/qunit-helpers";

query(lightboxImageSelector)
```

We have a helper for this.

test/javascripts/acceptance/chat-live-pane-collapse-test.js
```diff
@@ -52,29 +80,72 @@ acceptance("Discourse Chat - Chat live pane", function (needs) {
...
+
+    await click(collapseImage);
+    await click(expandImage);
+
+    await click(image);
+    assert.ok(
+      exists(document.querySelector(lightboxImage)),
+      "can see lightbox after collapse expand"
+    );
+    await click(document.querySelector(".mfp-container"));
```
tgxworld, line 149: I think this can just be `click(".mfp-container")`?

nattsw: `click(".mfp-container")` was the original but it didn't work I suspect due to how the overlay is rendered in the tests. To find the container we need to use the document.

test/javascripts/acceptance/chat-live-pane-collapse-test.js
nattsw: Same comment as https://github.com/discourse/discourse-chat/pull/528#discussion_r786429593.

test/javascripts/acceptance/chat-live-pane-collapse-test.js
tgxworld: ahh icic. Maybe it is some how rendered outside `#ember-testing`?

test/javascripts/acceptance/chat-live-pane-collapse-test.js
nattsw: Removed the `selector` here so I don't have to change the rest

tgxworld (approved):

test/javascripts/acceptance/chat-live-pane-collapse-test.js
nattsw: I suspect that's the case. Looking at core it seems it has to be done this way too.

nattsw: I told myself to add a comment about this so people won't get confused when I create my PR ... but I forgot. T_T

PR #529, opened 2022-01-18
UX: ensures avatar and username are clickable

jjaffeux:


nattsw:
What does making them clickable do?

jjaffeux:
> What does making them clickable do? Any side effects to test?

Not really this is just adding `cursor: pointer` through the clickable css class. Rest of the behavior is our super classic user card, not much to be worried about. Also it would probably require an acceptance test given the state of chat-message atm, so would be very heavy for something like this.

ZogStriP (approved):

PR #531, opened 2022-01-18
PERF: attempts to build only one expensive query instead of two

jjaffeux:


jjaffeux:
How do we feel about this @cvx @markvanlan ?

app/controllers/chat_controller.rb
```diff
@@ -242,31 +242,32 @@ def flag
...
+      .limit(40)
 
     # .with_deleted if include_deleted
     future_messages = base_query
       .where("created_at > ?", @message.created_at)
       .order(created_at: :asc)
 
-    messages = [past_messages.reverse, [@message], future_messages].reduce([], :concat)
+    messages = ChatMessage
+      .where(id: [past_messages.reverse.pluck(:id), @message.id, future_messages.pluck(:id)].flatten)
```
CvX, line 262:
```suggestion
      .where(id: [past_messages.pluck(:id).reverse, @message.id, future_messages.pluck(:id)].flatten)
```

right?

app/controllers/chat_controller.rb
```diff
@@ -242,31 +242,32 @@ def flag
 
   def lookup_message
     chat_channel = @message.chat_channel
-    chatable = nil
```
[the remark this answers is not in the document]
markvanlan: This was before when there were channels without chatables, and we would have to set that in to `nil` to avoid an error. Not needed anymore.

app/controllers/chat_controller.rb
```diff
@@ -242,31 +242,32 @@ def flag
...
+
     past_messages = base_query
       .where("created_at < ?", @message.created_at)
-      .order(created_at: :desc).limit(40)
+      .order(created_at: :desc)
+      .limit(40)
 
     # .with_deleted if include_deleted
     future_messages = base_query
       .where("created_at > ?", @message.created_at)
```
markvanlan, line 258: I can't remember the reasoning for using `created_at` instead of `id` for this - couldn't we grab any messages after using an ID greater than the `@message`? I assume that would be a PERF improvement

jjaffeux: I would prefer to investigate this in a different PR

CvX (approved):

PR #537, opened 2022-01-18
UX: Add subtle background color to widget header

markvanlan:
Spoke with @jordanvidrine about improving the sexiness of chat. I like this little change personally, and he agreed.

## NEW
Screen Shot 2022-01-18 at 3 50 08 PM

## OLD
Screen Shot 2022-01-18 at 3 55 59 PM

### DIFF
--- a/assets/stylesheets/common/common.scss
+++ b/assets/stylesheets/common/common.scss
@@ -153,6 +153,8 @@ body.composer-open .topic-chat-float-container {
 .topic-chat-drawer-header {
   box-sizing: border-box;
   border-bottom: solid 1px var(--primary-low-mid);
+  border-radius: 8px 8px 0 0;
+  background: var(--primary-very-low);
   padding: 0.5em;
   height: $header-height;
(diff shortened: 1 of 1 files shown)


jordanvidrine (approved):

jordanvidrine:
I have added some additional zazz and lightened up some of the border color of the chat widget, and modernized the chat input as well.

### After
image

image

CvX:
I'm not sure how I feel about low-contrast non-core-like composer…

jordanvidrine:
> I'm not sure how I feel about low-contrast non-core-like composer…

It's pretty much the same as Discord's chat style so I dont *think* the contrast should be an issue, especially once you type a message. 

image

I like it much more than the current box, and I feel it sets the chat apart from Discourse without being too complicated or different.

pmusaraj (approved):
Let's try it out. I personally use one of our dark themes, so I'm a bit curious what it would look like there.

PR #538, opened 2022-01-19
FIX: Add missing markdown features

martin-brennan:
In e0a4a48d3cba6457a593dd43e3d7f14577c6f3d0 we limited chat to a
few features of markdown, but we missed some commonly used ones.
This commit adds the following back in:

* `[spoiler]spoiler tags[/spoiler]` from discourse-spoiler-alert
* markdown lists
* autolink for links like `<https://somelink.com>` to avoid inline
  oneboxing those links
* emoji shortcuts (:P -> :stuck_out_tongue:) and inline emoji for
  one emoji making up the whole chat message

The test for spoilers will not run unless you have
discourse-spoiler-alert installed and enabled.

### DIFF
--- a/app/models/chat_message.rb
+++ b/app/models/chat_message.rb
@@ -43,6 +43,8 @@ class ChatMessage < ActiveRecord::Base
     censored
     discourse-local-dates
     emoji
+    emojiShortcuts
+    inlineEmoji
     html-img
     mentions
     onebox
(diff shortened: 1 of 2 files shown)


martin-brennan:
Really wasn't sure how I was supposed to write a test for spoilers since that's a separate plugin, so I just guarded it behind a site setting.

nattsw (approved):

PR #544, opened 2022-01-20
FEATURE: Chat transcripts using [chat] custom markdown BBCode

martin-brennan:
This commit introduces the new `[chat]` BBCode block, which will be used for quoting chat messages into regular posts. The chat messages quoted inside `[chat]` will be rendered with the same limited subset of BBCode defined in `ChatMessage::MARKDOWN_FEATURES` and `ChatMessage::MARKDOWN_IT_RULES` as the server-side chat messages are cooked with. This is done via the new `requestCustomMarkdownCookFunction` API in core.

The `[chat]` block requires a `quote` attribute in the following format: `quote="username;message_id;datetime"`. It allows for an optional `channel` attribute which should be the channel name, and this should only be used sparingly. Here is a complete example:

```
[chat quote="martin;2345;2022-01-25T05:40:39Z" channel="Lounge"]
This is my cool message!
[chat]
```

This is intended to be used as a single message per `[chat]` block, except for in the case of multiple subsequent messages from the same user. Most of the time this BBCode will be autogenerated, so the ergonomics of how it can be typed in are of a lower priority.

The formatted datetime is rendered using the `decorateCookedElement` plugin API for each post.

image

-----

Note: This won't pass tests until https://github.com/discourse/discourse/pull/15753 is merged


assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
```diff
@@ -0,0 +1,159 @@
...
+    let datetimeDivToken = state.push("div_chat_transcript_datetime", "div", 1);
+    datetimeDivToken.attrs = [["class", "chat-transcript-datetime"]];
+
+    let linkToken = state.push("link_open", "a", 1);
+    linkToken.attrs = [
+      ["href", options.getURL(`/chat/message/${messageIdStart}`)],
+    ];
+    linkToken.block = false;
+
+    let datetimeToken = state.push("html_inline", "", 0);
```
martin-brennan, line 86: Update: I realised in my convalescence that I can't do this, it will bake the date + time format server side which is not what we want. Need a cooked post decorator client-side for this instead.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -45,7 +47,30 @@ export default {
         "header-chat-link",
         "chat:rerender-header"
       );
+
       api.addToHeaderIcons("header-chat-link");
+
+      api.decorateCookedElement((elem) => {
```
nattsw, line 53: Btw, not sure how / when this api method gets invoked, does it show up well for chat messages?

The doc says: `Used for decorating the `cooked` content of a post after it is rendered`

martin-brennan: @nattsw I've added a screenshot to the OP, this BBCode is used within a post because we are quoting chat messages there. This method gets invoked for each post in the stream.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -45,7 +47,30 @@ export default {
...
+          const dateTimeRaw = el.dataset["datetime"];
+          const dateTimeLinkEl = el.querySelector(
+            ".chat-transcript-datetime a"
+          );
+
+          if (currentUserTimezone) {
+            dateTimeLinkEl.innerText = moment
+              .tz(dateTimeRaw, currentUserTimezone)
+              .format(I18n.t("dates.long_no_year"));
+          } else {
```
nattsw, line 67: Is this in the acceptance test? Should be easy to include.

assets/javascripts/discourse/initializers/chat-setup.js
nattsw: Got it. You even said `quoting chat messages into regular posts` I should read better.

assets/stylesheets/common/chat-transcript.scss
```diff
@@ -0,0 +1,46 @@
+.discourse-chat-transcript {
```
nattsw, line 1: Did you get a chance to view this in mobile?

config/locales/client.en.yml
```diff
@@ -115,6 +115,9 @@ en:
       composer:
         toggle_toolbar: "Toggle toolbar"
 
+      quote:
+        original_channel: "Originally posted in %{channel}."
```
nattsw, line 119: For terminology's sake: Do we use "post" a chat message? Just thought "post" may imply it's a post. (If you care, "sent" may be less conflicting)

nattsw (approved):
Took a while but overall looks okay, would be nice to confirm about the feel on mobile!

Personally also wondering how we would educate someone how to do that exact quote without hints while drafting a post:

```
[chat quote="martin;2345;2022-01-25T05:40:39Z" channel="Lounge"]
This is my cool message!
[chat]
```

config/locales/client.en.yml
martin-brennan: @nattsw haha you know I just thought that as I was reading the screenshot I posted :joy: "Originally sent" is probably better I will use that :+1:

martin-brennan:
@nattsw it is a good point you have about this:

> Personally also wondering how we would educate someone how to do that exact quote without hints while drafting a post

However, I don't really see this as BBCode that is intended to be typed out. Our "quote chat message" and "archive chat message" and "archive chat channel" tools will be generating these BBCode blocks, so the user will not have to know how to build them.

assets/stylesheets/common/chat-transcript.scss
martin-brennan: No, will make sure it works well there and fix if not

martin-brennan: @nattsw I think this is fine on mobile:

image

assets/javascripts/discourse/initializers/chat-setup.js
martin-brennan: Yep done!

nattsw (approved):

PR #545, opened 2022-01-20
FEATURE: Chat message auto-deletion with site settings

markvanlan:
Note: The serializer only serializes if the user HAS NOT cleared the reminder before. This will cut down on the current_user payload as most users who can chat will clear it.

The reminder also comes back if the site settings change, so that users stay up-to-date on how frequently old messages are removed. 

Site settings
Screen Shot 2022-01-20 at 11 00 22 AM

Public channel channel (only staff will see this)
Screen Shot 2022-01-20 at 11 01 06 AM
Screen Shot 2022-01-20 at 11 01 17 AM

Personal chat channel (Everyone will see this)
Screen Shot 2022-01-20 at 11 01 36 AM
Screen Shot 2022-01-20 at 11 01 28 AM

### DIFF
--- a/app/jobs/scheduled/delete_old_chat_messages.rb
+++ b/app/jobs/scheduled/delete_old_chat_messages.rb
@@ -28,10 +28,7 @@ module Jobs
     end
 
     def valid_day_value?(setting_name)
-      value = SiteSetting.public_send(setting_name)
-      return false if value.nil?
-      return false if value.zero?
-      true
+      (SiteSetting.public_send(setting_name) || 0).positive?
     end
   end
 end
--- a/app/jobs/scheduled/delete_old_chat_messages.rb
+++ b/app/jobs/scheduled/delete_old_chat_messages.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module Jobs
+  class DeleteOldChatMessages < ::Jobs::Scheduled
+    daily at: 0.hours
+
+    def execute(args = {})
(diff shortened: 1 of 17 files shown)


app/jobs/scheduled/delete_old_chat_messages.rb
```diff
@@ -0,0 +1,37 @@
...
+        .in_dm_channel
+        .created_before(SiteSetting.chat_dm_retention_days.days.ago)
+        .destroy_all
+    end
+
+    def valid_day_value?(setting_name)
+      value = SiteSetting.public_send(setting_name)
+      return false if value.nil?
+      return false if value.zero?
+      true
```
eviltrout, line 34:
```suggestion
      (SiteSetting.public_send(setting_name) || 0).positive?
```

assets/javascripts/discourse/components/chat-retention-reminder.js
```diff
@@ -0,0 +1,59 @@
...
+  tagName: "",
+  loading: false,
+
+  @discourseComputed(
+    "chatChannel.chatable_type",
+    "currentUser.{needs_dm_retention_reminder,needs_channel_retention_reminder}"
+  )
+  show(chatableType) {
+    return (
+      (chatableType === "DirectMessageChannel" &&
```
eviltrout, line 18: `"DirectMessageChannel"` should probably be a constant. It's used quite a bit.

eviltrout (approved):

PR #547, opened 2022-01-21
FIX: Move empty state to browse page

markvanlan:
I noticed a bug where topic channels aren't showing up on the browse page. There was also an empty state added to the browse page which made me realize... we have a much better empty state for chat in `chat.index` route that is no longer accessible.

This moves the former index blank page (which includes a DM creator UI) to the browse page.


assets/javascripts/discourse/controllers/chat-index.js
```diff
@@ -1,36 +1,2 @@
...
-  cancelDmCreation() {
-    this.set("creatingDm", false);
-  },
-
-  @action
-  selectChannel(channel) {
-    return this.router.transitionTo("chat.channel", channel.id, channel.title);
-  },
-});
+export default Controller.extend({});
```
eviltrout, line 2:
```suggestion
export default Controller.extend();
```
This is the most trivial thing I've ever suggested

assets/javascripts/discourse/templates/chat-browse.hbs
```diff
@@ -5,15 +5,42 @@
...
+    <div class="empty-state">
+      <span class="empty-state-title">{{i18n "chat.empty_state.title"}}</span>
+      <div class="empty-state-body">
+        {{#if model.length}}
+          {{#each model as |channel|}}
+            {{chat-channel-settings-row channel=channel}}
+          {{/each}}
+        {{else}}
+          <p>{{i18n "chat.empty_state.no_public_available"}}</p>
+        {{/if}}
```
eviltrout, line 28:
```suggestion
        {{#each model as |channel|}}
          {{chat-channel-settings-row channel=channel}}
        {{else}}
          <p>{{i18n "chat.empty_state.no_public_available"}}</p>
        {{/each}}
```
A little known feature of `{{#each}}` is it can take an `{{else}}` for this case!

eviltrout (approved):

assets/javascripts/discourse/controllers/chat-index.js
markvanlan: lol

assets/javascripts/discourse/templates/chat-browse.hbs
markvanlan: wowowow very cool!

PR #554, opened 2022-01-21
DEV: Plugin outlet for falco's voice chat test

markvanlan:


assets/javascripts/discourse/templates/components/channel-list.hbs
```diff
@@ -42,6 +42,7 @@
       {{#each publicChannels as |channel|}}
         {{chat-channel-row channel=channel switchChannel=onSelect}}
       {{/each}}
+      {{plugin-outlet name="below-plugin-chat-channels" tagName="" args=(hash inSidebar=inSidebar)}}
```
CvX, line 45: Should that outlet appear only when there are channels? (i.e. should it be inside that `if` block or below it?)

xfalcox (approved):

assets/javascripts/discourse/templates/components/channel-list.hbs
markvanlan: Good question.

markvanlan: IDK

CvX: Depends on the use-case, I guess. Probably would be better outside the `if`.

markvanlan:
@xfalcox this is an interesting problem; we probably need a more robust API for this. If we put the outlet outside of the `if`, you will see the voice channel AND the channel empty state when the user isn't following any other channels. Inside the `if`, the user will only see the voice channel if they follow at least 1 other public channel

xfalcox:
> @xfalcox this is an interesting problem; we probably need a more robust API for this. If we put the outlet outside of the `if`, you will see the voice channel AND the channel empty state when the user isn't following any other channels. Inside the `if`, the user will only see the voice channel if they follow at least 1 other public channel

Can you move it outside and merge? 

This way we can start with a design where audio channels are a whole block like:

Channels
  - channel A
  - channel B

Voice Rooms
  - channel C
  - channel D


Personal Chat
  - Person A
  - Person B

PR #557, opened 2022-01-24
FEATURE: Blocking messages that contain watched words

markvanlan:
Screen Shot 2022-01-24 at 11 23 19 AM

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -817,7 +817,7 @@ export default Component.extend({
     ) {
       let message = this.messages[messageIndex];
       if (message.user.id === this.currentUser.id && !message.error) {
-        lastUserMessage = this.messages[messageIndex];
+        lastUserMessage = message;
         break;
       }
     }
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -815,7 +815,8 @@ export default Component.extend({
       messageIndex >= 0;
       messageIndex--
     ) {
-      if (this.messages[messageIndex].user.id === this.currentUser.id) {
+      let message = this.messages[messageIndex];
+      if (message.user.id === this.currentUser.id && !message.error) {
         lastUserMessage = this.messages[messageIndex];
         break;
       }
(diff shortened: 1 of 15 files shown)


assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -767,10 +767,10 @@ export default Component.extend({
     return Promise.resolve();
   },
 
-  _onSendError(stagedId) {
+  _onSendError(stagedId, error) {
     const stagedMessage = this.messageLookup[`staged-${stagedId}`];
     if (stagedMessage) {
-      stagedMessage.set("error", true);
+      stagedMessage.set("error", error.jqXHR.responseJSON.errors[0]);
```
markvanlan, line 773: I know it's ugly to do this, but I don't want to use popupError, and want just to text of the error 🤷

lib/chat_message_updater.rb
```diff
@@ -21,21 +21,29 @@ def initialize(chat_message:, new_content:, upload_ids: nil)
...
+      if Rails.env.test?
+        puts "#" * 50
+        puts "Chat message update error:"
+        puts @error.inspect
+        puts "#" * 50
+      end
+    end
+  end
+
+  def validate_message!
```
markvanlan, line 43: Annoying to duplicate this logic in the updater, but I'd rather a few lines duplicated rather than a huge refactor or something

markvanlan: I don't want to raise the error in `chat_message.rb`, so I have to do it here.

lib/chat_message_updater.rb
```diff
@@ -21,21 +21,29 @@ def initialize(chat_message:, new_content:, upload_ids: nil)
...
+      Jobs.enqueue(:process_chat_message, { chat_message_id: @chat_message.id })
+      DiscourseChat::ChatNotifier.notify_edit(chat_message: @chat_message, timestamp: revision.created_at)
+    rescue => error
+      @error = error
+      if Rails.env.test?
+        puts "#" * 50
+        puts "Chat message update error:"
+        puts @error.inspect
+        puts "#" * 50
+      end
```
eviltrout, line 39: Not sure this needs to be in the commit?

eviltrout (approved):

lib/chat_message_updater.rb
markvanlan: That is just a whitespace change.. it's been in there. I find it helpful for local debugging as errors are swallowed in the `rescue` if you have a syntax error or something like that.

PR #560, opened 2022-01-25
DEV: Import test helpers explicitly, update eslint config

davidtaylorhq:
Our shared eslint config hasn't yet been updated for Ember CLI, so these globals were allowed. This commit applies core's modifications to the .eslintrc file so that we correctly warn about global use.


CvX (approved):

nattsw:
iirc the eslintrc file was taken from plugin skeleton. Do we have to update it?

davidtaylorhq:
@nattsw I think long-term we will move these rules into eslint-config-discourse, so there won't be any need for plugins to update their own `.eslintrc` files. I wanted to add them to `discourse-chat` specifically because it has a lot of developers working on it right now, and so it's easy for us to accidentally use a 'deprecated' global.

davidtaylorhq:
See https://github.com/discourse/eslint-config-discourse/pull/22

PR #569, opened 2022-01-26
DEV: displays a rebake button to admin for each message

jjaffeux:
This is an advanced feature but can be quite useful, especially while developing chat.

### DIFF
--- a/assets/javascripts/discourse/components/chat-message.js
+++ b/assets/javascripts/discourse/components/chat-message.js
@@ -142,6 +142,14 @@ export default Component.extend({
       });
     }
 
+    if (this.currentUser?.staff) {
+      buttons.push({
+        id: "rebakeMessage",
+        name: I18n.t("chat.rebake_message"),
+        icon: "sync-alt",
+      });
+    }
+
     return buttons;
   },
 
@@ -585,6 +593,16 @@ export default Component.extend({
   restore() {
     return ajax(
       `/chat/${this.details.chat_channel_id}/restore/${this.message.id}`,
+      {
+        type: "PUT",
+      }
+    ).catch(popupAjaxError);
+  },
+
+  @action
+  rebakeMessage() {
+    return ajax(
+      `/chat/${this.details.chat_channel_id}/${this.message.id}/rebake`,
       {
         type: "PUT",
       }
--- a/spec/requests/chat_controller_spec.rb
+++ b/spec/requests/chat_controller_spec.rb
@@ -253,7 +253,7 @@ RSpec.describe DiscourseChat::ChatController do
       end
 
       context "TL4 user" do
-        it "allows TL4 user to rebake" do
+        it "allows TL4 users to rebake" do
           sign_in(Fabricate(:user, trust_level: TrustLevel[4]))
           put "/chat/#{chat_channel.id}/#{chat_message.id}/rebake.json"
           expect(response.status).to eq(200)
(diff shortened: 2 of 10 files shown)


jjaffeux:
Screenshot 2022-01-26 at 18 05 19

assets/javascripts/discourse/components/chat-message.js
```diff
@@ -142,6 +142,14 @@ export default Component.extend({
       });
     }
 
+    if (this.currentUser?.staff) {
```
eviltrout, line 145: Perhaps it should be admin, not staff? It's very advanced.

eviltrout (approved):

assets/javascripts/discourse/components/chat-message.js
jjaffeux: Current rebake is `is_staff? || @user.has_trust_level?(TrustLevel[4])` for posts this is why I went with staff, I didn't go with TL4 because chat has an existing guardian for can_edit_message which seemed appropriate for this. But I could add TL4 if you think that's necessary.

eviltrout: I was just thinking of hiding things that most people won't need. But I'll leave it to your judgment! Whatever you think is best :)

spec/requests/chat_controller_spec.rb
```diff
@@ -208,6 +208,49 @@
...
+          put "/chat/#{chat_channel.id}/#{chat_message.id}/rebake.json"
+        end.to change { Jobs::ProcessChatMessage.jobs.size }.by(1)
+        expect(response.status).to eq(200)
+
+        job = Jobs::ProcessChatMessage.jobs.first
+        job_args = job["args"].first
+        expect(job_args["chat_message_id"]).to eq(chat_message.id)
+        expect(job_args["is_dirty"]).to be_blank
+        ChatPublisher.expects(:publish_processed!).never
+        job["class"].constantize.new.perform(*job["args"])
```
tgxworld, line 229: This assertion is a little odd. Why do we need to manually execute a job in a request spec? Perhaps you need to enable `Jobs.run_immediately!`? Also, we should try and avoid mocking as much as possible. Instead you can use `MessageBus.track_publish` to check that no message was published here.

jjaffeux: Yep I started with run_immediately but it prevented me to catch jobs size. ON second thought though, I just think this logic should be in the ProcessChatMessage spec, which is what I did. Concerning using `track_publish`, I'm also mixed on this, because we are testing that `ChatPublisher. publish_processed!` will be called, not that message bus will trigger a message, `publish_processed` implementation could change in the future and it's not the responsibility of the job to test the underlying implementation. I'm opened to a different refactor if you have a suggestion.

assets/javascripts/discourse/components/chat-message.js
jjaffeux: :+1: I went with exact parity with Discourse core and rebake for posts, I think this is what makes the most sense. People could find weird that they can rebake a post but not a chat message.

spec/requests/chat_controller_spec.rb
tgxworld: >  because we are testing that ChatPublisher. publish_processed! will be called, not that message bus will trigger a message, publish_processed implementation could change in the future and it's not the responsibility of the job to test the underlying implementation.

I disagree here because by testing that `ChatPublisher.publish_processed!` will be called, you're testing for the underlying implementation of the job. Asserting that a message_bus message is published is testing for the outcome of the job. Something along the lines of `I run this job X and I expect it to do Y`. In this case, Y is that a message_bus message is published.

PR #576, opened 2022-01-27
UX: Instantly filter chat channels in quick select modal

markvanlan:
`{{input onChange=}}` has a built on delay for something like 500ms after the user stops typing.

Hooking into `keyup` for the `input` element causes the filter to trigger instantly and it makes this modal much better.


eviltrout (approved):

CvX:
@markvanlan It was merged with failing tests 😛

PR #580, opened 2022-01-28
UX: Slightly tighter headers

CvX:
Previously I overshot it a bit with size of the buttons/dropdown.

---

~~Can a sidebar user please confirm it doesn't affect their layout? 😃~~
Checked and updated.

### DIFF
--- a/assets/stylesheets/common/chat-retention-reminder.scss
+++ b/assets/stylesheets/common/chat-retention-reminder.scss
@@ -31,5 +31,5 @@
 }
 
 .full-page-chat .chat-retention-reminder {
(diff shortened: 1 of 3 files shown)


markvanlan:
@CvX you really can't make css changes at all without testing both... I suggest cloning sidebar locally and adding `next false` to this and commenting it out when you want to test with sidebar 
https://github.com/discourse-org/discourse-teams-sidebar/blob/a02cb55e8fdc584253ee79c806235302fe58c506/plugin.rb#L56

pmusaraj (approved):

PR #581, opened 2022-01-28
DEV: Don't manually calc full chat height

CvX:
That method was error-prone, because if something were to be added/removed above the chat (e.g. an alert) it would not recalc the height.


markvanlan (approved):
Approving, but make sure it's tested 
* sidebar desktop
* no sidebar desktop
* mobile
 :)

jjaffeux:
What do we do with his @CvX  ?

jjaffeux:
:hugs: ?

CvX:
This is something we need to fix, but needs a different approach - core should provide the necessary tools, as it already has the required workarounds for mobile viewports, virtual keyboards and iPad quirks. Closing for now.

PR #582, opened 2022-01-28
UX: Sidebar rows 'active' when float is open

markvanlan:
This is a major improvement on how `activeChannel` is calc'd everywhere, by being manually set at the correct spots. The largest benefit of this is that channel-rows can hook into the chat services' `activeChannel` and highlight even when the float is open :)

Screen Shot 2022-01-28 at 2 16 09 PM

### DIFF
--- a/assets/javascripts/discourse/components/chat-channel-row.js
+++ b/assets/javascripts/discourse/components/chat-channel-row.js
@@ -1,18 +1,20 @@
 import Component from "@ember/component";
 import discourseComputed from "discourse-common/utils/decorators";
 import getURL from "discourse-common/lib/get-url";
+import { action } from "@ember/object";
 import { equal } from "@ember/object/computed";
 import { inject as service } from "@ember/service";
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse/components/chat-channel-row.js
```diff
@@ -75,11 +75,8 @@ export default Component.extend({
...
 
-  @discourseComputed("channel", "router.currentRoute")
-  active(channel, currentRoute) {
-    return (
-      currentRoute?.name === "chat.channel" &&
-      currentRoute?.params?.channelId === channel.id.toString(10)
-    );
+  @discourseComputed("channel", "chat.activeChannel")
+  active(channel, activeChannel) {
+    return channel.id === activeChannel?.id;
```
eviltrout, line 80: Looks like you could use `propertyEqual` here and shave a few lines off.

eviltrout (approved):

PR #588, opened 2022-02-03
FEATURE: Allow image markdown markup

xfalcox:
Will be used to integrate with discourse-gifs

https://user-images.githubusercontent.com/1385470/152429115-46046989-8ec7-4905-9535-2bf2d4f5123f.mp4


markvanlan (approved):
I think this should work, and not cause jumpiness on load...

PR #590, opened 2022-02-04
DEV: Use new model for chat drafts

nbianca:
This PR needs new plugin API: https://github.com/discourse/discourse/pull/15545

### DIFF
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -365,6 +365,22 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
     render json: success_json
   end
 
+  def set_draft
+    channel_id = params.require(:channel_id)
+
+    if params[:data].present?
+      ChatDraft
+        .find_or_initialize_by(user: current_user, chat_channel_id: channel_id)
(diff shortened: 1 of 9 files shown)


app/controllers/chat_controller.rb
```diff
@@ -329,6 +329,22 @@ def dismiss_retention_reminder
     render json: success_json
   end
 
+  def create_draft
```
markvanlan, line 332: This is pretty nit-picky, but what do you think about renaming this action to `set_draft`, since it's used for creating _and_ destroying drafts?

markvanlan (approved):

plugin.rb
```diff
@@ -296,6 +296,12 @@ def self.allowed_group_ids
     nil
   end
 
+  if respond_to?(:register_upload_in_use)
+    register_upload_in_use do |upload|
+      ChatDraft.exists?("data LIKE '%#{upload.sha1}%' OR data LIKE '%#{Base62.encode(upload.sha1.hex)}%'")
```
ZogStriP, line 301: Wouldn't `upload.base62_sha1` be better than directly calling `Base62.encode`?

PR #595, opened 2022-02-07
FIX: User could not re-enable chat after disabling it

martin-brennan:
In 18f45fc4c695407dec3df29edd3f163e86e17aab we made a change to not
call addSaveableUserOptionField for the chat-specific user options
if !userCanChat. However, this prevents the user from changing their
chat preferences if they disable chat and press save. They are not able
to re-enable it or change any other settings, because the user options
are no longer added.

### DIFF
--- a/assets/javascripts/discourse/controllers/preferences-chat.js
+++ b/assets/javascripts/discourse/controllers/preferences-chat.js
@@ -1,4 +1,5 @@
 import Controller from "@ember/controller";
+import { isTesting } from "discourse-common/config/environment";
 import discourseComputed from "discourse-common/utils/decorators";
 import I18n from "I18n";
 import { action } from "@ember/object";
@@ -22,7 +23,7 @@ export default Controller.extend({
 
   @action
   onChangeChatSound(sound) {
-    if (sound) {
+    if (sound && !isTesting()) {
       const audio = new Audio(CHAT_SOUNDS[sound]);
       audio.play();
     }
@@ -36,7 +37,9 @@ export default Controller.extend({
       .save(chatAttrs)
       .then(() => {
         this.set("saved", true);
-        location.reload();
+        if (!isTesting()) {
+          location.reload();
+        }
       })
       .catch(popupAjaxError);
   },
--- a/test/javascripts/acceptance/chat-test.js
+++ b/test/javascripts/acceptance/chat-test.js
@@ -33,6 +33,8 @@ import selectKit from "discourse/tests/helpers/select-kit-helper";
 import { next } from "@ember/runloop";
 import { Promise } from "rsvp";
 import { isLegacyEmber } from "discourse-common/config/environment";
+import sinon from "sinon";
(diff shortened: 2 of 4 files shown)


tgxworld:
Is there a test we can add for this?

martin-brennan:
@tgxworld hmm will see what I can do, I actually found a bug with this user option saving code in core that I am fixing and testing so might be able to make a similar test here.

assets/javascripts/discourse/controllers/preferences-chat.js
```diff
@@ -22,7 +23,7 @@ export default Controller.extend({
 
   @action
   onChangeChatSound(sound) {
-    if (sound) {
+    if (sound && !isTesting()) {
```
tgxworld, line 26: Just curious, why do we need to disable this?

test/javascripts/acceptance/chat-test.js
```diff
@@ -1225,6 +1227,32 @@ acceptance("Discourse Chat - chat preferences", function (needs) {
     assert.equal(currentURL(), "/u/eviltrout/preferences/chat");
     assert.equal(queryAll(".chat-setting").length, 4);
   });
+
+  test("The user can save the settings", async function (assert) {
+    updateCurrentUser({ has_chat_enabled: false });
+    let spy = sinon.spy(ajaxlib, "ajax");
```
tgxworld, line 1233:
```suggestion
    const spy = sinon.spy(ajaxlib, "ajax");
```

assets/javascripts/discourse/controllers/preferences-chat.js
martin-brennan: Because when we change the select kit option for the sound this triggers, and then we get an error in the test about the user not having interacted with the document. Pointless to try and play sounds in the test anyway

test/javascripts/acceptance/chat-test.js
```diff
@@ -1225,6 +1227,32 @@ acceptance("Discourse Chat - chat preferences", function (needs) {
...
+    await click("#user_chat_enabled");
+    await click("#user_chat_only_push_notifications");
+    await click("#user_chat_isolated");
+    await selectKit("#user_chat_sounds").expand();
+    await selectKit("#user_chat_sounds").selectRowByValue("bell");
+
+    await click(".save-changes");
+
+    assert.ok(
+      spy.calledWithMatch("/u/eviltrout.json", {
```
tgxworld, line 1244: Usually I would say we want to avoid mocks but I have no idea how else we can test this properly :man_shrugging:

tgxworld (approved):

test/javascripts/acceptance/chat-test.js
martin-brennan: @tgxworld yeah it's the same in core, have to use sinon to see what is actually going to get sent to the server. `spy` at least is just a wrapper around the original method, doesn't mock it out completely so that's nice

PR #596, opened 2022-02-07
FEATURE: Collapse non-oneboxed gifs and images in general

nattsw:
Just another type of image to collapse. The difference is that this one is cooked different from the rest, like this

```
<p><img src=...></p>
```

Screenshot 2022-02-07 at 12 26 07 PM

### DIFF
--- a/assets/javascripts/discourse/components/chat-message-collapser.js
+++ b/assets/javascripts/discourse/components/chat-message-collapser.js
@@ -64,6 +64,27 @@ export default Component.extend({
   },
 
   @computed("cooked")
+  get imageCooked() {
+    const elements = Array.prototype.slice.call(domFromString(this.cooked));
+
+    return elements.reduce((acc, e) => {
+      if (imagePredicate(e)) {
+        const link = e.firstElementChild.src;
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse/components/chat-message-collapser.js
```diff
@@ -124,6 +150,24 @@ function hasUploads(uploads) {
   return uploads?.length > 0;
 }
 
+function imagePredicate(e) {
```
tgxworld, line 153: Not really related to your PR but since the functions are not exported, is there a reason we choose to define the functions outside the component instead of declaring them as private functions in the component itself?

nattsw: Actually the last function in the file (`isCollapsible`) is exported.

tgxworld: Minor but what do you think about `isImage` as the function name as compared to using `imagePredicate`?

tgxworld (approved):

tgxworld:
Some minor comments but in general the PR looks really good to me :+1:

PR #597, opened 2022-02-07
FIX: Prevent youtube from being gigantic

nattsw:
There's currently no sizing for youtube so everything's gigantic. This adds some fixed styles to youtube oneboxes with this assumption https://support.google.com/youtube/answer/6375112:

> The standard aspect ratio for YouTube on a computer is 16:9. If your video has a different aspect ratio, the player will automatically change to the ideal size to match your video and the viewer’s device. 

Images are currently set to a max height of 150px, so we're basing our 16:9 off that value. Here's an example of the player doing padding (based on link above)

> For some video and device aspects ratios like 9:16 vertical videos on computer browsers, YouTube may add more padding for optimal viewing.

Screenshot 2022-02-07 at 1 11 21 PM


tgxworld (approved):

tgxworld:
I didn't test this out locally but the changes look safe and good to me :+1:

PR #599, opened 2022-02-09
FEATURE: Chat message flagging

markvanlan:
Screen Shot 2022-02-09 at 12 36 13 PM

Screen Shot 2022-02-09 at 12 35 53 PM

Screen Shot 2022-02-09 at 12 36 34 PM

Screen Shot 2022-02-09 at 12 36 49 PM

Screen Shot 2022-02-09 at 12 36 54 PM


app/models/chat_message.rb
```diff
@@ -49,6 +44,45 @@ def push_notification_excerpt
...
 
+  def add_flag(user)
+    reviewable = ReviewableChatMessage.needs_review!(
+      created_by: user,
+      target: self,
+    )
+    reviewable.update(target_created_by: self.user)
+    reviewable.add_score(
+      user,
+      ReviewableScore.types[:needs_approval],
```
eviltrout, line 55: I'm not sure "Needs Approval" is the correct type here. Approval refers to a specific thing in our app - when a post or other thing needs to be approved before it is posted. 

I would suggest adding a new score type. Maybe `review_requested` or `needs_review` or similar?

assets/javascripts/discourse/models/chat-message.js
```diff
@@ -0,0 +1,4 @@
+import RestModel from "discourse/models/rest";
+
+const ChatMessage = RestModel.extend();
```
eviltrout, line 3: Huge nitpick here but you needn't assign `ChatMessage` - you could `export default RestModel.extend()`

config/locales/server.en.yml
```diff
@@ -9,6 +9,54 @@ en:
...
+          title: "Delete Message"
+          description: "Delete the message so that users cannot see it."
+        delete_single:
+          title: "Delete"
+        delete:
+          title: "Delete..."
+        delete_and_ignore:
+          title: "Delete Message and Ignore"
+          description: "Delete message; if the first message, delete the topic as well"
+        delete_and_ignore_replies:
```
eviltrout, line 41: You're not using all these keys are you? This is suspicious to me:

> "Delete message and all of its replies; if the first message, delete the topic as well"

eviltrout (commented):
It looks good so far. It's too bad @romanrizzi is on vacation because he'd be great to review it. Maybe once he gets back he can go through it as a second pair of eyes?

config/locales/server.en.yml
markvanlan: good catch

assets/stylesheets/common/common.scss
```diff
@@ -1366,3 +1366,17 @@ body.has-full-page-chat {
     white-space: nowrap;
   }
 }
+
+.chat-flag-inputs {
```
pmusaraj, line 1370: Does this element exist? Can't find it from a quick grep of the branch, but maybe I'm doing it wrong.

PR #601, opened 2022-02-10
FEATURE: Chat quoting

martin-brennan:
This commit introduces chat message quoting functionality. Selected messages from any channel _except_ DM channels can be quoted:

image

This commit also allows non-staff users to bulk select messages, because it is necessary for quoting. The "Move to Topic" button is hidden for non-staff.

Quoting has a few different behaviours based on the current state:

1. When isolated on desktop, copy to clipboard and show a little message
2. Not isolated on desktop, open composer and insert markdown. This may open a reply composer if the user is currently inside a topic, otherwise it will open a new topic composer
3. On mobile, go to the category / topic that the chat is for and open the composer and insert markdown

When the quote button is clicked, a call to the `/chat/:channelId/quote` endpoint is sent with the selected message IDs. The quote BBCode is built with `ChatTranscriptService`. The bbcode is constructed server-side because other functions, such as Move to Topic and Archive Channel will need to build up the same bbcode later on. See the comment on `ChatTranscriptService` for further information on how that works.


app/controllers/chat_controller.rb
```diff
@@ -340,6 +340,17 @@ def dismiss_retention_reminder
     render json: success_json
   end
 
+  def quote_messages
+    params.require(:message_ids)
+
+    @chat_channel = ChatChannel.find(params[:chat_channel_id])
```
tgxworld, line 346: This will raise a `ActiveRecord::RecordNotFound` error if the param is invalid. We usually do a `find_by` and return the right response if we don't get anything back.

martin-brennan: ah yeah you are right, my bad

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -896,9 +899,28 @@ export default Component.extend({
     next(this.reStickScrollIfNeeded.bind(this));
   },
 
+  @discourseComputed()
+  canQuote() {
+    if (this.chatChannel.chatable_type === "DirectMessageChannel") {
```
nattsw, line 904: We've got this string littered many places (I've contributed to it too probably) but maybe we can start using `CHATABLE_TYPES.directMessageChannel`

lib/chat_transcript_service.rb
```diff
@@ -0,0 +1,99 @@
...
+# of the username and datetime of the message.
+class ChatTranscriptService
+  CHAINED_ATTR = "chained=\"true\""
+  MULTIQUOTE_ATTR = "multiQuote=\"true\""
+
+  class ChatTranscriptBBCode
+    attr_reader :channel, :multiquote, :chained
+
+    def initialize(
+      channel: nil,
```
tgxworld, line 25: Should this be a required param instead? Not sure if we `channel = nil` makes sense in the context of chat.

martin-brennan: It's not displayed in every chat BBCode block, these are the three different displays:

image

assets/javascripts/discourse/components/chat-live-pane.js
martin-brennan: Yeah I will do that in a separate PR, might even make sense to export a proper model for chatChannel and have shortcut methods like `isDirectMessage` like the ruby chat_channel model.

assets/javascripts/discourse/templates/components/chat-live-pane.hbs
```diff
@@ -80,21 +80,43 @@
...
+          title="chat.selection.cancel"
+          action=(action "cancelSelecting")
+        }}
+      </div>
+      {{#if showChatQuoteSuccess}}
+        <div class="chat-selection-message alert alert-success">
+          {{i18n "chat.quote.copy_success"}}
+        </div>
+      {{/if}}
     </div>
```
nattsw, line 120: I'm not sure how hard it is to extract values out, and I understand this PR is already quite big, but this div seems like it could be a component on its own? If it can be done it'll be nice so the `chat-live-pane` doesn't get too gnarly when it grows even bigger.

martin-brennan: Yeah that makes sense, should probably be called `chat-selection-management` component

test/javascripts/acceptance/chat-transcript-test.js
```diff
@@ -272,7 +272,7 @@ another cool reply<br>
       `[chat quote="martin;2321;2022-01-25T05:40:39Z"]\nhere is a message :P with category hashtag #test\n[/chat]`,
       { additionalOptions },
       generateTranscriptHTML(
-        `<p>here is a message <img src=\"/images/emoji/twitter/stuck_out_tongue.png?v=12\" title=\":stuck_out_tongue:\" class=\"emoji\" alt=\":stuck_out_tongue:\"> with category hashtag <span class=\"hashtag\">#test</span></p>`,
+        `<p>here is a message <img src=\"/images/emoji/twitter/stuck_out_tongue.png?v=12\" title=\":stuck_out_tongue:\" class=\"emoji\" alt=\":stuck_out_tongue:\" loading=\"lazy\" width=\"20\" height=\"20\"> with category hashtag <span class=\"hashtag\">#test</span></p>`,
```
nattsw, line 275: Oh this is interesting, what caused this?

spec/models/chat_message_spec.rb
```diff
@@ -109,7 +109,7 @@
     it 'supports emoji plugin' do
       cooked = ChatMessage.cook(":grin:")
 
-      expect(cooked).to eq("<p><img src=\"/images/emoji/twitter/grin.png?v=12\" title=\":grin:\" class=\"emoji only-emoji\" alt=\":grin:\"></p>")
+      expect(cooked).to eq("<p><img src=\"/images/emoji/twitter/grin.png?v=12\" title=\":grin:\" class=\"emoji only-emoji\" alt=\":grin:\" loading=\"lazy\" width=\"20\" height=\"20\"></p>")
```
nattsw, line 112: Oh this is interesting, what caused this?

assets/javascripts/discourse/components/chat-live-pane.js
nattsw: Thank you <3

lib/chat_transcript_service.rb
```diff
@@ -0,0 +1,99 @@
...
+      attrs << CHAINED_ATTR if chained
+
+      <<~MARKDOWN
+      [chat #{attrs.join(" ")}]
+      #{@messages.map(&:message).join("\n\n")}
+      [/chat]
+      MARKDOWN
+    end
+
+    def quote_attr(message)
```
tgxworld, line 52:
```suggestion
    private

    def quote_attr(message)
```

lib/chat_transcript_service.rb
```diff
@@ -0,0 +1,99 @@
...
+      open_bbcode_tag << message
+      previous_message = message
+    end
+
+    # tie off the last open bbcode + render
+    rendered_bbcode << open_bbcode_tag.render
+    rendered_bbcode.join("\n")
+  end
+
+  def messages
```
tgxworld, line 94:
```suggestion
  private

  def messages
```

spec/models/chat_message_spec.rb
tgxworld: O this is from a recent commit where we cook emojis differently.

spec/lib/chat_transcript_service_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    [/chat]
+    MARKDOWN
+  end
+
+  it "generates a single chat transcript from multiple subsequent messages from the same user" do
+    message1 = Fabricate(:chat_message, user: user1, chat_channel: channel, message: "an extremely insightful response :)")
+    message2 = Fabricate(:chat_message, user: user1, chat_channel: channel, message: "if i say so myself")
+    message3 = Fabricate(:chat_message, user: user1, chat_channel: channel, message: "yay!")
+
+    rendered = service([message1.id, message2.id, message3.id]).generate_bbcode
```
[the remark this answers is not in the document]
martin-brennan: Sure, will add a test to make sure they are ordered correctly no matter the ID sequence

spec/requests/chat_controller_spec.rb
```diff
@@ -630,4 +630,64 @@
...
+
+    it "returns a 403 if the user can't chat" do
+      SiteSetting.chat_allowed_groups = nil
+      sign_in(user)
+      post "/chat/#{channel.id}/quote.json", params: { message_ids: [message1.id, message2.id, message3.id] }
+      expect(response.status).to eq(403)
+    end
+
+    it "returns a 403 if the user can't see the channel" do
+      category.update(read_restricted: true)
```
tgxworld, line 649:
```suggestion
      category.update!(read_restricted: true)
```

spec/requests/chat_controller_spec.rb
```diff
@@ -630,4 +630,64 @@
...
+      expect(response.status).to eq(403)
+    end
+
+    it "returns a 403 if the channel is a DM channel" do
+      sign_in(user)
+      dm_channel_chatable = Fabricate(:direct_message_channel, users: [user, user2])
+      dm_channel = Fabricate(:chat_channel, chatable: dm_channel_chatable)
+      message1.update(chat_channel: dm_channel)
+      message2.update(chat_channel: dm_channel)
+      message3.update(chat_channel: dm_channel)
```
tgxworld, line 661:
```suggestion
      message1.update!(chat_channel: dm_channel)
      message2.update!(chat_channel: dm_channel)
      message3.update!(chat_channel: dm_channel)
```

tgxworld: Also, I'm not sure if we need to create 3 messages here since the messages are kind of irrelevant in this case.

lib/chat_transcript_service.rb
```diff
@@ -0,0 +1,99 @@
...
+      "channel=\"#{channel.name}\""
+    end
+  end
+
+  def initialize(channel, message_ids)
+    @channel = channel
+    @message_ids = message_ids
+  end
+
+  def generate_bbcode
```
tgxworld, line 66: I wonder if we should name this something more generic. `bbcode` is an internal implementation detail which may or may not change in the future. If it does change, might be annoying to  have to update all the references everywhere. Maybe something like `generate_markdown`?

test/javascripts/acceptance/chat-quoting-test.js
```diff
@@ -0,0 +1,235 @@
...
+  an extremely insightful response :)
+  [/chat]`,
+};
+
+function setupPretenders(server, helper) {
+  server.get("/chat/chat_channels.json", () => helper.response(chatChannels));
+  server.post(`/chat/9/quote.json`, () => helper.response(quoteResponse));
+  server.post(`/chat/4/quote.json`, () => helper.response(quoteResponse));
+  server.post(`/chat/7/quote.json`, () => helper.response(quoteResponse));
+  server.get("/chat/:chatChannelId/messages.json", () =>
```
tgxworld, line 26:
```suggestion
  server.get("/chat/:chat_channel_id/messages.json", () =>
```

Minor but kind of odd since this isn't a JS variable.

test/javascripts/acceptance/chat-quoting-test.js
```diff
@@ -0,0 +1,235 @@
...
+    await dropdown.expand();
+    await dropdown.selectRowByValue("selectMessage");
+
+    assert.ok(firstMessage.classList.contains("selecting-messages"));
+    const quoteBtn = query(".chat-live-pane #chat-quote-btn");
+    assert.equal(
+      quoteBtn.disabled,
+      false,
+      "button is enabled as a message is selected"
+    );
```
tgxworld, line 65: I think it is more relevant to test that the button is disabled when no message is selected. This assertion is indirectly handled when we try to click the button below.

test/javascripts/acceptance/chat-quoting-test.js
```diff
@@ -0,0 +1,235 @@
...
+    await click("#chat-quote-btn");
+    assert.ok(exists("#reply-control.composer-action-createTopic"));
```
tgxworld, line 67: not related to your changes but the casing on this class name is so weird.

martin-brennan:
Test to see if this lets me comment on sub-convos.

spec/requests/chat_controller_spec.rb
martin-brennan: Yeah good point will just use one

lib/chat_transcript_service.rb
martin-brennan: Yeah good point I was wondering about that a bit, will change

test/javascripts/acceptance/chat-quoting-test.js
martin-brennan: Just copied it from another test, but yeah will update

test/javascripts/acceptance/chat-quoting-test.js
martin-brennan: Yup np will add that

test/javascripts/acceptance/chat-quoting-test.js
martin-brennan: Yeah it's because it comes directly from the composer actions defined here:

https://github.com/discourse/discourse/blob/febc2361bd99be26317d9fe06c8024b2696e887a/app/assets/javascripts/discourse/app/models/composer.js#L37-L44

test/javascripts/acceptance/chat-quoting-test.js
```diff
@@ -0,0 +1,235 @@
...
+  needs.settings({
+    chat_enabled: true,
+  });
+
+  needs.pretender((server, helper) => {
+    setupPretenders(server, helper);
+  });
+
+  test("it does not show the quote button in direct messages", async function (assert) {
+    await visit("/chat/channel/75/@hawk");
```
tgxworld, line 94: I'm not sure about the existing chat code but do we pass in the symbol in the URL? Seems odd to me.

martin-brennan:
Ah just realised I need to make sure that messages with images get quoted correctly too.

test/javascripts/acceptance/chat-quoting-test.js
martin-brennan: Yup we sure do `/chat/channel/389/@osama`

test/javascripts/acceptance/chat-quoting-test.js
```diff
@@ -0,0 +1,235 @@
...
+
+  needs.settings({
+    chat_enabled: true,
+  });
+
+  needs.pretender((server, helper) => {
+    setupPretenders(server, helper);
+  });
+
+  test("it opens the composer and appends the quote", async function (assert) {
```
tgxworld, line 52: I'm wondering if we should assert that the category is prefilled in the composer too.

martin-brennan: Ah I did that in one of the other tests, will add it here too:

https://github.com/discourse-org/discourse-topic-chat/blob/09c088eb4a478f1dfa4d4c153eda4ed228188c81/test/javascripts/acceptance/chat-quoting-test.js#L229-L233

test/javascripts/acceptance/chat-test.js
```diff
@@ -725,6 +725,20 @@ acceptance("Discourse Chat - without unread", function (needs) {
     );
   });
 
+  test("message selection and live pane buttons for regular staff", async function (assert) {
```
tgxworld, line 728: I'm not sure which is wrong here because the description says for staff but `admin` and `moderator` is `false` below.

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -937,19 +959,79 @@ export default Component.extend({
...
   },
 
+  @action
+  onChannelTitleClick() {
+    return this._goToChatableUrl();
+  },
+
+  @action
+  quoteMessages() {
+    ajax(getURL(`/chat/${this.chatChannel.id}/quote.json`), {
```
tgxworld, line 975: I don't think we need to test this but we probably want to handle any potential errors.

test/javascripts/acceptance/chat-test.js
martin-brennan: Ah brain fart, should have said `for regular user`

nattsw (approved):

spec/lib/chat_transcript_service_spec.rb
```diff
@@ -74,4 +74,46 @@ def service(message_ids)
...
+    image_markdown = UploadMarkdown.new(image).to_markdown
+
+    expect(service(message.id).generate_markdown).to eq(<<~MARKDOWN)
+    [chat quote="martinchat;#{message.id};#{message.created_at.iso8601}" channel="The Beam Discussions"]
+    #{video_markdown}
+    #{audio_markdown}
+    #{attachment_markdown}
+    #{image_markdown}
+    [/chat]
+    MARKDOWN
```
nattsw, line 101: 🧐 hmm should we be asserting the result of `UploadMarkdown`?

I guess the responsibility of this test is interesting. I'm thinking perhaps you can shorten this (to just one authorized extension and one non-authorized one, perhaps??) by creating unit tests for `chat_message#to_markdown`

martin-brennan: > asserting the result of UploadMarkdown

Nah I think that should be the responsibility of the UploadMarkdown test, which I added onto in core with https://github.com/discourse/discourse/pull/15930

Though I will add a unit test for `to_markdown`. The authorized extension stuff shouldn't matter too much either, the upload would not exist if the extension was unauthorized. I mainly wanted in this test to check that all the different types of uploads would get their markdown.

PR #602, opened 2022-02-10
FIX: Prevent update read timer from stalling

markvanlan:
Currently the `_updateReadTimer` is set in `willIInsertElement`, and if something goes wrong and the timer is canceled, it will only be started again if the component is destroyed and re-created. Changing channels does not fix this.

This moves the initialization of `updateReadTimer` into a function called on channel change.


CvX:
Oh, looks like the specs need to be updated after the core emoji change

PR #609, opened 2022-02-15
FEATURE: Add simple composer shortcuts

martin-brennan:
This commit adds the simplest "surround" formatting shortcuts
from the composer into the chat composer:

* ctrl+b - bold selection
* ctrl+i - italicize selection
* ctrl+e - inline code selection

These shortcuts only work when the chat composer is focused.

Requires https://github.com/discourse/discourse/pull/15950 to work.

### DIFF
--- a/assets/javascripts/discourse/initializers/chat-keyboard-shortcuts.js
+++ b/assets/javascripts/discourse/initializers/chat-keyboard-shortcuts.js
@@ -32,6 +32,16 @@ export default {
       chatService.switchChannelUpOrDown("down");
     };
 
+    const isChatComposer = (el) => el.classList.contains("chat-composer-input");
+    const modifyComposerSelection = (event, type) => {
+      if (!isChatComposer(event.target)) {
+        return;
+      }
+      event.preventDefault();
+      event.stopPropagation();
+      appEvents.trigger("chat:modify-selection", { type });
+    };
+
     withPluginApi("0.12.1", (api) => {
       api.addKeyboardShortcut("command+k", openChannelSelector, {
         global: true,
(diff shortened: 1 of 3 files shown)


assets/javascripts/discourse/initializers/chat-keyboard-shortcuts.js
```diff
@@ -41,6 +51,22 @@ export default {
...
+        (event) => modifyComposerSelection(event, "bold"),
+        { global: true }
+      );
+      api.addKeyboardShortcut(
+        "ctrl+i",
+        (event) => modifyComposerSelection(event, "italic"),
+        { global: true }
+      );
+      api.addKeyboardShortcut(
+        "ctrl+e",
```
CvX, line 66: Are these `ctrl+?` on all platforms? If so, can we make it `cmd+?` on apple platforms like we do in d-editor?

https://github.com/discourse/discourse/blob/a48231041b3f533f9d26be6e68ed1cadaea93495/app/assets/javascripts/discourse/app/components/d-editor.js#L197-L206

config/locales/client.en.yml
```diff
@@ -121,6 +121,9 @@ en:
 
       composer:
         toggle_toolbar: "Toggle toolbar"
+        italic_text: "emphasized text"
+        bold_title: "Strong"
+        code_text: "indent preformatted text by 4 spaces"
```
CvX, line 126: Sentence case or lower case? Either way, we should stick to one 😉

martin-brennan: I copied and pasted the wrong key :joy: should have gotten `bold_text`, fixing

assets/javascripts/discourse/initializers/chat-keyboard-shortcuts.js
martin-brennan: Ah yes they need to be dynamic per platform, ty will fix

martin-brennan:
@CvX made those fixes now :+1:

config/locales/client.en.yml
```diff
@@ -121,6 +121,9 @@ en:
 
       composer:
         toggle_toolbar: "Toggle toolbar"
+        italic_text: "emphasized text"
+        bold_text: "strong text"
+        code_text: "indent preformatted text by 4 spaces"
```
CvX, line 126: Correct me if I'm wrong 😉 but it doesn't indent but surrounds with \`\`? 🙂

martin-brennan: hmm good point...will just change to `code text`

martin-brennan: This only shows up if you press the shortcut with nothing selected anyway

CvX (approved):

CvX (approved):

PR #610, opened 2022-02-15
UX: Fix style inconsistencies in chat quoting

CvX:
* Got rid of the border to match the regular quotes (btw. always use `var(--x)`, never `$x`; that's why it was broken on dark theme 😃)
* Changed the username color to match regular quotes
* Fixed timestamp alignment
* Increased the minimal font size (header)
* Removed the period after the channel link
* Made the margins of `<p>` elements smaller

Before / After

Screen Shot 2022-02-15 at 11 52 03

Screen Shot 2022-02-15 at 11 51 13


martin-brennan (approved):
tighten

Thanks @CvX looks awesome :)

PR #612, opened 2022-02-16
FEATURE: Rate limit message sending with btn to re-send

markvanlan:
Screen Shot 2022-02-16 at 11 18 59 AM

### DIFF
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -77,6 +77,8 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
   end
 
   def create_message
+    RateLimiter.new(current_user, "create_chat_message", 1, 1.second).performed! # 1 message per second
+
     set_channel_and_chatable
     @user_chat_channel_membership = UserChatChannelMembership.find_by(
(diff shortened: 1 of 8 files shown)


app/controllers/chat_controller.rb
```diff
@@ -77,6 +77,8 @@ def disable_chat
   end
 
   def create_message
+    RateLimiter.new(current_user, "create_chat_message", 1, 1.second).performed! # 1 message per second
```
oblakeerickson, line 80: Can we add an rspec test this as well?

oblakeerickson (changes requested):

app/controllers/chat_controller.rb
markvanlan: Yeah I guess! It feels a bit like testing the implementation, since `RateLimiter#perform` is tested, but it can't hurt

oblakeerickson: Ya I like it more for regression testing. So that any future changes we make to the controller we know we didn't accidentally remove the rate limiter.

oblakeerickson (approved):

PR #619, opened 2022-02-21
FEATURE: Chat channel statuses

martin-brennan:
This PR adds `status` to chat channel, with `open`, `closed`, `read_only`, and `archived` statuses (default `open`).

Rules have been introduced in the UI and in the backend to handle the various channel statuses, with the `Guardian` extensions focused on, with a lot of new tests there. No UI exists yet to change the channel status, but only staff will be able to do that in future. When the channel status is changed, we log it in the staff action log, fire a `DiscourseEvent`, and a message bus message as well.

The rules for each status are:

**open**

|         | staff | user |
|---------|-------|------|
| create  | ✔️     | ✔️    |
| destroy | ✔️     | ✔️    |
| edit    | ✔️     | ✔️    |
| react   | ✔️     | ✔️    |
| restore | ✔️     | ❌    |
| rebuild | ✔️     | ❌    |

**closed**

|         | staff | user |
|---------|-------|------|
| create  | ✔️     | ❌     |
| destroy | ✔️     | ❌     |
| edit    | ✔️     | ❌     |
| react   | ✔️     | ❌     |
| restore | ✔️     | ❌    |
| rebuild | ✔️     | ❌    |

**read_only / archived**

|         | staff | user |
|---------|-------|------|
| create  | ❌     | ❌     |
| destroy | ❌     | ❌     |
| edit    | ❌     | ❌     |
| react   | ❌     | ❌     |
| restore | ❌     | ❌    |
| rebuild | ❌     | ❌    |

------

Also some improvements to this PR in the chat JS. I've now started using the `ChatChannel` JS model so we can start adding shortcut functions for it, replacing the bare `EmberObject` we were using previously. I would like to make another PR later to avoid things like this, and have `.isDirectMessageChannel()` instead:

```javascript
const inDmChannel = activeChannel.chatable_type === CHATABLE_TYPES.directMessageChannel;
```


app/services/chat_publisher.rb
```diff
@@ -130,6 +130,17 @@ def self.publish_channel_edit(chat_channel)
     )
   end
 
+  def self.publish_channel_status(chat_channel)
```
martin-brennan, line 133: I think later on we can handle some sort of dynamic reaction to this. Perhaps for now a reload of the page is enough?

app/controllers/chat_channels_controller.rb
```diff
@@ -111,6 +111,8 @@ def create
   end
 
   def edit
+    # TODO (martin) Do we need to prevent editing the channel name/description
+    # when archived/read_only?
```
CvX, line 115: Ditto, do we want to resolve the question in this PR or later? 🙂

assets/javascripts/discourse/components/chat-message.js
```diff
@@ -95,7 +95,8 @@ export default Component.extend({
     "canFlagMessage",
     "showDeleteButton",
     "showRestoreButton",
-    "showEditButton"
+    "showEditButton",
+    "showRebakeButton"
   )
   get moreButtons() {
```
CvX, line 101: For consistency, I'd be great to convert that to a `discourseComputed`

assets/javascripts/discourse/components/chat-channel-selector-modal-inner.js
```diff
@@ -140,7 +141,7 @@ export default Component.extend({
             }
           });
           this.setProperties({
-            channels: channels.map((c) => EmberObject.create(c)),
+            channels: channels.map((channel) => ChatChannel.create(channel)),
```
CvX, line 144: This is a bigger topic, but… we should not manually create ChatChannel instances all over the place. The same with calling ajax in components like here.

We should extract that logic to model/adapter/service, and then rely on the store, e.g. `this.store.find("chat-channel", id)` or `this.store.findFiltered("chat-channel", { filter })`

app/controllers/direct_messages_controller.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class DiscourseChat::DirectMessagesController < DiscourseChat::ChatBaseController
+  # TODO (martin) Do we allow archiving DM channels, and if so what does that look like?
```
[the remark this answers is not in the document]
martin-brennan: Good question to my question :joy: I will post them both in the archive topic.

app/models/chat_channel.rb
```diff
@@ -16,6 +16,35 @@ class ChatChannel < ActiveRecord::Base
...
   has_many :user_chat_channel_memberships
 
+  def self.statuses
+    @statuses ||= Enum.new(
+      open: 0,
+      read_only: 1,
+      closed: 2,
+      archived: 3
+    )
+  end
```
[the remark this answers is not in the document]
martin-brennan: It doesn't seem like the PR for that is merged yet, but I will still change anyway, this is the syntax correct?

image

config/locales/server.en.yml
```diff
@@ -9,6 +9,8 @@ en:
     errors:
      channel_exists_for_category: "A channel already exists for this category and name"
      channel_exists_for_topic: "A channel already exists for this topic"
+     channel_new_message_disallowed: "The channel is %{status}, no new messages can be created"
+     channel_modify_message_disallowed: "The channel is %{status}, no messages can be modified"
```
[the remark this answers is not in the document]
martin-brennan: Fair enough, for modify_message I will say `no messages can be edited or deleted` and for the other I will say `no new messages can be sent`

app/services/chat_publisher.rb
nattsw: > dynamic reaction

Do you mean something like booting people from the channel or disabling the chat box?

martin-brennan: Yeah basically changing the UI in response to the status changing. The server-side guardian checks will make sure that no new messages get through, just not sure how we want the user to see this change. Maybe will ask in the topic.

martin-brennan:
@CvX made those initial review changes now :)

app/models/chat_channel.rb
```diff
@@ -99,6 +154,32 @@ def self.is_enabled?(t)
 
     ChatChannel.where(chatable: topic).exists?
   end
+
+  private
+
+  def log_channel_status_change(acting_user:, new_status:, old_status:)
+    new_status = new_status.to_sym
+    old_status = old_status.to_sym
```
nattsw, line 162: Is there a chance old_status is nil?

nattsw, line 162: Do we need a default value for old_status?

assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -254,7 +255,11 @@ export default Component.extend(TextareaTextManipulation, ComposerUploadUppy, {
   didReceiveAttrs() {
     this._super(...arguments);
 
-    if (!this.editingMessage && this.draft) {
+    if (
+      !this.editingMessage &&
+      this.draft &&
+      this.chatChannel.canModifyMessages(this.currentUser)
+    ) {
```
nattsw, line 262: Something else to think of is if we delete drafts for channels that have been closed?

martin-brennan: Good question -- I don't think we do anything to drafts for topics that are closed, so IMO we should leave these too

app/models/chat_channel.rb
martin-brennan: No, it will always have a value. The default status for a channel is `"open"`

martin-brennan: No no chance, it will always have a value

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1012,6 +1013,49 @@ export default Component.extend({
...
+
+    switch (channelStatus) {
+      case CHANNEL_STATUSES.closed:
+        if (this.currentUser.staff) {
+          return I18n.t("chat.channel_status.closed_staff_header");
+        } else {
+          return I18n.t("chat.channel_status.closed_header");
+        }
+        break;
+      case CHANNEL_STATUSES.readOnly:
```
nattsw, line 1030: Okay this is silly but I was staring hard at `readOnly` and I thought it should be [readonly](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly).

You don't have to change this, but it just gives me the funnies.

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1012,6 +1013,49 @@ export default Component.extend({
...
+        // FIXME (martin): Use a pencil-slash icon here, we don't have one in FA5
+        return "far-eye";
+        break;
+      case CHANNEL_STATUSES.archived:
+        return "folder";
+        break;
+    }
+  },
+
```
nattsw, line 1058: I'm not sure if it's worth it to do this, but the `chat-live-pane` is one of those files that's just getting bigger, and bigger, and bigger...

I'm wondering if it's worth it to extract a `channel-status` component to isolate this logic.

assets/javascripts/discourse/components/chat-live-pane.js
martin-brennan: But then maybe people would mix it up with the special attribute :P I know in rails I had to call it read_only because readonly had special meaning on the models

assets/javascripts/discourse/components/chat-live-pane.js
martin-brennan: Great idea! I have moved it now :)

nattsw (approved):
🚀 👏 
wow

assets/javascripts/discourse/components/chat-message.js
CvX: Yeah, discourseComputed should be applied to a regular method/function

martin-brennan: Done!

assets/javascripts/discourse/components/chat-message.js
```diff
@@ -91,15 +91,15 @@ export default Component.extend({
...
+  @discourseComputed(
     "selectingMessages",
     "canFlagMessage",
     "showDeleteButton",
     "showRestoreButton",
     "showEditButton",
     "showRebakeButton"
   )
-  get moreButtons() {
+  moreButtons() {
```
nattsw, line 110: Did you get any input on what we should do? Favouring `@discourseComputed` vs `@computed get`?

martin-brennan: Yeah CVX said to use `@discourseComputed`...maybe I will add a post in our lore topic about it.

PR #622, opened 2022-02-22
DEV: Move chat reaction code into lib class

martin-brennan:
This will make the reaction code easier to work with
and also will add a nicer place to add more validations
before creating/removing the reaction (such as in the
upcoming readonly chat channel work).

We already follow this pattern with ChatMessageCreator
and ChatMessageUpdater.


jjaffeux (approved):

jjaffeux:
I'm not sure add/remove need to be symbols, but this is not due to the PR

PR #624, opened 2022-02-22
DEV: adds a plugin API to chat

jjaffeux:


jjaffeux:
> What's the intended use? 🙂 And a follow up question: is manipulating the DOM element the best api for that?

So the use case is for example using mermaid in chat: https://github.com/discourse/discourse-mermaid-theme-component/pull/6/files#diff-346c71c851e37031e05989a947922a438122fe17af786aace7b5475dbe154559R91

I'm not sure I follow the second question. This is already what we do in core.

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -526,6 +536,14 @@ export default Component.extend({
       this._scrollGithubOneboxes();
       this._pluginsDecorators();
       this._highlightCode();
+
+      document
+        .querySelectorAll(".chat-message-text")
```
CvX, line 541: This resolves my question about passing the whole `this.element`. 🙂

CvX (approved):

assets/javascripts/discourse/components/chat-live-pane.js
jjaffeux: ok I see ;) I guess we agree then :D

PR #628, opened 2022-02-23
FEATURE: Allow using [quote] bbcode in chat messages

martin-brennan:
This commit enables the "quotes" markdown feature for
chat messages, and also makes sure the quote always has
a link to the quoted post with the force_quote_link
param for PrettyText.

**This core PR must be merged first: https://github.com/discourse/discourse/pull/16034**

### DIFF
--- a/app/models/chat_message.rb
+++ b/app/models/chat_message.rb
@@ -121,11 +121,12 @@ class ChatMessage < ActiveRecord::Base
     html-img
     mentions
     onebox
+    quotes
+    spoiler-alert
+    table
     text-post-process
     upload-protocol
     watched-words
-    table
-    spoiler-alert
   }
 
   MARKDOWN_IT_RULES = %w{
@@ -145,7 +146,12 @@ class ChatMessage < ActiveRecord::Base
   }
 
   def self.cook(message, opts = {})
-    cooked = PrettyText.cook(message, features_override: MARKDOWN_FEATURES, markdown_it_rules: MARKDOWN_IT_RULES)
+    cooked = PrettyText.cook(
+      message,
+      features_override: MARKDOWN_FEATURES,
+      markdown_it_rules: MARKDOWN_IT_RULES,
+      force_quote_link: true
+    )
 
     result = Oneboxer.apply(cooked) do |url|
       if opts[:invalidate_oneboxes]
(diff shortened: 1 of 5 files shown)


app/models/chat_message.rb
```diff
@@ -121,11 +121,12 @@ def self.uncooked
...
     mentions
     onebox
+    quotes
+    spoiler-alert
+    table
     text-post-process
     upload-protocol
     watched-words
-    table
-    spoiler-alert
```
nattsw, line 128: :D

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -618,6 +618,13 @@ export default Component.extend({
           stagedId: null,
           rateLimited: false,
         });
+
+        // some markdown is cooked differently on the server-side, e.g.
+        // quotes, avatar images etc.
+        if (data.chat_message.cooked !== stagedMessage.cooked) {
+          stagedMessage.set("cooked", data.chat_message.cooked);
+        }
+
```
nattsw, line 627: What would happen if we had just replaced it all the time?

martin-brennan: I think that would be fine, just didn't want to have to do extra work in Ember land if we didn't have to :)

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -618,6 +618,16 @@ export default Component.extend({
           stagedId: null,
           rateLimited: false,
         });
+
+        // some markdown is cooked differently on the server-side, e.g.
+        // quotes, avatar images etc.
+        if (
+          data.chat_message.cooked &&
+          data.chat_message.cooked !== stagedMessage.cooked
+        ) {
```
CvX, line 627:
```suggestion
        if (data.chat_message.cooked !== stagedMessage.cooked) {
```

CvX (approved):

assets/javascripts/discourse/components/chat-live-pane.js
martin-brennan: This won't work, sometimes it is undefined or null for whatever reason! :upside_down_face:

CvX: 👀 I'd like us to know the reason 😅

PR #633, opened 2022-02-23
DEV: Api for Extra Autocomplete

janzenisaac:
Update api with `_applyExtraAutocompletes` to allow for us to hook in to chat with new autocomplete functions


assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -32,6 +32,11 @@ export function addChatToolbarButton(toolbarButton) {
   toolbarButtons.push(toolbarButton);
 }
 
+const _extraAutocompletes = [];
+export function addChatAutocompleteFn(fn) {
+  _extraAutocompletes.push(fn);
```
eviltrout, line 37: Whenever adding a module local API like this we also need a `reset` method that is called after each test run.

PR #638, opened 2022-02-25
DEV: Add tests for onebox rendering within [chat] 

martin-brennan:
This is a companion commit to https://github.com/discourse/discourse/pull/16050
to add tests for the fix there, which was that we weren't hoisting
the html_raw tokens from the custom markdown engine properly, which
caused issues with onebox rendering.

### DIFF
--- a/test/javascripts/acceptance/chat-transcript-test.js
+++ b/test/javascripts/acceptance/chat-transcript-test.js
@@ -4,7 +4,7 @@ import topicFixtures from "discourse/tests/fixtures/topic";
 import { cloneJSON, deepMerge } from "discourse-common/lib/object";
 import QUnit, { test } from "qunit";
 
-import { visit } from "@ember/test-helpers";
+import { click, fillIn, visit } from "@ember/test-helpers";
 import {
   acceptance,
   loggedInUser,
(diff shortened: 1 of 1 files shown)


martin-brennan:
This will not pass until https://github.com/discourse/discourse/pull/16050 is merged

test/javascripts/acceptance/chat-transcript-test.js
```diff
@@ -62,28 +64,36 @@ function generateTranscriptHTML(messageContent, opts) {
...
+      originallySent = originallySent.replace(">#", tabIndexHTML + ">#");
+    }
     transcript.push(`<div class=\"chat-transcript-meta\">
 ${originallySent}</div>`);
   }
 
+  const dateTimeText = opts.showDateTimeText
+    ? moment
+        .tz(opts.datetime, "Australia/Brisbane")
+        .format(I18n.t("dates.long_no_year"))
```
nattsw, line 83: Do we need to `tz` for formatting or will `moment(opts.datetime).format(...)` be good enough?

martin-brennan: @nattsw We use moment.tz for it within the decorator code so I think the tests should be the same. This would work just as well with any other timezone, I just put my own one, it just needs to have something.

CvX (approved):

PR #641, opened 2022-02-25
FEATURE: Auto silence users from flags

markvanlan:


app/models/reviewable_chat_message.rb
```diff
@@ -4,6 +4,26 @@
 
 class ReviewableChatMessage < Reviewable
 
+  def self.on_score_updated(reviewable)
+    # Silence user if new score is over the `score_to_silence_user`
+    return if reviewable.type != self.name
+    return if reviewable.score <= self.score_to_silence_user
+
+    auto_silence_duration = SiteSetting.chat_auto_silence_from_flags_duration
```
jjaffeux, line 12: that's a nitpick but probably the first check we should do ?

spec/models/reviewable_chat_message_spec.rb
```diff
@@ -53,4 +53,30 @@
...
+  describe ".on_score_updated" do
+    it "silences the user for the correct time when the threshhold is met" do
+      SiteSetting.chat_auto_silence_from_flags_duration = 3
+      reviewable.update!(score: ReviewableChatMessage.score_to_silence_user + 1)
+      expect {
+        ReviewableChatMessage.on_score_updated(reviewable)
+      }.to change { user.reload.silenced? }.to be true
+    end
+
+    it "does nothing if the new score is less thatn the score to auto-silence" do
```
jjaffeux, line 66:
```suggestion
    it "does nothing if the new score is less than the score to auto-silence" do
```

app/models/reviewable_chat_message.rb
jjaffeux: I don't know maybe I'm wrong, but I usually put SiteSetting check at top because it's mostly free and I don't have to think about what it could actually call (like score_to_silence_user could actually be a method doing some db call)

spec/models/reviewable_chat_message_spec.rb
```diff
@@ -53,4 +53,30 @@
     expect(reviewable).to be_ignored
     expect(chat_message.reload.deleted_at).not_to be_present
   end
+
+  describe ".on_score_updated" do
+    it "silences the user for the correct time when the threshhold is met" do
```
CvX, line 58:
```suggestion
    it "silences the user for the correct time when the threshold is met" do
```

CvX (approved):

PR #644, opened 2022-02-27
DEV: Standardise chatable_type checks in JS

martin-brennan:
We now create ChatChannel model instances in the chat
JS code, so we can add shorthand methods for checking
the chatable type instead of comparing directly to
a hardcoded string, and in the places where this does
not fit this commit changes the comparison to look
at the CHATABLE_TYPES const.

Also standardises in more places to use instances of
`ChatChannel` instead of `EmberObject`

### DIFF
--- a/assets/javascripts/discourse/models/chat-channel.js
+++ b/assets/javascripts/discourse/models/chat-channel.js
@@ -46,4 +46,20 @@ export default RestModel.extend({
 
     return !READONLY_STATUSES.includes(this.status);
   },
+
+  isDirectMessageChannel() {
+    return this.chatable_type === CHATABLE_TYPES.directMessageChannel;
+  },
+
+  isTopicChannel() {
+    return this.chatable_type === CHATABLE_TYPES.topicChannel;
+  },
+
+  isCategoryChannel() {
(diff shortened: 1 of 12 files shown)


assets/javascripts/discourse/models/chat-channel.js
```diff
@@ -46,4 +46,20 @@ export default RestModel.extend({
...
+    return this.chatable_type === CHATABLE_TYPES.topicChannel;
+  },
+
+  isCategoryChannel() {
+    return this.chatable_type === CHATABLE_TYPES.categoryChannel;
+  },
+
+  isTagChannel() {
+    return this.chatable_type === CHATABLE_TYPES.tagChannel;
+  },
```
CvX, line 64: All these would be better as getters (e.g. `get isDirectMessageChannel() {}`) so they could be used like regular properties (`channel.isDirectMessageChannel`)

martin-brennan: There is no chef's kiss reaction on github, so :chefs_kiss: . Always forget about getters, I haven't used them enough!

martin-brennan:
@CvX I've done those getters and fixed up specs now

lis2 (approved):

PR #645, opened 2022-02-28
DEV: Make topic selection component reusable from move modal

martin-brennan:
The move chat messages to topic modal has the same inner
selection component as the "move posts to topic" modal from
core. This topic selection functionality will be needed for
the archive chat channel functionality.

This commit moves the inner elements into a new chat-to-topic-selector
component, which can be used from any modal component.

For reference it is this modal:

image


tgxworld (approved):
I think we're just shuffling code around here and it looks :+1: to me.

PR #651, opened 2022-03-03
FEATURE: Chat channel archiving

martin-brennan:
This PR introduces the ability for staff to archive chat channels. The flow looks like this:

1. The admin chooses "Archive Channel" from the new settings dropdown
2. The admin selects either a new topic or an existing topic to archive the messages in the channel to
3. We create a `ChatChannelArchive` record in the database to store their selection and the user doing the archiving, and we change the channel status to `read_only` so no messages in the channel can be added or changed
4. We run a background job that creates a post for each 100 messages using the `[chat]` quoting markdown, as well as creating the new topic if that was required. We trash the messages in each batch before proceeding to the next batch.
5. We send a PM to the user who initiated the archive on completion

The job is designed to be idempotent, so if the archive errors halfway through for whatever reason the job can be manually retried by the admin from the UI. If the job fails the admin is sent a PM prompting them to retry. We log errors in the background.

**Note:** If you want to test this process locally, run `be rake chat:make_channel_to_test_archiving\[YOUR_USERNAME\]` to create a new channel for a topic with ~1k messages that you can archive.

image

image

image

image

image

TODO in follow-up PR:

* Mobile channel settings menu with archive button
* Retry UI for failed archive in channel


lib/make_channel_for_archive.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
```
martin-brennan, line 1: Ignore this whole file for review, but if you want to test this archiving functionality you can use this to make topics with a channel and 1000 messages:

`DiscourseChat::MakeChannelForArchive.run("martin")`

martin-brennan: I am not really sure where to put this yet, I just committed it so I could move between my PC and laptop.

assets/javascripts/discourse/components/chat-channel-settings-btn.js
```diff
@@ -0,0 +1,44 @@
...
+import { inject as service } from "@ember/service";
+
+export default Component.extend({
+  tagName: "",
+  channel: null,
+  chat: service(),
+
+  @discourseComputed()
+  channelSettingsOptions() {
+    // TODO (martin): Add closeChannel and deleteChannel options at a later date
```
martin-brennan, line 15: This is for a later PR

lib/tasks/chat_message.rake
```diff
@@ -49,3 +49,86 @@ def rebake_chat_message(chat_message, opts = {})
 rescue => e
   puts "", "Failed to rebake chat message (chat_message_id: #{chat_message.id})", e, e.backtrace.join("\n")
 end
+
+task 'chat:make_channel_to_test_archiving', [:user_for_membership] => :environment do |t, args|
```
martin-brennan, line 53: Not sure if this part really needs review, it's just to make test channels in local development.

app/controllers/chat_channels_controller.rb
```diff
@@ -184,6 +184,38 @@ def search
...
+    params.require(:type)
+
+    topic_type = params[:type] == "newTopic" ? :new : :existing
+    topic_params_ok = \
+      (topic_type == :existing && params[:topic_id].present?) ||
+      (topic_type == :new && params[:title].present?)
+
+    raise Discourse::InvalidParameters if !topic_params_ok
+
+    chat_channel = ChatChannel.find_by(id: params[:chat_channel_id])
```
lis2, line 198: Maybe let's use find_by! here so next `raise` will be not required

app/controllers/chat_channels_controller.rb
```diff
@@ -184,6 +184,38 @@ def search
...
   end
 
+  def archive
+    params.require(:chat_channel_id)
+    params.require(:type)
+
+    topic_type = params[:type] == "newTopic" ? :new : :existing
+    topic_params_ok = \
+      (topic_type == :existing && params[:topic_id].present?) ||
+      (topic_type == :new && params[:title].present?)
```
lis2, line 194: that topic_type is not used further down in the method, so I don't see the benefit of that assignment

```
    topic_params_ok = params[:type] == "newTopic" ? params[:title].present? : params[:topic_id].present?
```
or maybe even 
```
if params[:type] == "newTopic" ? params[:title].present? : params[:topic_id].present?
  raise Discourse::InvalidParameters 
end
```

assets/javascripts/discourse/components/chat-channel-row.js
```diff
@@ -17,6 +17,15 @@ export default Component.extend({
   active: propertyEqual("channel.id", "chat.activeChannel.id"),
   options: null,
 
+  init() {
+    this._super(...arguments);
+    this.options = this.options || {};
+    this.set(
+      "showSettingsButton",
+      this.options.settingsButton && this.currentUser.staff
```
lis2, line 25: Is my understanding right that archive is the only option right now?

And later that options is controller by:
```
    this.siteSettings.chat_allow_archiving_channels &&
      !this.channel.isArchived &&
      !this.channel.isReadOnly
```

It means that when channel is currently archiving, then setting button is visible but nothing really happens on click. 

I see that later we will have option to mark channel as readonly and delete but maybe we should also check if any options are available to decide if button should be visible or not

lib/chat_channel_archive_service.rb
```diff
@@ -0,0 +1,200 @@
...
+        # automatic background process
+        import_mode: true,
+
+        # don't want to be stopped by watched word or post length validations
+        skip_validations: true,
+
+        topic_id: chat_channel_archive.destination_topic_id
+      )
+
+      pc.create
```
lis2, line 108: I am not sure, but when we create a new post for archive, do we want to relink other models like user_actions, uploads etc?

Something which is done by PostMover https://github.com/discourse/discourse/blob/main/app/models/post_mover.rb#L81;L88

lis2: If we decided to move those items, then it should be done later in `delete_message_batch`

app/jobs/regular/chat_channel_archive.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Jobs
+  class ChatChannelArchive < ::Jobs::Base
+    sidekiq_options retry: false
+
+    def execute(args = {})
+      channel_archive = ::ChatChannelArchive.find_by(id: args[:chat_channel_archive_id])
+      return if channel_archive.blank?
```
lis2, line 9: When ChatChannelArchive can be blank? If the record is blank, doesn't it mean that something bad happened and here we are suppressing that?

app/controllers/chat_channels_controller.rb
martin-brennan: I think we use `Discourse::NotFound` as a standard though, it's through core a lot and even within the chat controllers this pattern

app/controllers/chat_channels_controller.rb
martin-brennan: Good point, I think I planned to use that type at some point but didn't

assets/javascripts/discourse/components/chat-channel-row.js
martin-brennan: In the template for `chat-channel-settings-btn` I don't show the button if there are no options:

https://github.com/discourse-org/discourse-topic-chat/blob/b8266e0ad02c297b8e2c7c8138656f6179854542/assets/javascripts/discourse/templates/components/chat-channel-settings-btn.hbs#L1-L8

lib/chat_channel_archive_service.rb
martin-brennan: I don't think there is anything to relink really...I already checked that uploads are linked to the post correctly in this spec:

https://github.com/discourse-org/discourse-topic-chat/blob/b8266e0ad02c297b8e2c7c8138656f6179854542/spec/lib/chat_channel_archive_service_spec.rb#L112-L118

app/jobs/regular/chat_channel_archive.rb
martin-brennan: Yeah, I will throw a warning in there, shouldn't really happen but better than throwing errors in the job :)

lib/chat_channel_archive_service.rb
lis2: Thank you, I understand it now :)

lis2 (approved):

lib/chat_transcript_service.rb
```diff
@@ -60,9 +60,15 @@ def channel_attr
     end
   end
 
-  def initialize(channel, message_ids)
+  def initialize(channel, message_ids: [], messages: [])
```
lis2, line 63: I have mixed feeling about that. I love how it is cached so when messages are passed then no additional queries are sent to the database. 

The thing I don't like is that if I want to use that class quickly and just read initializer without details, I may use it as:
`ChatTranscriptService.new("channel", messages: [Message.last], message_ids: [1, 2, 3])`

I have seen before helper initializers to avoid that problem, something like:
```ruby
private_class_method :new
def self.create_with_message_ids(channel, ids)
  new(channel, message_ids: ids)
end
def self.create_with_messages(channel, messages)
  new(channel, messages: messages)
end
```

Not sure if that would be better, so I am leaving decision to you :)

martin-brennan: Yeah fair enough...another way I could do it is like this:

```ruby
def initializer(channel, messages_or_ids:)
  if messages_or_ids.all? {|m| m.is_a? Numeric}
    @message_ids = messages_or_ids
  else
    @messages = messages_or_ids
  end
end
```

Might make that change actually, it is a bit neater.

assets/javascripts/discourse/components/chat-channel-archive-modal-inner.js
```diff
@@ -0,0 +1,103 @@
...
+          text: I18n.t("chat.channel_archive.process_started"),
+          messageClass: "success",
+        });
+        later(() => {
+          if (!isTesting()) {
+            window.location.reload();
+          }
+        }, 3000);
+      })
+      .catch((error) => popupAjaxError(error))
```
jjaffeux, line 52:
```suggestion
      catch(popupAjaxError)
```

jjaffeux, line 52:
```suggestion
      .catch(popupAjaxError)
```

assets/javascripts/discourse/components/chat-channel-row.js
```diff
@@ -17,6 +17,15 @@ export default Component.extend({
   active: propertyEqual("channel.id", "chat.activeChannel.id"),
   options: null,
 
+  init() {
+    this._super(...arguments);
+    this.options = this.options || {};
```
jjaffeux, line 22: any reason you are not using set here ?

assets/javascripts/discourse/components/chat-channel-row.js
```diff
@@ -59,6 +68,13 @@ export default Component.extend({
       return;
     }
 
+    if (
```
jjaffeux, line 71: could you explain the reason of this please ? I didn't dig but that seems hacky

assets/javascripts/discourse/components/chat-channel-row.js
martin-brennan: Nah, will change

assets/javascripts/discourse/components/chat-channel-row.js
martin-brennan: Full context:

```javascript
@action
handleClick(event) {
  if (event.target.classList.contains("chat-channel-leave-btn")) {
    return;
  }

  if (
    event.target.classList.contains("chat-channel-settings-btn") ||
    event.target.parentElement.classList.contains("select-kit-header-wrapper")
  ) {
    return;
  }

  this.handleSwitchChannel(event);
},
```

I am not a fan of this either, but without this special case I added the select-kit menu does not show when you click on this cog on hover:

image

Also, I assume the first condition is to make this X button work:

image

If there is a way of making these both work without these weird class overrides I can change it, just wasn't sure how else to make the select-kit menu show up correctly on click in this situation.

assets/javascripts/discourse/components/chat-channel-settings-btn.js
```diff
@@ -0,0 +1,44 @@
...
+
+    return options;
+  },
+
+  @action
+  handleChannelSettingsAction(id) {
+    if (!this.currentUser.staff) {
+      return;
+    }
+    this[id]();
```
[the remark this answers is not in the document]
martin-brennan: Yes fair enough, I will change this and the other place we are doing this (channel-list)

PR #658, opened 2022-03-04
PERF: removes two unnecessary enclosing divs for each message

jjaffeux:


jjaffeux:
This will make the DOM structure of chat messages much cleaner.

assets/javascripts/discourse/components/chat-message.js
```diff
@@ -528,11 +539,15 @@ export default Component.extend({
   },
 
   _repositionEmojiPicker(btn, position) {
-    if (!this.element) {
+    const element = document.querySelector(
+      `.chat-message-container[data-id='${this.message.id}']`
+    );
```
CvX, line 544: Maybe this should be extracted to `get element() {}`?

jjaffeux: ok :+1:

assets/javascripts/discourse/components/chat-message.js
```diff
@@ -43,11 +45,13 @@ export default Component.extend({
...
-    if (!this.currentUser) {
+
+    if (!this.currentUser || !this.message) {
       return;
     }
-    this.element
-      .querySelector(".chat-message-content .chat-message-text")
+
+    document
+      .querySelector(`.chat-message-container[data-id='${this.message.id}']`)
```
CvX, line 54:
```suggestion
    this.element
```

CvX (approved):

PR #659, opened 2022-03-04
 FEATURE: Auto-focus composer on key press in full page

markvanlan:


assets/javascripts/discourse/components/full-page-chat.js
```diff
@@ -46,10 +47,31 @@ export default Component.extend({
...
     this.appEvents.off("chat:refresh-channels", this, "refreshModel");
     window.removeEventListener("resize", this._calculateHeight, false);
+    document.removeEventListener("keydown", this._autoFocusChatComposer);
     document.body.classList.remove("has-full-page-chat");
     this.chat.set("fullScreenChatOpen", false);
   },
 
+  @bind
+  _autoFocusChatComposer(e) {
+    if (!e.target || e.target.classList.contains("chat-composer-input")) {
```
CvX, line 57: Won't this auto-focus logic trigger when using the search dropdown or when typing into a modal?

markvanlan: hmm yeah I'm curious how to do this safely 🤔

markvanlan: I can't have a list of elements to avoid b/c that will break as things are added. But maybe a list of selectors where if the activeElement is in X array or selectors... then we trigger this? IDK

davidtaylorhq:
Will this break keyboard shortcuts in chat? e.g. `g` `l` will now put me in the chat composer, rather than going to `/latest` 🤔

assets/javascripts/discourse/components/full-page-chat.js
CvX: You could check if the active element is an input/textarea/etc. but still, David's concern about global app shortcuts is valid

markvanlan:
Yes @davidtaylorhq it will break them intentionally. I guess maybe we only want to implement this if chat is isolated? Not totally sure but yes, the breaking of that is intentional.

assets/javascripts/discourse/components/full-page-chat.js
```diff
@@ -46,10 +47,36 @@ export default Component.extend({
...
+    if (!e.key || e.key.length > 1 || e.metaKey || e.ctrlKey) {
+      return; // Only care about single characters, unlike `Escape`
+    }
+    const target = e.target;
+    if (!target || target.classList.contains("chat-composer-input")) {
+      return;
+    }
+
+    const lowerTagName = target.tagName.toLowerCase();
+    if (lowerTagName === "input" || lowerTagName === "textarea") {
```
CvX, line 66:
```suggestion
    if (!target) {
      return;
    }

    if (/^(INPUT|TEXTAREA|SELECT)$/.test(target.tagName)) {
```

CvX (approved):

PR #660, opened 2022-03-07
FIX: Show deleted messages if can moderate

martin-brennan:
When a user can moderate a channel, we allow the API to
return deleted messages in the JSON payload, but we do
not show the messages in the UI with the "A message was
deleted" message and [view] button. This PR corrects this
by providing the can_moderate boolean in the chat view
meta payload, and using that to determine whether to
show the message alongside existing conditions.


nattsw (approved):
Cool thanks!

PR #661, opened 2022-03-07
DEV: Fix chat_channel timestamp migrations

martin-brennan:
The way the fix from 31de9ae70a021e32fee92b343f392e524f32c9b5
was applied for sites that have never had chat enabled left
ActiveRecord in a weird state until the next deploy, where
there were errors raised saying "created_at can not be null".
Presumably this is because of "Rails Magic TM" that is fixed
on a subsequent deploy. This commit attempts to fix the original
migration even further, so we should not encounter this again
when turning on chat for sites in future.


nattsw (approved):
Looks good.

There exists

```
change_table :chat_channels do |t|
  t.timestamps
end
```

(i believe) that we can use next time for other tables if we have the same scenario

martin-brennan:
Oh that's cool I didn't know that! Thanks for the review :+1:

PR #662, opened 2022-03-07
FEATURE: Add chat_default_channel_id site setting

martin-brennan:
This setting will make the user navigate to the selected
public channel when opening chat IF they have no unread
messages and no mentions that they need to see. If there is
no default channel set, then the first public channel will
be shown, otherwise the first DM channel will be shown.

### DIFF
--- a/assets/javascripts/discourse/services/chat.js
+++ b/assets/javascripts/discourse/services/chat.js
@@ -346,16 +346,22 @@ export default Service.extend({
 
   getIdealFirstChannelId() {
     // When user opens chat we need to give them the 'best' channel when they enter.
+    //
     // Look for public channels with mentions. If one exists, enter that.
     // Next best is a DM channel with unread messages.
     // Next best is a public channel with unread messages.
+    // Then we fall back to the chat_default_channel_id site setting
+    // if that is present and in the list of channels the user can access.
+    // If none of these options exist, then we get the first public channel,
+    // or failing that the first DM channel.
     return this.getChannels().then(() => {
       // Defined in order of significance.
       let publicChannelWithMention,
         dmChannelWithUnread,
         publicChannelWithUnread,
         publicChannel,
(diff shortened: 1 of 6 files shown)


assets/javascripts/discourse/services/chat.js
```diff
@@ -372,6 +378,11 @@ export default Service.extend({
             break; // <- We have a public channel with a mention. Break and return this.
           } else if (!publicChannelWithUnread && state.unread_count > 0) {
             publicChannelWithUnread = channel;
+          } else if (
+            !defaultChannel &&
```
nattsw, line 382: Is this defaultChannelId?

I don't think we will ever have 0 but 0 is falsy.

config/settings.yml
```diff
@@ -41,3 +41,7 @@ plugins:
   chat_auto_silence_from_flags_duration:
     default: 60
     min: 0
+  chat_default_channel_id:
+    default: ""
```
nattsw, line 45: I suppose this equates to a null

plugin.rb
```diff
@@ -33,6 +33,9 @@
 # route: /admin/plugins/chat
 add_admin_route 'chat.admin.title', 'chat'
 
+# Site setting validators must be loaded before initialize
```
nattsw, line 36: Thanks for the note.

nattsw (approved):
Looks good 👍 

If you have time would be good for a test backfill another day.

nattsw:
Looking at test failure

assets/javascripts/discourse/services/chat.js
```diff
@@ -372,6 +378,11 @@ export default Service.extend({
...
+            this.siteSettings.chat_default_channel_id == channel
```
nattsw, line 383: Aha eslint hates `==`

PR #665, opened 2022-03-08
FIX: Revert chat-composer keyUp to keyDown event

martin-brennan:
In 009f7321bf83699ce0997b04a35b68ab85757717 we changed the composer
key event to keyup instead of keydown. This does not work correctly
for the composer's case though, because we want to prevent the default
behaviour of Enter if we do not hit certain special cases, such as
when we want to a) send the message or b) save the edit to a message.
On keyup, these special events still worked, but they would insert
an additional \n in the message because the Enter key event already
happened.


gabenp (approved):

CvX:
> we want to prevent the default behaviour of Enter

@martin-brennan is it possible to test that? 😃

PR #673, opened 2022-03-09
FIX: Archive channel refinements

martin-brennan:
* Make all users leave the channel when it gets archived
* Reload the channel window when the message bus channel
  status change comes through, but only for the active
  channel and only when the change is to read only or archived

### DIFF
--- a/assets/javascripts/discourse/services/chat.js
+++ b/assets/javascripts/discourse/services/chat.js
@@ -8,6 +8,7 @@ import { generateCookFunction } from "discourse/lib/text";
 import { next } from "@ember/runloop";
 import { Promise } from "rsvp";
 import ChatChannel, {
+  CHANNEL_STATUSES,
   CHATABLE_TYPES,
 } from "discourse/plugins/discourse-chat/discourse/models/chat-channel";
 import simpleCategoryHashMentionTransform from "discourse/plugins/discourse-chat/discourse/lib/simple-category-hash-mention-transform";
@@ -61,6 +62,7 @@ export default Service.extend({
       this._subscribeToNewDmChannelUpdates();
       this._subscribeToUserTrackingChannel();
       this._subscribeToChannelEdits();
+      this._subscribeToChannelStatusChange();
       this.presenceChannel = this.presence.getChannel("/chat/online");
       this.draftStore = {};
 
@@ -80,6 +82,7 @@ export default Service.extend({
       this._unsubscribeFromNewDmChannelUpdates();
       this._unsubscribeFromUserTrackingChannel();
       this._unsubscribeFromChannelEdits();
+      this._unsubscribeFromChannelStatusChange();
       this._unsubscribeFromAllChatChannels();
     }
   },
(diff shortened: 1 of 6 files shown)


assets/javascripts/discourse/services/chat.js
```diff
@@ -545,6 +548,27 @@ export default Service.extend({
...
+      if (
+        busData.chat_channel_id === activeChannel?.id &&
+        (busData.status === CHANNEL_STATUSES.readOnly ||
+          busData.status === CHANNEL_STATUSES.archived)
+      ) {
+        // This is not very elegant. Ideally at some point we want to have
+        // a nice reactive magical transformation of the channel status
+        // before the user's very eyes...but for now let's just reload so
+        // they can see they are no longer allowed to chat.
+        window.location.reload();
```
nattsw, line 563: I understand your concern: Do you think you could log this somewhere, or inform @markvanlan about this so he could add it in a list of improvements to be made?

assets/javascripts/discourse/services/chat.js
```diff
@@ -545,6 +548,27 @@ export default Service.extend({
     });
   },
 
+  _subscribeToChannelStatusChange() {
+    this.messageBus.subscribe("/chat/channel-status", (busData) => {
+      let activeChannel = this.getActiveChannel();
+      if (
+        busData.chat_channel_id === activeChannel?.id &&
+        (busData.status === CHANNEL_STATUSES.readOnly ||
+          busData.status === CHANNEL_STATUSES.archived)
```
nattsw, line 557: Do we do nothing if it's not the active channel? Like visual greying out of the channel or something in the sidebar.

assets/javascripts/discourse/templates/components/chat-live-pane.hbs
```diff
@@ -9,9 +9,10 @@
         }}
       {{/if}}
 
-      {{chat-channel-title onClick=(action "onChannelTitleClick") channel=chatChannel}}
-
-      {{chat-channel-status channel=chatChannel}}
+      <div class="chat-channel-title-with-status">
+        {{chat-channel-title onClick=(action "onChannelTitleClick") channel=chatChannel}}
+        {{chat-channel-status channel=chatChannel}}
+      </div>
```
nattsw, line 15: Don't have to do it now, but usually when I see a well encapsulated `<div>` I think of a new component.

assets/javascripts/discourse/services/chat.js
```diff
@@ -545,6 +548,27 @@ export default Service.extend({
     });
   },
 
+  _subscribeToChannelStatusChange() {
+    this.messageBus.subscribe("/chat/channel-status", (busData) => {
```
nattsw, line 552: I'm not sure if you can do it (I swear I thought I saw you do it before) but is there any way we can test this?

One day if we remove that `window.location.reload`, the person who can make the change can hitch on the test.

nattsw (approved):
Looks good overall though do respond to the questions.

assets/javascripts/discourse/services/chat.js
martin-brennan: I did mention it in the internal archive topic, however I will make a new one for this specific todo :)

assets/javascripts/discourse/services/chat.js
martin-brennan: I don't think it's worth it, if you open that channel it completely reloads anyway so you will see the new status :) That would fit in with the "fancy reactive channel status change" v2 feature I will make the TODO for

assets/javascripts/discourse/templates/components/chat-live-pane.hbs
martin-brennan: Yeah I am adding a third component here in my other PR, we may need a todo to slice up chat-live-pane internally

assets/javascripts/discourse/services/chat.js
martin-brennan: Actually yeah I think I can test it, however we usually wrap `window.location.reload` with `if (!isTesting())`, so I will fix that, but there would be nothing really to test then

PR #674, opened 2022-03-09
FEATURE: Retry failed archive

martin-brennan:


app/serializers/chat_channel_serializer.rb
```diff
@@ -45,4 +50,52 @@ def chatable
...
+  def include_archive_completed?
+    include_archive_status?
+  end
+
+  def archive_completed
+    archive.complete?
+  end
+
+  def include_archive_failed?
+    include_archive_status?
```
lis2, line 71: In many serializes in core, I see we are using `alias_method`

lis2 (approved):

app/serializers/chat_channel_serializer.rb
martin-brennan: Ah yeah that would be better, forgot about that!

PR #675, opened 2022-03-09
FEATURE: new setting to define default emoji reactions

techAPJ:
FEATURE: add dedicated emoji store for chat plugin

This commit adds a new setting to define default emoji reactions for chat messages. The setting defaults to 3 reactions.

Screen Shot 2022-03-09 at 21 32 22

This commit also adds a new emoji store for chat plugin so that the favorites (recently used) emojis in chat emoji-picker is independent of topics/posts.

### DIFF
--- a/assets/javascripts/discourse/services/chat-emoji-store.js
+++ b/assets/javascripts/discourse/services/chat-emoji-store.js
@@ -1,0 +1,61 @@
+// This class is duplicated from emoji-store class in core with an addition of `reactions` property.
+// We want to maintain separate emoji store for chat plugin.
+// https://github.com/discourse/discourse/blob/892f7e0506f3a4d40d9a59a4c926ff0a2aa0947e/app/assets/javascripts/discourse/app/services/emoji-store.js
+
+import KeyValueStore from "discourse/lib/key-value-store";
+import Service from "@ember/service";
+
+const EMOJI_USAGE = "emojiUsage";
+const EMOJI_SELECTED_DIVERSITY = "emojiSelectedDiversity";
+const TRACKED_EMOJIS = 15;
+const STORE_NAMESPACE = "discourse_chat_emojis_";
+
+export default Service.extend({
+  init() {
+    this._super(...arguments);
+
+    this.store = new KeyValueStore(STORE_NAMESPACE);
+
+    if (!this.store.getObject(EMOJI_USAGE)) {
+      this.favorites = [];
+    }
+  },
+
+  get diversity() {
+    return this.store.getObject(EMOJI_SELECTED_DIVERSITY) || 1;
+  },
+
+  set diversity(value) {
+    this.store.setObject({ key: EMOJI_SELECTED_DIVERSITY, value: value || 1 });
+  },
+
+  get reactions() {
+    if (!this.siteSettings.default_emoji_reactions) {
+      return [];
+    }
+    return this.siteSettings.default_emoji_reactions.split("|").filter(Boolean);
+  },
+
+  get favorites() {
+    return this.store.getObject(EMOJI_USAGE) || [];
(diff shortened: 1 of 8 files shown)


techAPJ:
The frontend tests are passing locally... both via browser and in terminal. 🤔 

```
↪ Discourse Chat | Unit | chat-emoji-store::diversity [✔]
```

assets/javascripts/discourse/services/chat-emoji-store.js
```diff
@@ -0,0 +1,57 @@
+import KeyValueStore from "discourse/lib/key-value-store";
+import Service from "@ember/service";
+
+const EMOJI_USAGE = "emojiUsage";
+const EMOJI_SELECTED_DIVERSITY = "emojiSelectedDiversity";
+const TRACKED_EMOJIS = 15;
+const STORE_NAMESPACE = "discourse_chat_emojis_";
+
+export default Service.extend({
```
martin-brennan, line 9: I wonder if we could just import the `EmojiStore` service from core here and do `Service.extend(EmojiStore)`? That would save having to copy all this stuff from core

martin-brennan: Oh I see that `get reactions()` is new here too, maybe like this?

```javascript
Service.extend(EmojiStore, {
   get reactions() {
    if (!this.siteSettings.default_emoji_reactions) {
      return [];
    }
    return this.siteSettings.default_emoji_reactions.split("|").filter(Boolean);
  },
})
```

martin-brennan (commented):
Looks great apart from one thing with the store :)

martin-brennan:
Also the way I read this requirement:

> Add specific store for discourse chat reactions, so users can amend the quick access list over time. Stop using the core favorite list.

Is that the list of reactions should _start_ with these defaults but then still use people's favourites from chat over time, so I think maybe instead of `get reactions()` we should just do something like this:

```javascript
 if (!this.store.getObject(EMOJI_USAGE)) {
    if (!this.siteSettings.default_emoji_reactions) {
        this.favorites = [];
    } else {
        this.favorites = this.siteSettings.default_emoji_reactions.split("|").filter(Boolean);
    }
}
```

jordanvidrine:
> _start_ with these defaults but then still use people's favourites from chat over time

Just wanting to make sure, by peoples favorites... we *do* mean actual reactions to messages right? And *not* emojis used in the chat interface?

assets/javascripts/discourse/services/chat-emoji-store.js
techAPJ: Thanks for the suggestion! I tried doing this but it seems like it's not possible to import `EmojiStore` service from core.

techAPJ:
@martin-brennan I believe we should stick to default emoji reactions instead of favourites because as @jordanvidrine pointed out the favs will also include emojis from chat messages and not just reactions.

The benefit of having a dedicated emoji-store in chat will be that the "Recently used" section in emoji-picker in chat will be independent of core discourse. So only emojis used in chat interface will be shown in chat's emoji-picker.

martin-brennan:
@techAPJ oh I see I was talking about this part:

image

But if the recently used section inside the "more emojis" button updates based on what you are using in chat then that's all good :+1:

assets/javascripts/discourse/services/chat-emoji-store.js
martin-brennan: Yep no problem just need the comment in this class explaining it's duplicated from core for future devs

techAPJ: Done.

martin-brennan (approved):

PR #678, opened 2022-03-10
FIX: Improvements to archiving tool

martin-brennan:
* When generating markdown for the archive, do not create links
  to individual chat messages, because these will just show a
  404 message when clicked.
* Add a site setting to determine what to change the status of
  the archive destination topic to. Before it was always changed
  to archived, new options are open, archived, closed. This only
  affects new destination topics.
* Hide the chat icon next to the topic title if the related channel
  is read_only or archived.
* Fix some issues in chat-setup initializer -- we were getting the chat
  service twice, and also the chat quote datetime decorator would not
  work for users without chat enabled. This is wrong because the post
  decorations should always work.

### DIFF
--- a/assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
+++ b/assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
@@ -103,9 +103,7 @@ const chatTranscriptRule = {
     // chat message because it will just result in a 404
     if (noLink) {
       let spanToken = state.push("span_open", "span", 1);
-      spanToken.attrs = [
-        ["title", messageTimeStart],
-      ];
+      spanToken.attrs = [["title", messageTimeStart]];
 
       spanToken.block = false;
       spanToken = state.push("span_close", "span", -1);
--- a/assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
+++ b/assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
@@ -12,12 +12,12 @@ const chatTranscriptRule = {
     }
 
     const options = state.md.options.discourse;
-    let [username, messageIdStart, messageTimeStart] =
+    const [username, messageIdStart, messageTimeStart] =
       (tagInfo.attrs.quote && tagInfo.attrs.quote.split(";")) || [];
-    let multiQuote = !!tagInfo.attrs.multiQuote;
-    let noLink = !!tagInfo.attrs.noLink;
-    let channelName = tagInfo.attrs.channel;
-    let channelLink = channelName
+    const multiQuote = !!tagInfo.attrs.multiQuote;
+    const noLink = !!tagInfo.attrs.noLink;
+    const channelName = tagInfo.attrs.channel;
+    const channelLink = channelName
       ? options.getURL(
           `/chat/chat_channels/${encodeURIComponent(channelName.toLowerCase())}`
         )
--- a/assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
+++ b/assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
@@ -15,6 +15,7 @@ const chatTranscriptRule = {
     let [username, messageIdStart, messageTimeStart] =
       (tagInfo.attrs.quote && tagInfo.attrs.quote.split(";")) || [];
     let multiQuote = !!tagInfo.attrs.multiQuote;
+    let noLink = !!tagInfo.attrs.noLink;
     let channelName = tagInfo.attrs.channel;
     let channelLink = channelName
       ? options.getURL(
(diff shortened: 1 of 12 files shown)


nattsw:
> When generating markdown for the archive

Can you share with me what this means?

assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
```diff
@@ -15,6 +15,7 @@ const chatTranscriptRule = {
     let [username, messageIdStart, messageTimeStart] =
       (tagInfo.attrs.quote && tagInfo.attrs.quote.split(";")) || [];
     let multiQuote = !!tagInfo.attrs.multiQuote;
+    let noLink = !!tagInfo.attrs.noLink;
```
nattsw, line 18: Actually I have no idea why everything's `let` in this chunk given they're not mutated. 🤔

assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
```diff
@@ -98,15 +99,26 @@ const chatTranscriptRule = {
...
+    // chat message because it will just result in a 404
+    if (noLink) {
+      let spanToken = state.push("span_open", "span", 1);
+      spanToken.attrs = [["title", messageTimeStart]];
 
-    linkToken.block = false;
-    linkToken = state.push("link_close", "a", -1);
-    linkToken.block = false;
+      spanToken.block = false;
+      spanToken = state.push("span_close", "span", -1);
```
nattsw, line 109: Hmm you're reassigning spanToken. is this intended?

assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
```diff
@@ -98,15 +99,26 @@ const chatTranscriptRule = {
...
+      spanToken.block = false;
+    } else {
+      let linkToken = state.push("link_open", "a", 1);
+      linkToken.attrs = [
+        ["href", options.getURL(`/chat/message/${messageIdStart}`)],
+        ["title", messageTimeStart],
+      ];
+
+      linkToken.block = false;
+      linkToken = state.push("link_close", "a", -1);
```
nattsw, line 119: Same here.

martin-brennan:
@nattsw all the `let` usages are basically unintentional, will change to const where it makes sense :)

config/locales/server.en.yml
```diff
@@ -11,6 +11,7 @@ en:
     chat_auto_silence_from_flags_duration: "Number of minutes that users will be silenced for when they are automatically silenced due to flagged chat messages."
     chat_default_channel_id: "The chat channel that will be opened by default when a user has no unread messages or mentions in other channels."
     chat_duplicate_message_sensitivity: "The likelihood that a duplicate message by the same sender will be blocked in a short period. Decimal number between 0 and 1.0, with 1.0 being the highest setting (blocks messages more frequently in a shorter amount of time). Set to `0` to allow duplicate messages."
+    chat_archive_destination_topic_status: "The status that the destination topic should be once a channel archive is completed."
```
nattsw, line 14: Just to check, we **do** want this much flexibility?

Trying to think of cases where a this site setting will cause some behaviour we don't want as opposed to something done manually.

Like setting is "closed", and someone says on chat "let's take this chat to the existing topic". But archiving the chat would cause the existing topic to close..

config/settings.yml
```diff
@@ -45,6 +45,13 @@ plugins:
     default: false
     hidden: true
     client: true
+  chat_archive_destination_topic_status:
+    type: enum
+    default: archived
```
nattsw, line 50: Would it make sense to keep it open, then direct the admin to the topic after "archiving", to decide what to do with it manually?

config/locales/server.en.yml
martin-brennan: > someone says on chat "let's take this chat to the existing topic"

Hmm I don't think generally you'd be archiving the whole channel if someone said that. You can just quote the relevant messages into the existing topic.

I also think in general archiving into an existing topic is kind of weird, but that's a different issue really.

config/settings.yml
nattsw: Cool, can ignore this comment based on above

martin-brennan: I think it's best to have an automatic behaviour. By default it's changing the target topic to archived, which I think makes the most sense because you're creating a static record of chat in a point of time. However, maybe if the archive is being done to an existing topic we just skip this status change?

nattsw: > if the archive is being done to an existing topic

aye, that was the use case I was thinking.

martin-brennan: I will make that change then, there will be no unexpected side effects then :) Will update the site setting text to match

nattsw (approved):

assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
nattsw: Also the array `.push`

> The push() method adds one or more elements to the end of an array and returns the new length of the array.

I'm guessing this is a different `push`?

lib/chat_channel_archive_service.rb
```diff
@@ -148,10 +148,15 @@ def ensure_destination_topic_exists!
   end
 
   def update_destination_topic_status
-    if SiteSetting.chat_archive_destination_topic_status == "archived"
-      chat_channel_archive.destination_topic.update!(archived: true)
-    elsif SiteSetting.chat_archive_destination_topic_status == "closed"
-      chat_channel_archive.destination_topic.update!(closed: true)
+    # we only want to do this when the destination topic is new, not an
+    # existing topic, because we don't want to update the status unexpectedly
+    # on an existing topic
```
nattsw, line 153: 👍 thanks!!

PR #679, opened 2022-03-10
FEATURE: Close/open chat channel

martin-brennan:
This commit adds the ability for admins to close and open a chat channel. Closing a channel prevents non-staff users from sending messages and from editing existing messages.


assets/javascripts/discourse/components/chat-channel-settings-btn.js
```diff
@@ -44,4 +58,18 @@ export default Component.extend({
...
+      this.channel
+    );
+  },
+
+  openChannel() {
+    showModal("chat-channel-toggle-open-modal").set(
+      "chatChannel",
+      this.channel
+    );
+  },
```
techAPJ, line 74: It looks like both `closeChannel` and `openChannel` actions are doing the same thing. Is it possible to merge them into one action `toggleChannel` instead?

techAPJ (approved):
Looks good to me!

assets/javascripts/discourse/components/chat-channel-settings-btn.js
martin-brennan: Oh yeah makes sense...I probably made these at the start then forgot to update :sweat_smile:

app/controllers/chat_channels_controller.rb
```diff
@@ -240,6 +241,25 @@ def retry_archive
     render json: success_json
   end
 
+  def toggle_open_status
```
tgxworld, line 244: Something to consider here is whether we should just let the client tell us what state it wants the channel to be in. Toggling assumes the client's impression of the state and the server's state is the same which might not always be the case. I've been bitten by this in the past with regards to the toggling of status for topics.

app/controllers/chat_channels_controller.rb
```diff
@@ -240,6 +241,25 @@ def retry_archive
...
+    raise Discourse::NotFound if chat_channel.blank?
+
+    if chat_channel.open?
+      guardian.ensure_can_change_channel_status!(chat_channel, :closed)
+      chat_channel.close!(current_user)
+    elsif chat_channel.closed?
+      guardian.ensure_can_change_channel_status!(chat_channel, :open)
+      chat_channel.open!(current_user)
+    else
+      raise Discourse::InvalidAccess
```
tgxworld, line 257: Is it possible for a channel to have any other state?

tgxworld:
Did we consider testing the client side changes?

assets/javascripts/discourse/components/chat-channel-toggle-open-modal-inner.js
```diff
@@ -0,0 +1,58 @@
...
+import { ajax } from "discourse/lib/ajax";
+import { inject as service } from "@ember/service";
+import { popupAjaxError } from "discourse/lib/ajax-error";
+
+export default Component.extend({
+  chat: service(),
+  tagName: "",
+  chatChannel: null,
+
+  @discourseComputed()
```
tgxworld, line 15: Does this need to be a computed property when it isn't computing anyting?

spec/requests/chat_channel_controller_spec.rb
```diff
@@ -610,4 +610,50 @@
...
+    end
+
+    it "returns error if user is not admin" do
+      sign_in(user)
+      put "/chat/chat_channels/#{channel.id}/toggle_open_status.json"
+      expect(response.status).to eq(403)
+    end
+
+    it "returns a 404 if the channel does not exist" do
+      channel.destroy
```
tgxworld, line 631:
```suggestion
      channel.destroy!
```

app/controllers/chat_channels_controller.rb
tgxworld: O nevermind this :) I just saw the specs.

martin-brennan: Yes, archived and read_only, those are only done via the `archive` endpoint though. I will change to dedicated endpoints as your other comment suggests.

app/controllers/chat_channels_controller.rb
martin-brennan: Yeah thanks for the context, I will change to dedicated endpoints.

assets/javascripts/discourse/components/chat-channel-toggle-open-modal-inner.js
martin-brennan: A philosophical question for the ages :) Will change to a simple `get X`

martin-brennan:
> Did we consider testing the client side changes?

@tgxworld no i did not...will reconsider and add a test ;)

app/controllers/chat_channels_controller.rb
tgxworld: Instead of `toggle_state`, how about `change_state`? We can keep it as a single endpoint in that case.

martin-brennan: Yeah makes sense, good idea :+1:

martin-brennan:
Failing test is unrelated, need to investigate that separately.

PR #687, opened 2022-03-13
DEV: `settled` works in the legacy env now

CvX:


jjaffeux:
What do we do with this @CvX  ?

jjaffeux:
:hugs: ?

CvX:
The premise of this PR was wrong, it turned out. 😉 

We are dropping the support for the legacy ember env, so we'd want this in, but we/I can redo this change then. 🙂 Closing for now!

PR #688, opened 2022-03-14
DEV: Actually fix the test flakyness

CvX:
#685 didn't do anything…

The issue was that the function passed to `decorateCookedElement()` kept the `currentUser` (or `container`, after the just merged PR) from the first time the initalizer was executed.

### DIFF
--- a/assets/javascripts/discourse/initializers/chat-setup.js
+++ b/assets/javascripts/discourse/initializers/chat-setup.js
@@ -1,11 +1,11 @@
 import { withPluginApi } from "discourse/lib/plugin-api";
 import I18n from "I18n";
 import { bind } from "discourse-common/utils/decorators";
+import { getOwner } from "discourse-common/lib/get-owner";
 
 export default {
   name: "chat-setup",
   initialize(container) {
-    this.currentUser = container.lookup("current-user:main");
     this.chatService = container.lookup("service:chat");
 
     withPluginApi("0.12.1", (api) => {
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -1,28 +1,25 @@
...
-          if (!this.currentUser) {
-            this.currentUser = container.lookup("current-user:main");
-          }
-
-          const currentUserTimezone = this.currentUser?.resolvedTimezone(
-            this.currentUser
-          );
+          const currentUser = getOwner(this).lookup("current-user:main");
+          const currentUserTimezone =
+            currentUser?.resolvedTimezone(currentUser);
```
tgxworld, line 18: Not related to this PR but the API here feels weird. We have to pass the `currentUser` as an argument to the function called in the context of `currentUser`

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -1,28 +1,25 @@
...
       api.decorateCookedElement(
         (elem) => {
-          if (!this.currentUser) {
-            this.currentUser = container.lookup("current-user:main");
-          }
-
-          const currentUserTimezone = this.currentUser?.resolvedTimezone(
-            this.currentUser
-          );
+          const currentUser = getOwner(this).lookup("current-user:main");
```
tgxworld, line 16: Just curious, isn't  `currentUser` a singleton so looking up here vs during initialisation should yield the same object?

assets/javascripts/discourse/initializers/chat-setup.js
CvX: > the API here feels weird

Yeah, that's **exactly** what I said when reviewing another PR 😅 https://github.com/discourse/discourse/pull/15749#pullrequestreview-866816280 I suppose it could be turned into a low-pri todo 🙂

assets/javascripts/discourse/initializers/chat-setup.js
CvX: It does! But… in every test after that the app instance changes, and the function does not. And it keeps the reference to that first user object / container.

davidtaylorhq (approved):

assets/javascripts/discourse/initializers/chat-setup.js
martin-brennan: Ah well I was close I guess :joy: Thanks @CvX , I only didn't do this lookup here for "efficiency" but guess it wasn't far enough anyway.

PR #692, opened 2022-03-15
FIX: Clear unread when channel archived

martin-brennan:
When the channel was archived some users could end
up with an unclearable unread indicator. This commit
fixes the issue but for now the user still needs to
reload for the channel to drop off the list.

### DIFF
--- a/assets/javascripts/discourse/services/chat.js
+++ b/assets/javascripts/discourse/services/chat.js
@@ -1,4 +1,5 @@
 import { popupAjaxError } from "discourse/lib/ajax-error";
+import { CHANNEL_STATUSES } from "discourse/plugins/discourse-chat/discourse/models/chat-channel";
 import Service, { inject as service } from "@ember/service";
 import Site from "discourse/models/site";
 import { addChatToolbarButton } from "discourse/plugins/discourse-chat/discourse/components/chat-composer";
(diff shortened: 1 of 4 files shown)


lib/chat_channel_archive_service.rb
```diff
@@ -225,7 +226,9 @@ def notify_archiver(result, error: nil)
     )
   end
 
-  def kick_all_users
-    UserChatChannelMembership.where(chat_channel: chat_channel).update_all(following: false)
+  def kick_all_users(last_message_id)
```
tgxworld, line 229: Just wonder why we need to pass this as an argument when this method can derive the `last_message_id` from the `chat_channel` object.

spec/lib/chat_channel_archive_service_spec.rb
```diff
@@ -127,16 +127,29 @@ def start_archive
...
 
-      it "unfollows (leaves) the channel for all users" do
-        create_messages(3)
-        channel.chat_messages.map(&:user).each do |user|
-          UserChatChannelMembership.create(chat_channel: channel, user: user, following: true)
+      describe "channel members" do
+        before do
+          create_messages(3)
+          channel.chat_messages.map(&:user).each do |user|
+            UserChatChannelMembership.create(chat_channel: channel, user: user, following: true)
```
tgxworld, line 134:
```suggestion
            UserChatChannelMembership.create!(chat_channel: channel, user: user, following: true)
```

assets/javascripts/discourse/services/chat.js
```diff
@@ -560,6 +561,20 @@ export default Service.extend({
         }
 
         channel.set("status", busData.status);
+
+        // it is not possible for the user to set their last read message id
+        // if the channel has been archived, because all the messages have
+        // been deleted. we don't want them seeing the blue dot anymore so
+        // just completely reset the unreads
+        if (busData.status === CHANNEL_STATUSES.archived) {
```
tgxworld, line 569: Will we be able to test for this?

tgxworld (approved):
Just a couple of comments but the approach is OK

lib/chat_channel_archive_service.rb
martin-brennan: good point will change

assets/javascripts/discourse/services/chat.js
martin-brennan: Hmm I think so, we can send MessageBus messages in JS tests

martin-brennan: Will add one

PR #699, opened 2022-03-16
FEATURE: Include reactions in archive chat quotes

martin-brennan:
This commit introduces a new reactions attribute to
the [chat] BBcode in the format `emoji:user1,user2;emoji2:user1`.
When included, we store this data in a data-reactions attr
on the wrapping chat transcript div, and also construct
reaction indicators at the bottom of each [chat] quoted
set of messages.

We have intentionally left off the following:

* Fidelity is only per-group of messages. For example, if there
  are 3 messages from a single user grouped into one [chat]
  block, and each individual message has a reaction, all 3
  reactions are placed at the bottom of the set of messages.
* When hovering the reactions, we do not show which username
  reacted to what, we should be easily able to do this in
  V2 of this feature.

Regular quoting of chat messages will not include
reaction metadata, only when the messages are archived.

image

### DIFF
--- a/assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
+++ b/assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
@@ -1,4 +1,5 @@
 import I18n from "I18n";
+import { performEmojiUnescape } from "pretty-text/emoji";
 
 let customMarkdownCookFn;
 
@@ -14,6 +15,7 @@ const chatTranscriptRule = {
     const options = state.md.options.discourse;
     const [username, messageIdStart, messageTimeStart] =
       (tagInfo.attrs.quote && tagInfo.attrs.quote.split(";")) || [];
+    const reactions = tagInfo.attrs.reactions;
     const multiQuote = !!tagInfo.attrs.multiQuote;
     const noLink = !!tagInfo.attrs.noLink;
     const channelName = tagInfo.attrs.channel;
@@ -38,6 +40,10 @@ const chatTranscriptRule = {
     wrapperDivToken.attrs.push(["data-message-id", messageIdStart]);
     wrapperDivToken.attrs.push(["data-username", username]);
     wrapperDivToken.attrs.push(["data-datetime", messageTimeStart]);
+
+    if (reactions) {
+      wrapperDivToken.attrs.push(["data-reactions", reactions]);
+    }
 
     if (channelName) {
       wrapperDivToken.attrs.push(["data-channel-name", channelName]);
@@ -147,6 +153,41 @@ const chatTranscriptRule = {
     token.content = customMarkdownCookFn(content);
     state.push("html_raw", "", -1);
 
+    if (reactions) {
+      let emojiHtmlCache = {};
+      let reactionsToken = state.push(
+        "div_chat_transcript_reactions",
+        "div",
+        1
+      );
+      reactionsToken.attrs = [["class", "chat-transcript-reactions"]];
+
+      const reactionsMap = reactions.split(";");
+      reactionsMap.forEach((reaction) => {
+        const split = reaction.split(":");
+        const emoji = split[0];
+        const usernames = split[1].split(",");
+
+        const reactToken = state.push("div_chat_transcript_reaction", "div", 1);
+        reactToken.attrs = [["class", "chat-transcript-reaction"]];
+        const emojiToken = state.push("html_inline", "", 0);
+        if (!emojiHtmlCache[emoji]) {
+          emojiHtmlCache[emoji] = performEmojiUnescape(`:${emoji}:`, {
+            getURL: options.getURL,
+            emojiSet: options.emojiSet,
+            emojiCDNUrl: options.emojiCDNUrl,
+            enableEmojiShortcuts: options.enableEmojiShortcuts,
+            inlineEmoji: options.inlineEmoji,
+            lazy: true,
+          });
+        }
+        emojiToken.content =
+          emojiHtmlCache[emoji] + " " + usernames.length.toString();
+        state.push("div_chat_transcript_reaction", "div", -1);
+      });
+      state.push("div_chat_transcript_reactions", "div", -1);
+    }
+
     state.push("div_chat_transcript_messages", "div", -1);
     state.push("div_chat_transcript_wrap", "div", -1);
     return true;
@@ -163,6 +204,8 @@ export function setup(helper) {
     "div.chat-transcript-user-avatar",
     "div.chat-transcript-messages",
     "div.chat-transcript-datetime",
+    "div.chat-transcript-reactions",
+    "div.chat-transcript-reaction",
     "span[title]",
     "div[data-message-id]",
     "div[data-channel-name]",
--- a/assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
+++ b/assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
@@ -162,8 +162,7 @@ const chatTranscriptRule = {
       );
       reactionsToken.attrs = [["class", "chat-transcript-reactions"]];
 
-      const reactionsMap = reactions.split(";");
-      reactionsMap.forEach((reaction) => {
+      reactions.split(";").forEach((reaction) => {
         const split = reaction.split(":");
         const emoji = split[0];
         const usernames = split[1].split(",");
@@ -181,8 +180,9 @@ const chatTranscriptRule = {
             lazy: true,
           });
         }
-        emojiToken.content =
-          emojiHtmlCache[emoji] + " " + usernames.length.toString();
+        emojiToken.content = `${
+          emojiHtmlCache[emoji]
+        } ${usernames.length.toString()}`;
         state.push("div_chat_transcript_reaction", "div", -1);
       });
       state.push("div_chat_transcript_reactions", "div", -1);
--- a/lib/chat_transcript_service.rb
+++ b/lib/chat_transcript_service.rb
@@ -60,12 +60,14 @@ class ChatTranscriptService
 
     def reactions_attr
       reaction_data = @message_data.reduce([]) do |array, msg_data|
-        next if msg_data[:reactions].empty?
-        array << msg_data[:reactions].map do |react|
-          "#{react.emoji}:#{react.usernames}"
+        if msg_data[:reactions].any?
+          array << msg_data[:reactions].map do |react|
+            "#{react.emoji}:#{react.usernames}"
+          end
         end
+        array
       end
-      return if reaction_data.blank? || reaction_data.empty?
+      return if reaction_data.empty?
       "reactions=\"#{reaction_data.join(";")}\""
     end
 
@@ -128,7 +130,9 @@ class ChatTranscriptService
   private
 
   def messages
-    @messages ||= ChatMessage.includes(:user, :uploads, :reactions).where(
+    @messages ||= ChatMessage.includes(
+      :user, chat_uploads: :upload
+    ).where(
       id: @message_ids, chat_channel_id: @channel.id
     ).order(:created_at)
   end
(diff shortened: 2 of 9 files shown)


lib/chat_transcript_service.rb
```diff
@@ -108,8 +128,31 @@ def generate_markdown
...
+  ##
+  # Queries reactions and returns them in this format
+  #
+  # emoji   |  usernames  |  chat_message_id
+  # ----------------------------------------
+  # +1      | foo,bar,baz | 102
+  # heart   | foo         | 102
+  # sob     | bar,baz     | 103
+  def reactions
+    @reactions ||= DB.query(<<~SQL, @messages.map(&:id))
```
tgxworld, line 145: This isn't immediately clear to me but potentially how many messages can the `@messages` array be?

martin-brennan: There is no limit, _however_ the current highest amount of messages is 100 at a time from the archiving service:

https://github.com/discourse/discourse-chat/blob/ee9724e8af18dfcf6cdd1da0f1ff6cdaac5b228f/lib/chat_channel_archive_service.rb#L61-L81

Theoretically I guess someone could click tons of messages in the UI to create a quote for. Maybe I could introduce a limit on the number of messages that the transcript service will accept?

martin-brennan: `ARCHIVED_MESSAGES_PER_POST` is 100

lib/chat_transcript_service.rb
```diff
@@ -20,41 +20,55 @@ class ChatTranscriptService
...
 
     private
 
+    def reactions_attr
+      reaction_data = @message_data.map do |msg_data|
+        next if msg_data[:reactions].empty?
+        msg_data[:reactions].map do |react|
+          "#{react.emoji}:#{react.usernames}"
+        end
+      end.compact
```
tgxworld, line 67:
```suggestion
      reaction_data = @message_data.reduce([]) do |array, msg_data|
        next if msg_data[:reactions].empty?

        array << msg_data[:reactions].map do |react|
          "#{react.emoji}:#{react.usernames}"
        end
      end
```

Using `Array#reduce` can simplify the code somewhat here.

lib/chat_transcript_service.rb
tgxworld: ahh OK so there is a upper limit. I just wanted to make sure of that. Btw, can you check if this query is using an index? I had a look at the model and aside from a unique index which I do not think this query will use, we do not have any other indexes on the foreign keys.

test/javascripts/acceptance/chat-transcript-test.js
```diff
@@ -242,6 +264,27 @@ acceptance("Discourse Chat | discourse-chat-transcript", function (needs) {
     );
   });
 
+  test("renders with the reactions attribute", function (assert) {
```
tgxworld, line 267: I'm not sure if this is considered an acceptance test when it does not test any interaction behaviour of the app.

assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
```diff
@@ -147,6 +153,41 @@ const chatTranscriptRule = {
...
+      let emojiHtmlCache = {};
+      let reactionsToken = state.push(
+        "div_chat_transcript_reactions",
+        "div",
+        1
+      );
+      reactionsToken.attrs = [["class", "chat-transcript-reactions"]];
+
+      const reactionsMap = reactions.split(";");
+      reactionsMap.forEach((reaction) => {
```
tgxworld, line 166:
```suggestion
      reactions.split(";").forEach((reaction) => {
```

assets/javascripts/lib/discourse-markdown/discourse-chat-transcript.js
```diff
@@ -147,6 +153,41 @@ const chatTranscriptRule = {
...
+            getURL: options.getURL,
+            emojiSet: options.emojiSet,
+            emojiCDNUrl: options.emojiCDNUrl,
+            enableEmojiShortcuts: options.enableEmojiShortcuts,
+            inlineEmoji: options.inlineEmoji,
+            lazy: true,
+          });
+        }
+        emojiToken.content =
+          emojiHtmlCache[emoji] + " " + usernames.length.toString();
```
tgxworld, line 185:
```suggestion
          `${emojiHtmlCache[emoji]} {usernames.length.toString()}`;
```

tgxworld:
> Screenshot from 2022-03-16 14-27-09

Just curious, will the reactions just flow to the next line if a message has a ton of different reactions?

martin-brennan:
> Just curious, will the reactions just flow to the next line if a message has a ton of different reactions?

Not sure, will give it a try, I don't even know what happens within chat if there are tons of different reactions

lib/chat_transcript_service.rb
martin-brennan: Nice!

test/javascripts/acceptance/chat-transcript-test.js
martin-brennan: I guess...I really don't want to have to move all this around though lol. But if it's really necessary I can, I guess everything under `acceptance("Discourse Chat | discourse-chat-transcript", function (needs) {` would be an integration test?

tgxworld: Hmm maybe something for the future then since this is already the status quo.

martin-brennan:
> Not sure, will give it a try, I don't even know what happens within chat if there are tons of different reactions

As per internal chat, this is fine but we need some limits.

lib/chat_transcript_service.rb
martin-brennan: Yeah I will take a look, probably wouldn't hurt to add an index to the `chat_message_id`

lib/chat_transcript_service.rb
```diff
@@ -20,41 +20,55 @@ class ChatTranscriptService
...
     private
 
+    def reactions_attr
+      reaction_data = @message_data.reduce([]) do |array, msg_data|
+        next if msg_data[:reactions].empty?
+        array << msg_data[:reactions].map do |react|
+          "#{react.emoji}:#{react.usernames}"
+        end
+      end
+      return if reaction_data.blank? || reaction_data.empty?
```
tgxworld, line 68: I believe `reaction_data` will always be an array, is there a reason we need to try with `blank?` and `empty?` here?

martin-brennan: Yes, it's not always an array @tgxworld without this a bunch of the specs failed (if there are no reactions)

tgxworld: Hmmm blank? should cover both nil and an empty array though

lib/chat_transcript_service.rb
martin-brennan: Even after adding an index to `chat_message_id` on that table it is not used for this query:

```
                                            Table "public.chat_message_reactions"
     Column      |              Type              | Collation | Nullable |                      Default
-----------------+--------------------------------+-----------+----------+----------------------------------------------------
 id              | bigint                         |           | not null | nextval('chat_message_reactions_id_seq'::regclass)
 chat_message_id | integer                        |           |          |
 user_id         | integer                        |           |          |
 emoji           | character varying              |           |          |
 created_at      | timestamp(6) without time zone |           | not null |
 updated_at      | timestamp(6) without time zone |           | not null |
Indexes:
    "chat_message_reactions_pkey" PRIMARY KEY, btree (id)
    "chat_message_reactions_index" UNIQUE, btree (chat_message_id, user_id, emoji)
    "index_chat_message_reactions_on_chat_message_id" btree (chat_message_id)

discourse_development=# EXPLAIN ANALYZE SELECT emoji, STRING_AGG(DISTINCT users.username, ',') AS usernames, chat_message_id
discourse_development-# FROM chat_message_reactions
discourse_development-# INNER JOIN users on users.id = chat_message_reactions.user_id
discourse_development-# WHERE chat_message_id IN (8,9,10,11,12,13,14)
discourse_development-# GROUP BY emoji, chat_message_id
discourse_development-# ORDER BY chat_message_id, emoji
discourse_development-# ;
                                                                QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=153.01..192.04 rows=1735 width=47) (actual time=2.532..4.248 rows=1735 loops=1)
   Group Key: chat_message_reactions.chat_message_id, chat_message_reactions.emoji
   ->  Sort  (cost=153.01..157.34 rows=1735 width=30) (actual time=2.511..2.611 rows=1735 loops=1)
         Sort Key: chat_message_reactions.chat_message_id, chat_message_reactions.emoji
         Sort Method: quicksort  Memory: 181kB
         ->  Hash Join  (cost=6.21..59.66 rows=1735 width=30) (actual time=0.142..0.668 rows=1735 loops=1)
               Hash Cond: (chat_message_reactions.user_id = users.id)
               ->  Seq Scan on chat_message_reactions  (cost=0.00..48.53 rows=1735 width=19) (actual time=0.020..0.298 rows=1735 loops=1)
                     Filter: (chat_message_id = ANY ('{8,9,10,11,12,13,14}'::integer[]))
               ->  Hash  (cost=5.54..5.54 rows=54 width=19) (actual time=0.092..0.092 rows=54 loops=1)
                     Buckets: 1024  Batches: 1  Memory Usage: 11kB
                     ->  Seq Scan on users  (cost=0.00..5.54 rows=54 width=19) (actual time=0.012..0.050 rows=54 loops=1)
 Planning time: 2.488 ms
 Execution time: 4.540 ms
(14 rows)
```

However it doesn't seem to be too heavy of a query, that's 4ms which includes the chat message I made yesterday with 1700 reactions. Skipping adding the index for now as it seems to make no difference, here is the same query with no index:

```
                                                                QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=153.01..192.04 rows=1735 width=47) (actual time=5.274..6.972 rows=1735 loops=1)
   Group Key: chat_message_reactions.chat_message_id, chat_message_reactions.emoji
   ->  Sort  (cost=153.01..157.34 rows=1735 width=30) (actual time=5.264..5.318 rows=1735 loops=1)
         Sort Key: chat_message_reactions.chat_message_id, chat_message_reactions.emoji
         Sort Method: quicksort  Memory: 181kB
         ->  Hash Join  (cost=6.21..59.66 rows=1735 width=30) (actual time=0.153..2.303 rows=1735 loops=1)
               Hash Cond: (chat_message_reactions.user_id = users.id)
               ->  Seq Scan on chat_message_reactions  (cost=0.00..48.53 rows=1735 width=19) (actual time=0.018..0.953 rows=1735 loops=1)
                     Filter: (chat_message_id = ANY ('{8,9,10,11,12,13,14}'::integer[]))
               ->  Hash  (cost=5.54..5.54 rows=54 width=19) (actual time=0.082..0.082 rows=54 loops=1)
                     Buckets: 1024  Batches: 1  Memory Usage: 11kB
                     ->  Seq Scan on users  (cost=0.00..5.54 rows=54 width=19) (actual time=0.008..0.043 rows=54 loops=1)
 Planning time: 0.571 ms
 Execution time: 7.092 ms
(14 rows)
```

lib/chat_transcript_service.rb
martin-brennan: I realised I was using the reduce wrong, about to push a fix

martin-brennan: Done

lib/chat_transcript_service.rb
tgxworld: Ooo I just a query on dev and realized the `chat_message_reactions_index` will actually be used in cases where the filter dataset is large since `chat_message_id` is the first column in the unique index. All good here :)

tgxworld (approved):

PR #700, opened 2022-03-16
FIX: ensures tc-reply-display is taking using necessary space

jjaffeux:


jjaffeux:
Avoid this case:

image

CvX (approved):

PR #701, opened 2022-03-16
FIX: moves unread indicator to chat-channel-title

jjaffeux:
This way it's included in the grid of chat-channel-title and ensures we have enough space.


jjaffeux:
This would prevent a bug with long chat channel titles messing with the indicator size:

Screenshot 2022-03-16 at 11 02 02

ZogStriP (approved):

PR #704, opened 2022-03-16
FIX: Don't error when user no longer exists

markvanlan:
1. If a user no longer exists in the DB, do not completely error all of chat.
2. Make sure that we don't update `last_read_message_id` to a message that has already been read

Screen Shot 2022-03-16 at 11 34 45 AM


tgxworld:
Is there anyway we can test this?

markvanlan:
@tgxworld Yeah, I can do an acceptance test.

On storing the user data - I think that would be a GDPR issue

markvanlan:
The setup for the test here is 50x the length of the test.. but it's nice to have setup for future component testing on `chat-message`

CvX (approved):

PR #705, opened 2022-03-16
FIX: Only return open channels for quick selector

markvanlan:


lib/chat_channel_fetcher.rb
```diff
@@ -39,9 +39,11 @@ def self.preload_custom_fields_for(channels)
...
 
   def self.public_channels_with_filter(guardian, memberships, filter)
-    channels = ChatChannel.includes(:chatable)
-    channels = channels.where(chatable_type: ChatChannel.public_channel_chatable_types)
-    channels = channels.where("LOWER(name) LIKE ?", "#{filter}%")
+    channels = ChatChannel
+      .includes(:chatable)
+      .where(chatable_type: ChatChannel.public_channel_chatable_types)
+      .where(status: ChatChannel.statuses[:open])
+      .where("LOWER(name) LIKE ?", "#{filter}%")
```
tgxworld, line 46:
```suggestion
    channels = ChatChannel
      .includes(:chatable)
      .where(
        chatable_type: ChatChannel.public_channel_chatable_types,
        status: ChatChannel.statuses[:open]
      )
      .where("LOWER(name) LIKE ?", "#{filter}%")
```

Minor suggestion which maintains readability but removes one extra method call.

tgxworld (approved):

PR #711, opened 2022-03-17
FEATURE: Deleting chat channels

martin-brennan:
This commit introduces the delete functionality for chat
channels. This will trash the chat channel and its messages,
and permanently delete all related records like chat uploads,
chat reactions, and so on.

image

### DIFF
--- a/app/controllers/chat_channels_controller.rb
+++ b/app/controllers/chat_channels_controller.rb
@@ -262,7 +262,7 @@ class DiscourseChat::ChatChannelsController < DiscourseChat::ChatBaseController
     render json: success_json
   end
 
-  def delete
+  def destroy
     params.require(:chat_channel_id)
     params.require(:channel_name_confirmation)
 
@@ -272,17 +272,24 @@ class DiscourseChat::ChatChannelsController < DiscourseChat::ChatBaseController
     guardian.ensure_can_delete_chat_channel!
 
     if chat_channel.name.downcase != params[:channel_name_confirmation].downcase
-      raise Discourse::InvalidParameters
+      raise Discourse::InvalidParameters.new(:channel_name_confirmation)
     end
 
-    chat_channel.trash!(current_user)
-    StaffActionLogger.new(current_user).log_custom(
-      "chat_channel_delete",
-      {
-        chat_channel_id: chat_channel.id,
-        chat_channel_name: chat_channel.name
-      }
-    )
+    begin
+      ChatChannel.transaction do
+        chat_channel.trash!(current_user)
+        StaffActionLogger.new(current_user).log_custom(
+          "chat_channel_delete",
+          {
+            chat_channel_id: chat_channel.id,
+            chat_channel_name: chat_channel.name
+          }
+        )
+      end
+    rescue ActiveRecord::Rollback
+      return render_json_error(I18n.t("chat.errors.delete_channel_failed"))
+    end
+
     Jobs.enqueue(:chat_channel_delete, { chat_channel_id: chat_channel.id })
     render json: success_json
   end
--- a/app/controllers/chat_channels_controller.rb
+++ b/app/controllers/chat_channels_controller.rb
@@ -262,6 +262,31 @@ class DiscourseChat::ChatChannelsController < DiscourseChat::ChatBaseController
     render json: success_json
   end
 
+  def delete
+    params.require(:chat_channel_id)
+    params.require(:channel_name_confirmation)
+
+    chat_channel = ChatChannel.find_by(id: params[:chat_channel_id])
+    raise Discourse::NotFound if chat_channel.blank?
+
+    guardian.ensure_can_delete_chat_channel!
+
+    if chat_channel.name.downcase != params[:channel_name_confirmation].downcase
+      raise Discourse::InvalidParameters
+    end
+
+    chat_channel.trash!(current_user)
+    StaffActionLogger.new(current_user).log_custom(
+      "chat_channel_delete",
+      {
+        chat_channel_id: chat_channel.id,
+        chat_channel_name: chat_channel.name
+      }
+    )
+    Jobs.enqueue(:chat_channel_delete, { chat_channel_id: chat_channel.id })
+    render json: success_json
+  end
+
   private
 
   def render_channel_for_chatable(channel)
--- a/app/jobs/regular/chat_channel_delete.rb
+++ b/app/jobs/regular/chat_channel_delete.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+module Jobs
+  class ChatChannelDelete < ::Jobs::Base
+    def execute(args = {})
+      chat_channel = ::ChatChannel.with_deleted.find_by(id: args[:chat_channel_id])
+
+      # this should not really happen, but better to do this than throw an error
+      if chat_channel.blank?
+        Rails.logger.warn("Chat channel #{args[:chat_channel_id]} could not be found, aborting delete job.")
+        return
+      end
+
+      DistributedMutex.synchronize(
+        "delete_chat_channel_#{chat_channel.id}",
+        validity: 20.minutes
+      ) do
+        Rails.logger.debug("Deleting webhooks and events for channel #{chat_channel.id}")
+        ChatMessage.transaction do
+          webhooks = IncomingChatWebhook.where(chat_channel: chat_channel)
+          ChatWebhookEvent.where(incoming_chat_webhook_id: webhooks.pluck(:id)).delete_all
+          webhooks.delete_all
+        end
+
+        Rails.logger.debug("Deleting drafts and memberships for channel #{chat_channel.id}")
+        ChatDraft.where(chat_channel: chat_channel).delete_all
+        UserChatChannelMembership.where(chat_channel: chat_channel).delete_all
+
+        Rails.logger.debug("Deleting chat messages, mentions, revisions, and uploads for channel #{chat_channel.id}")
+        ChatMessage.transaction do
+          chat_messages = ChatMessage.where(chat_channel: chat_channel)
+          ChatMention.where(chat_message_id: chat_messages.pluck(:id)).delete_all
+          ChatMessageRevision.where(chat_message_id: chat_messages.pluck(:id)).delete_all
+
+          # if the uploads are not used anywhere else they will be deleted
+          # by the CleanUpUploads job in core
+          ChatUpload.where(chat_message_id: chat_messages.pluck(:id)).delete_all
+
+          # only the messages and the channel are Trashable, everything else gets
+          # permanently destroyed
+          chat_messages.update_all(
+            deleted_by_id: chat_channel.deleted_by_id, deleted_at: Time.zone.now
+          )
+        end
+      end
+    end
+  end
+end
--- a/app/jobs/regular/chat_channel_delete.rb
+++ b/app/jobs/regular/chat_channel_delete.rb
@@ -29,12 +29,14 @@ module Jobs
         Rails.logger.debug("Deleting chat messages, mentions, revisions, and uploads for channel #{chat_channel.id}")
         ChatMessage.transaction do
           chat_messages = ChatMessage.where(chat_channel: chat_channel)
-          ChatMention.where(chat_message_id: chat_messages.select(:id)).delete_all
-          ChatMessageRevision.where(chat_message_id: chat_messages.select(:id)).delete_all
+          message_ids = chat_messages.select(:id)
+          ChatMention.where(chat_message_id: message_ids).delete_all
+          ChatMessageRevision.where(chat_message_id: message_ids).delete_all
+          ChatMessageReaction.where(chat_message_id: message_ids).delete_all
 
           # if the uploads are not used anywhere else they will be deleted
           # by the CleanUpUploads job in core
-          ChatUpload.where(chat_message_id: chat_messages.select(:id)).delete_all
+          ChatUpload.where(chat_message_id: message_ids).delete_all
 
           # only the messages and the channel are Trashable, everything else gets
           # permanently destroyed
--- a/app/jobs/regular/chat_channel_delete.rb
+++ b/app/jobs/regular/chat_channel_delete.rb
@@ -13,7 +13,7 @@ module Jobs
 
       DistributedMutex.synchronize(
         "delete_chat_channel_#{chat_channel.id}",
-        validity: 20.minutes
+        validity: 1.minute
       ) do
         Rails.logger.debug("Deleting webhooks and events for channel #{chat_channel.id}")
         ChatMessage.transaction do
--- a/app/jobs/regular/chat_channel_delete.rb
+++ b/app/jobs/regular/chat_channel_delete.rb
@@ -18,7 +18,7 @@ module Jobs
         Rails.logger.debug("Deleting webhooks and events for channel #{chat_channel.id}")
         ChatMessage.transaction do
           webhooks = IncomingChatWebhook.where(chat_channel: chat_channel)
-          ChatWebhookEvent.where(incoming_chat_webhook_id: webhooks.pluck(:id)).delete_all
+          ChatWebhookEvent.where(incoming_chat_webhook_id: webhooks.select(:id)).delete_all
           webhooks.delete_all
         end
 
@@ -29,12 +29,12 @@ module Jobs
         Rails.logger.debug("Deleting chat messages, mentions, revisions, and uploads for channel #{chat_channel.id}")
         ChatMessage.transaction do
           chat_messages = ChatMessage.where(chat_channel: chat_channel)
-          ChatMention.where(chat_message_id: chat_messages.pluck(:id)).delete_all
-          ChatMessageRevision.where(chat_message_id: chat_messages.pluck(:id)).delete_all
+          ChatMention.where(chat_message_id: chat_messages.select(:id)).delete_all
+          ChatMessageRevision.where(chat_message_id: chat_messages.select(:id)).delete_all
 
           # if the uploads are not used anywhere else they will be deleted
           # by the CleanUpUploads job in core
-          ChatUpload.where(chat_message_id: chat_messages.pluck(:id)).delete_all
+          ChatUpload.where(chat_message_id: chat_messages.select(:id)).delete_all
 
           # only the messages and the channel are Trashable, everything else gets
           # permanently destroyed
--- a/app/jobs/regular/chat_channel_delete.rb
+++ b/app/jobs/regular/chat_channel_delete.rb
@@ -11,10 +11,7 @@ module Jobs
         return
       end
 
-      DistributedMutex.synchronize(
-        "delete_chat_channel_#{chat_channel.id}",
-        validity: 1.minute
-      ) do
+      DistributedMutex.synchronize("delete_chat_channel_#{chat_channel.id}") do
         Rails.logger.debug("Deleting webhooks and events for channel #{chat_channel.id}")
         ChatMessage.transaction do
           webhooks = IncomingChatWebhook.where(chat_channel: chat_channel)
--- a/assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
+++ b/assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
@@ -43,9 +43,10 @@ export default Component.extend({
           text: I18n.t("chat.channel_delete.process_started"),
           messageClass: "success",
         });
+        this.appEvents.trigger("chat-channel:deleted", this.chatChannel);
         later(() => {
           if (!isTesting()) {
-            window.location.reload();
+            this.closeModal();
           }
         }, 3000);
       })
--- a/assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
+++ b/assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
@@ -14,10 +14,12 @@ export default Component.extend({
   tagName: "",
   chatChannel: null,
   channelNameConfirmation: null,
+  deleting: false,
+  confirmed: false,
 
-  @discourseComputed("deleting", "channelNameConfirmation")
-  buttonDisabled(deleting, channelNameConfirmation) {
-    if (deleting) {
+  @discourseComputed("deleting", "channelNameConfirmation", "confirmed")
+  buttonDisabled(deleting, channelNameConfirmation, confirmed) {
+    if (deleting || confirmed) {
       return true;
     }
 
@@ -39,6 +41,7 @@ export default Component.extend({
       data: { channel_name_confirmation: this.channelNameConfirmation },
     })
       .then(() => {
+        this.set("confirmed", true);
         this.appEvents.trigger("modal-body:flash", {
           text: I18n.t("chat.channel_delete.process_started"),
           messageClass: "success",
--- a/assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
+++ b/assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
@@ -1,0 +1,58 @@
+import Component from "@ember/component";
+import { isTesting } from "discourse-common/config/environment";
+import { later } from "@ember/runloop";
+import { isEmpty } from "@ember/utils";
+import { CHANNEL_STATUSES } from "discourse/plugins/discourse-chat/discourse/models/chat-channel";
+import I18n from "I18n";
+import discourseComputed from "discourse-common/utils/decorators";
+import { action } from "@ember/object";
+import { ajax } from "discourse/lib/ajax";
+import { inject as service } from "@ember/service";
+import { popupAjaxError } from "discourse/lib/ajax-error";
+
+export default Component.extend({
+  chat: service(),
+  tagName: "",
+  chatChannel: null,
+  channelNameConfirmation: null,
+
+  @discourseComputed("deleting", "channelNameConfirmation")
+  buttonDisabled(deleting, channelNameConfirmation) {
+    if (deleting) {
+      return true;
+    }
+
+    if (
+      isEmpty(channelNameConfirmation) ||
+      channelNameConfirmation.toLowerCase() !== this.chatChannel.title.toLowerCase()
+    ) {
+      return true;
+    }
+    return false;
+  },
+
+  @action
+  deleteChannel() {
+    this.set("deleting", true);
+    return ajax(
+      `/chat/chat_channels/${this.chatChannel.id}.json`,
+      {
+        method: "DELETE",
+        data: { channel_name_confirmation: this.channelNameConfirmation },
+      }
+    )
+      .then(() => {
+        this.appEvents.trigger("modal-body:flash", {
+          text: I18n.t("chat.channel_delete.process_started"),
+          messageClass: "success",
+        });
+        later(() => {
+          if (!isTesting()) {
+            window.location.reload();
+          }
+        }, 3000);
+      })
+      .catch(popupAjaxError)
+      .finally(() => this.set("deleting", false));
+  },
+});
--- a/assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
+++ b/assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
@@ -2,7 +2,6 @@ import Component from "@ember/component";
 import { isTesting } from "discourse-common/config/environment";
 import { later } from "@ember/runloop";
 import { isEmpty } from "@ember/utils";
-import { CHANNEL_STATUSES } from "discourse/plugins/discourse-chat/discourse/models/chat-channel";
 import I18n from "I18n";
 import discourseComputed from "discourse-common/utils/decorators";
 import { action } from "@ember/object";
@@ -24,7 +23,8 @@ export default Component.extend({
 
     if (
       isEmpty(channelNameConfirmation) ||
-      channelNameConfirmation.toLowerCase() !== this.chatChannel.title.toLowerCase()
+      channelNameConfirmation.toLowerCase() !==
+        this.chatChannel.title.toLowerCase()
     ) {
       return true;
     }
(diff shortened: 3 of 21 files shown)


app/controllers/chat_channels_controller.rb
```diff
@@ -262,6 +262,31 @@ def change_status
     render json: success_json
   end
 
+  def delete
```
tgxworld, line 265:
```suggestion
  def destroy
```

We mainly use `destroy` in our codebase. Perhaps worth standardising here?

app/controllers/chat_channels_controller.rb
```diff
@@ -262,6 +262,31 @@ def change_status
...
+    params.require(:chat_channel_id)
+    params.require(:channel_name_confirmation)
+
+    chat_channel = ChatChannel.find_by(id: params[:chat_channel_id])
+    raise Discourse::NotFound if chat_channel.blank?
+
+    guardian.ensure_can_delete_chat_channel!
+
+    if chat_channel.name.downcase != params[:channel_name_confirmation].downcase
+      raise Discourse::InvalidParameters
```
tgxworld, line 275:
```suggestion
      raise Discourse::InvalidParameters.new(:channel_name_confirmation)
```

app/controllers/chat_channels_controller.rb
```diff
@@ -262,6 +262,31 @@ def change_status
...
+    end
+
+    chat_channel.trash!(current_user)
+    StaffActionLogger.new(current_user).log_custom(
+      "chat_channel_delete",
+      {
+        chat_channel_id: chat_channel.id,
+        chat_channel_name: chat_channel.name
+      }
+    )
```
tgxworld, line 285: Perhaps it is safer for this to be in the same DB transaction so we'll never end up with a case where the channel is deleted but no staff action log has been created.

martin-brennan: Yep good call!

app/controllers/chat_channels_controller.rb
martin-brennan: Ah yeah my bad, will change

martin-brennan:
@tgxworld made those changes now

app/jobs/regular/chat_channel_delete.rb
```diff
@@ -0,0 +1,48 @@
...
+      if chat_channel.blank?
+        Rails.logger.warn("Chat channel #{args[:chat_channel_id]} could not be found, aborting delete job.")
+        return
+      end
+
+      DistributedMutex.synchronize(
+        "delete_chat_channel_#{chat_channel.id}",
+        validity: 20.minutes
+      ) do
+        Rails.logger.debug("Deleting webhooks and events for channel #{chat_channel.id}")
```
tgxworld, line 18:
```suggestion
        Rails.logger.debug { "Deleting webhooks and events for channel #{chat_channel.id}" }
```

Using a block here will save a string allocation because the block is only called when the log level is met.

app/jobs/regular/chat_channel_delete.rb
```diff
@@ -0,0 +1,48 @@
...
+        ChatMessage.transaction do
+          webhooks = IncomingChatWebhook.where(chat_channel: chat_channel)
+          ChatWebhookEvent.where(incoming_chat_webhook_id: webhooks.pluck(:id)).delete_all
```
tgxworld, line 21:
```suggestion
          ChatWebhookEvent.where(incoming_chat_webhook_id: webhooks.select(:id)).delete_all
```

This will allow us to run a single SQL query instead of 2.

app/jobs/regular/chat_channel_delete.rb
```diff
@@ -0,0 +1,48 @@
...
+        end
+
+        Rails.logger.debug("Deleting drafts and memberships for channel #{chat_channel.id}")
+        ChatDraft.where(chat_channel: chat_channel).delete_all
+        UserChatChannelMembership.where(chat_channel: chat_channel).delete_all
+
+        Rails.logger.debug("Deleting chat messages, mentions, revisions, and uploads for channel #{chat_channel.id}")
+        ChatMessage.transaction do
+          chat_messages = ChatMessage.where(chat_channel: chat_channel)
+          ChatMention.where(chat_message_id: chat_messages.pluck(:id)).delete_all
```
tgxworld, line 32:
```suggestion
          ChatMention.where(chat_message_id: chat_messages.select(:id)).delete_all
```

app/jobs/regular/chat_channel_delete.rb
```diff
@@ -0,0 +1,48 @@
...
+          ChatMessageRevision.where(chat_message_id: chat_messages.pluck(:id)).delete_all
```
tgxworld, line 33:
```suggestion
          ChatMessageRevision.where(chat_message_id: chat_messages.select(:id)).delete_all
```

app/jobs/regular/chat_channel_delete.rb
martin-brennan: I feel like this is kind of a micro-optimization? We don't do that at all in core

app/jobs/regular/chat_channel_delete.rb
```diff
@@ -0,0 +1,48 @@
...
+
+          # if the uploads are not used anywhere else they will be deleted
+          # by the CleanUpUploads job in core
+          ChatUpload.where(chat_message_id: chat_messages.pluck(:id)).delete_all
```
tgxworld:
```suggestion
          ChatUpload.where(chat_message_id: chat_messages.select(:id)).delete_all
```

app/jobs/regular/chat_channel_delete.rb
```diff
@@ -0,0 +1,48 @@
...
+          webhooks.delete_all
+        end
+
+        Rails.logger.debug("Deleting drafts and memberships for channel #{chat_channel.id}")
+        ChatDraft.where(chat_channel: chat_channel).delete_all
+        UserChatChannelMembership.where(chat_channel: chat_channel).delete_all
+
+        Rails.logger.debug("Deleting chat messages, mentions, revisions, and uploads for channel #{chat_channel.id}")
+        ChatMessage.transaction do
+          chat_messages = ChatMessage.where(chat_channel: chat_channel)
```
tgxworld, line 31: A minor performance concern here, potentially how many chat messages could we be dealing with? Perhaps it might make sense at some point to break this into smaller jobs that can execute in parallel?

martin-brennan: Hmm yeah I did think about that briefly, could be thousands I suppose :/ I could just do find_in_batch like I do for archiving:

```
      chat_channel.chat_messages.find_in_batches(
        batch_size: ARCHIVED_MESSAGES_PER_POST
      ) do |chat_messages|
```

And delete batches of them at a time

app/jobs/regular/chat_channel_delete.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      # this should not really happen, but better to do this than throw an error
+      if chat_channel.blank?
+        Rails.logger.warn("Chat channel #{args[:chat_channel_id]} could not be found, aborting delete job.")
+        return
+      end
+
+      DistributedMutex.synchronize(
+        "delete_chat_channel_#{chat_channel.id}",
+        validity: 20.minutes
```
tgxworld, line 16: I have some concerns around this duration because we saying this job can potentially lock up a Sidekiq worker for a long period of time. Therefore, if I am smart what I can do is to create many chat channels with a ton of chat messages and delete them all at once to lock up all the Sidekiq workers. On a multi-site cluster, this might affect availability. 

A safer design might be to enforce some limitations on the queries that we're running below and queue a new job if we know there are more resources we need to clean up.

martin-brennan: Hmm this is probably just me not understanding how `DistributedMutex` works correctly. Do we even need it at all? I wouldn't think that the deleting of all the records would take long at all, especially since we don't use foreign keys which usually slow down the delete queries.

I would prefer not to have to introduce more complexity by splitting this into multiple jobs or having to do extra cleanups, we should be able to delete everything in one go. I will time deleting a 10k message channel and report back.

assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
```diff
@@ -0,0 +1,55 @@
...
+        this.appEvents.trigger("modal-body:flash", {
+          text: I18n.t("chat.channel_delete.process_started"),
+          messageClass: "success",
+        });
+        later(() => {
+          if (!isTesting()) {
+            window.location.reload();
+          }
+        }, 3000);
+      })
```
tgxworld, line 51: I didn't pull down the changes locally but I think having the page suddenly reload might be surprising especially if I've navigated away from the page. Is the reload to used to clear the channel object from chat?

assets/javascripts/discourse/templates/components/chat-channel-delete-modal-inner.hbs
```diff
@@ -0,0 +1,28 @@
...
+    placeholderKey="chat.channel_delete.confirm_channel_name"
+    autocorrect="off"
+    autocapitalize="off"
+  }}
+{{/d-modal-body}}
+
+<div class="modal-footer">
+  {{d-button
+    disabled=buttonDisabled
+    class="btn-primary"
```
tgxworld, line 18:
```suggestion
    class="btn-danger"
```

What do you think about making the button red instead?

assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
martin-brennan: Yes, and I've already used the same pattern in the Archive modal. You get shown a message before this happens so it shouldn't be too surprising.

Maybe there is some way to cancel this when the modal closes?

assets/javascripts/discourse/templates/components/chat-channel-delete-modal-inner.hbs
martin-brennan: It did cross my mind, I think I will change :+1:

config/locales/client.en.yml
```diff
@@ -19,10 +20,10 @@ en:
...
       channel_settings:
         title: "Channel settings"
-        close_channel: "Close Channel"
-        open_channel: "Open Channel"
-        archive_channel: "Archive Channel"
-        delete_channel: "Delete Channel"
+        close_channel: "Close Channel..."
+        open_channel: "Open Channel..."
+        archive_channel: "Archive Channel..."
+        delete_channel: "Delete Channel..."
```
tgxworld, line 26: :thinking: What does `...` mean here?

config/locales/client.en.yml
```diff
@@ -34,6 +35,13 @@ en:
       channel_close:
         title: "Close Channel"
         instructions: "Closing the channel prevents non-staff users from sending new messages or editing existing messages. Are you sure you want to close this channel?"
+      channel_delete:
+        title: "Delete Channel"
+        instructions: "<p>Deletes the <strong>%{name}</strong> channel and chat history. All messages and related data, such as reactions and uploads, will be permanently deleted. If you want to preserve the channel history and decomission it, you may want to Archive the channel instead.</p>
```
tgxworld, line 40:
```suggestion
        instructions: "<p>Deletes the <strong>%{name}</strong> channel and chat history. All messages and related data, such as reactions and uploads, will be permanently deleted. If you want to preserve the channel history and decomission it, you may want to archive the channel instead.</p>
```

tgxworld: I also wonder if it'll be helpful if we can link `archive` to the archiving modal? Only if it is easy :)

config/locales/client.en.yml
```diff
@@ -34,6 +35,13 @@ en:
...
+        <p>Are you sure you want to <strong>permanently delete</strong> the channel? To confirm, type the name of the channel in the box below.</p>"
+        confirm: "I understand the consequences, delete the channel"
+        confirm_channel_name: "Enter channel name"
+        process_started: "Deleting process has started. This page will be reloaded shortly, you will no longer see the deleted channel anywhere."
```
tgxworld, line 44:
```suggestion
        process_started: "The process to delete the channel has started. This page will be reloaded shortly, you will no longer see the deleted channel anywhere."
```

app/jobs/regular/chat_channel_delete.rb
martin-brennan: @tgxworld I don't think this will be a problem at all, I just deleted 10k messages in milliseconds, IMO we come back to it  if it becomes a problem. I do think maybe I can get rid of DistributedMutex though?

```
  ChatChannel Load (0.4ms)  SELECT "chat_channels".* FROM "chat_channels" WHERE "chat_channels"."id" = 18 LIMIT 1
Deleting webhooks and events for channel 18
  TRANSACTION (0.4ms)  BEGIN
  ChatWebhookEvent Destroy (0.4ms)  DELETE FROM "chat_webhook_events" WHERE "chat_webhook_events"."incoming_chat_webhook_id" IN (SELECT "incoming_chat_webhooks"."id" FROM "incoming_chat_webhooks" WHERE "incoming_chat_webhooks"."chat_channel_id" = 18)
  IncomingChatWebhook Destroy (0.2ms)  DELETE FROM "incoming_chat_webhooks" WHERE "incoming_chat_webhooks"."chat_channel_id" = 18
  TRANSACTION (0.4ms)  COMMIT
Deleting drafts and memberships for channel 18
  ChatDraft Destroy (0.4ms)  DELETE FROM "chat_drafts" WHERE "chat_drafts"."chat_channel_id" = 18
  UserChatChannelMembership Destroy (2.1ms)  DELETE FROM "user_chat_channel_memberships" WHERE "user_chat_channel_memberships"."chat_channel_id" = 18
Deleting chat messages, mentions, revisions, and uploads for channel 18
  TRANSACTION (0.5ms)  BEGIN
  ChatMention Destroy (0.7ms)  DELETE FROM "chat_mentions" WHERE "chat_mentions"."chat_message_id" IN (SELECT "chat_messages"."id" FROM "chat_messages" WHERE "chat_messages"."deleted_at" IS NULL AND "chat_messages"."chat_channel_id" = 18)
  ChatMessageRevision Destroy (0.5ms)  DELETE FROM "chat_message_revisions" WHERE "chat_message_revisions"."chat_message_id" IN (SELECT "chat_messages"."id" FROM "chat_messages" WHERE "chat_messages"."deleted_at" IS NULL AND "chat_messages"."chat_channel_id" = 18)
  ChatUpload Destroy (0.6ms)  DELETE FROM "chat_uploads" WHERE "chat_uploads"."chat_message_id" IN (SELECT "chat_messages"."id" FROM "chat_messages" WHERE "chat_messages"."deleted_at" IS NULL AND "chat_messages"."chat_channel_id" = 18)
  ChatMessage Update All (55.4ms)  UPDATE "chat_messages" SET "deleted_by_id" = NULL, "deleted_at" = '2022-03-17 06:45:01.458358' WHERE "chat_messages"."deleted_at" IS NULL AND "chat_messages"."chat_channel_id" = 18
  TRANSACTION (7.7ms)  COMMIT
  0.034136   0.008146   0.042282 (  0.125489)
```

config/locales/client.en.yml
martin-brennan: See internal chat :)

assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
tgxworld: Hmm how about just displaying a different view for the chat channel once the delete request is successful? Deferring a page reload based on an arbitrary duration feels like it will be hard to get right UX wise. I might be reading through the message in the modal midway when the page is refreshed or I might have closed the modal.

app/jobs/regular/chat_channel_delete.rb
tgxworld: Ah it seems like validity is just used to log a warning message if the lock is held for too long. I think we can just rely on the default here which is 60 seconds.

assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
martin-brennan: Fair enough will think of something else for this and the archive modal, thinking closer about it the refresh isn't actually necessary. The user is on the /chat/browse page when they do this action.

app/jobs/regular/chat_channel_delete.rb
tgxworld: This wouldn't be a problem now that we're not plucking the `id`s. Previously, each pluck will allocate an array the size of the number of messages in the channel.

app/jobs/regular/chat_channel_delete.rb
martin-brennan: Okay will change to 60s :+1:

assets/javascripts/discourse/components/chat-channel-delete-modal-inner.js
```diff
@@ -0,0 +1,55 @@
...
+import { later } from "@ember/runloop";
+import { isEmpty } from "@ember/utils";
+import I18n from "I18n";
+import discourseComputed from "discourse-common/utils/decorators";
+import { action } from "@ember/object";
+import { ajax } from "discourse/lib/ajax";
+import { inject as service } from "@ember/service";
+import { popupAjaxError } from "discourse/lib/ajax-error";
+
+export default Component.extend({
```
tgxworld, line 12: I think it is worth having an acceptance test here to test the whole deletion flow.

martin-brennan: Ah yep no problem, will add one, cause I have one for Archive + Change Status too

spec/jobs/chat_channel_delete_spec.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
```
tgxworld, line 3:
```suggestion
```

Btw this is no longer required in all spec files going forward. See https://github.com/discourse/discourse/commit/c9dab6fd087dd400c904441657f53134393fb298.

spec/jobs/chat_channel_delete_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+  fab!(:chat_channel) { Fabricate(:chat_channel) }
+  fab!(:user1) { Fabricate(:user) }
+  fab!(:user2) { Fabricate(:user) }
+  fab!(:user3) { Fabricate(:user) }
+  let(:users) { [user1, user2, user3] }
+
+  before do
+    messages = []
+    20.times do
+      messages << Fabricate(:chat_message, chat_channel: chat_channel, user: users.sample)
```
tgxworld, line 15: The usage of `sample` here makes me nervous because the start state of every test run is going to be different. If the start state is different, we cannot rely on the failures of a previous test run to debug the next test route.

spec/jobs/chat_channel_delete_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+  it "deletes all of the messages and related records completely" do
+    described_class.new.execute(chat_channel_id: chat_channel.id)
+    expect(IncomingChatWebhook.where(chat_channel_id: chat_channel.id).count).to eq(0)
+    expect(ChatWebhookEvent.where(incoming_chat_webhook_id: @incoming_chat_webhook_id).count).to eq(0)
+    expect(ChatDraft.where(chat_channel: chat_channel).count).to eq(0)
+    expect(UserChatChannelMembership.where(chat_channel: chat_channel).count).to eq(0)
+    expect(ChatMessageRevision.where(chat_message_id: @message_ids).count).to eq(0)
+    expect(ChatMention.where(chat_message_id: @message_ids).count).to eq(0)
+    expect(ChatUpload.where(chat_message_id: @message_ids).count).to eq(0)
+    expect(ChatMessage.where(id: @message_ids).count).to eq(0)
```
tgxworld, line 64: Usually I like to err on the side of caution when the final count of something is going to be `0`. I prefer to assert for state change rather than just the final state because it prevents false positive should someone accidentally nuke a line or two in the `before` block.

config/locales/client.en.yml
```diff
@@ -34,6 +35,13 @@ en:
       channel_close:
         title: "Close Channel"
         instructions: "Closing the channel prevents non-staff users from sending new messages or editing existing messages. Are you sure you want to close this channel?"
+      channel_delete:
+        title: "Delete Channel"
+        instructions: "<p>Deletes the <strong>%{name}</strong> channel and chat history. All messages and related data, such as reactions and uploads, will be permanently deleted. If you want to preserve the channel history and decomission it, you may want to archive the channel instead.</p>
+        <p>Are you sure you want to <strong>permanently delete</strong> the channel? To confirm, type the name of the channel in the box below.</p>"
```
tgxworld, line 41: I think this was briefly mentioned in dev but should we be including the channel name in this message?

app/controllers/chat_channels_controller.rb
```diff
@@ -262,6 +262,38 @@ def change_status
...
+  def destroy
+    params.require(:chat_channel_id)
+    params.require(:channel_name_confirmation)
+
+    chat_channel = ChatChannel.find_by(id: params[:chat_channel_id])
+    raise Discourse::NotFound if chat_channel.blank?
+
+    guardian.ensure_can_delete_chat_channel!
+
+    if chat_channel.name.downcase != params[:channel_name_confirmation].downcase
```
tgxworld, line 274:
```suggestion
    if chat_channel.name.downcase != params[:channel_name_confirmation]
```

I don't think we need to downcase the param on the server side since we're expecting a name that is already lowercased?

spec/requests/chat_channel_controller_spec.rb
```diff
@@ -656,4 +656,49 @@
...
+    fab!(:channel) do
+      Fabricate(
+        :chat_channel,
+        chatable: topic,
+        name: "Ambrose Channel",
+        status: :open
+      )
+    end
+
+    it "returns error if user is not admin" do
```
tgxworld, line 670:
```suggestion
    it "returns error if user is not staff" do
```

spec/requests/chat_channel_controller_spec.rb
```diff
@@ -656,4 +656,49 @@
...
+      sign_in(admin)
+      delete "/chat/chat_channels/#{channel.id}.json", params: { channel_name_confirmation: "some Other channel" }
+      expect(response.status).to eq(400)
+    end
+
+    it "deletes the channel right away and enqueues the background job to delete all its chat messages and related content" do
+      sign_in(admin)
+      delete "/chat/chat_channels/#{channel.id}.json", params: { channel_name_confirmation: "ambrose channel" }
+      expect(response.status).to eq(200)
+      expect(ChatChannel.find_by(id: channel.id)).to eq(nil)
```
tgxworld, line 693:
```suggestion
      expect(channel.reload.trashed?).to eq(true)
```

I think this is a more accurate representation of what we're asserting against. The query assumes the `default_scope` context exists on the model.

spec/jobs/chat_channel_delete_spec.rb
martin-brennan: Ooh nice!

config/locales/client.en.yml
martin-brennan: Yes, we do already? "Deletes the <strong>%{name}</strong> channel"

app/controllers/chat_channels_controller.rb
martin-brennan: IMO the server should do the same thing as the frontend, you could  send a DELETE request independently from the frontend

spec/requests/chat_channel_controller_spec.rb
martin-brennan: It always does though, I don't think that's a bad assumption to make if we trust `Trashable`

spec/jobs/chat_channel_delete_spec.rb
martin-brennan: The state change is that they are all deleted though? I don't know checking for count 0 wouldn't be valid here when we know exactly what we are deleting

spec/jobs/chat_channel_delete_spec.rb
martin-brennan: I don't think this is a problem, `users.sample` is just used here to avoid creating tons of extra user records unnecessarily from fabricators.

spec/jobs/chat_channel_delete_spec.rb
martin-brennan: Would be nice to have changes like this posted on dev tbh, if you're not watching the core commits you're likely to miss stuff like this.

app/controllers/chat_channels_controller.rb
tgxworld: But we're not exposing this as an API outside of use for the client so I don't think we need to handle cases where people send a DELETE request independently from the frontend. I do wonder if it is even necessary for us to be comparing the channel's name on the server side. If it is, we can just drop the `chat_channel_id` params because both are unique columns that can be used to identify the same resource.

config/locales/client.en.yml
tgxworld: Ahh I didn't catch this when I was looking at the screenshot. I feel like it isn't as clear as what Github does but we can always try this out first and refine later.

spec/jobs/chat_channel_delete_spec.rb
tgxworld: Asserting for state change would be something like `expect { }.to change { ... }.from(x).to(y)` or 

```
expect(ChatMessage.where(id: @message_ids).count).to eq(1)
described_class.new.execute(chat_channel_id: chat_channel.id)
expect(ChatMessage.where(id: @message_ids).count).to eq(0)
```

Right now I can remove `ChatDraft.create(chat_channel: chat_channel, user: users.sample, data: "wow some draft")` from the `before` block accidentally and this test will still pass.

spec/requests/chat_channel_controller_spec.rb
tgxworld: But why introduce an indirection when we can assert for the state of the object directly?

tgxworld: There is an additional level of indirection introduced here though compared to asserting for the state of the object directly.

app/controllers/chat_channels_controller.rb
martin-brennan: I mean...by default we are exposing everything. IMO we should just leave this as is, it adds an extra layer of confirmation and I would rather look up by ID than name.

spec/jobs/chat_channel_delete_spec.rb
martin-brennan: Will change :+1:

spec/requests/chat_channel_controller_spec.rb
martin-brennan: Okay understood, will change :+1:

app/controllers/chat_channels_controller.rb
martin-brennan: And by default exposing everything I mean there is no way to stop anyone just using this as is. I will check what github POSTS when you delete a repo, and see if they send through the confirmation, and if not then I will remove it here

tgxworld: Yea I do wonder what Github does in this case too. I've always thought having to type the name of the repo is just to make sure that the trigger to delete the repository is an intentional one.

martin-brennan:
```
_method: delete
authenticity_token: w8gErekdNAu5gEeIaYMyVmkyo73PLra5dhpD9Ex6LEIT4V9yevmdjHVnSPkMWKCtUS5-eRqYKGhHo1hNnIp29Q
repository: blahblahdeletetest
sudo_referrer: https://github.com/martin-brennan/blahblahdeletetest/settings
user_id: martin-brennan
verify: martin-brennan/blahblahdeletetest
```

GitHub does the same thing with the verify param :)

tgxworld: OK :ok_hand: Let's keep it as it is then :)

app/jobs/regular/chat_channel_delete.rb
tgxworld, line 16: I think we've decided to remove this?

martin-brennan: Yah just hadn't pushed it yet, just pushed it a minute ago :) Still working on some more changes, adding the acceptance test now

app/jobs/regular/chat_channel_delete.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      # this should not really happen, but better to do this than throw an error
+      if chat_channel.blank?
+        Rails.logger.warn("Chat channel #{args[:chat_channel_id]} could not be found, aborting delete job.")
+        return
+      end
+
+      DistributedMutex.synchronize(
+        "delete_chat_channel_#{chat_channel.id}",
+        validity: 1.minute
```
tgxworld, line 16: Sorry what I meant previously is that we should just follow whatever default that comes with `DistributedMutex` unless we think there is a chance this will hit 1 minute.

martin-brennan: Oh right, the default is 1m but I don't need to tell it that, will remove

test/javascripts/acceptance/delete-channel-test.js
```diff
@@ -0,0 +1,113 @@
...
+  server.get("/chat/:chatChannelId/messages.json", () =>
+    helper.response(chatView)
+  );
+  server.post("/chat/:chatChannelId.json", () => {
+    return helper.response({ success: "OK" });
+  });
+  server.get("/chat/lookup/:message_id.json", () => helper.response(chatView));
+  server.post("/uploads/lookup-urls", () => {
+    return helper.response([]);
+  });
```
tgxworld, line 24: Minor nit here but we're mixing cases here. `:chatChannelId` and `:message_id`. I think we usually just follow snake case in URLs.

tgxworld (approved):
Looks good to me :+1:  Thank you for your work here.

martin-brennan:
No probs, thanks for reviewing! :+1:

PR #713, opened 2022-03-17
FIX: displays unread-indicator only when needed

jjaffeux:
https://github.com/discourse/discourse-chat/commit/aed2d84add728edcb49b4719948292ef136fd02c made it appear on places where its interest is very limited due to the fact it becomes instantly read and as a result just flashes the dot.

### DIFF
--- a/test/javascripts/components/chat-channel-title-test.js
+++ b/test/javascripts/components/chat-channel-title-test.js
@@ -10,6 +10,7 @@ import {
 import hbs from "htmlbars-inline-precompile";
 import fabricate from "../helpers/fabricators";
 import { CHATABLE_TYPES } from "discourse/plugins/discourse-chat/discourse/models/chat-channel";
+import { set } from "@ember/object";
 
 discourseModule(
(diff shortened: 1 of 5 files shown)


test/javascripts/components/chat-channel-title-test.js
```diff
@@ -184,5 +185,33 @@ discourseModule(
...
+      beforeEach() {
+        const channel = fabricate("chat-channel", {
+          chatable_type: CHATABLE_TYPES.directMessageChannel,
+        });
+
+        const state = {};
+        state[channel.id] = {
+          unread_count: 1,
+        };
+        set(this.currentUser, "chat_channel_tracking_state", state);
```
CvX, line 201: Why `set(this.currentUser)` and not `this.currentUser.set()`? 😃

CvX (approved):

test/javascripts/components/chat-channel-title-test.js
jjaffeux: no specific reason, I changed to this.currentUser.set

PR #714, opened 2022-03-17
FIX: we were comparing user object and id

jjaffeux:


jjaffeux:
I think you wrote this in this commit @markvanlan and that looks wrong?

 https://github.com/discourse/discourse-chat/commit/1502201194548474f3600410d6222a7a4b401884#diff-7bdd9b24dbf8b6da550a287db4a0937f314a61fff49203543a292166fef8f404R22

markvanlan (approved):
Nice

PR #716, opened 2022-03-17
FIX: ensures only one msg-action on mobile

jjaffeux:


assets/javascripts/discourse/components/chat-message.js
```diff
@@ -639,7 +641,7 @@ export default Component.extend({
     this.notifyPropertyChange("emojiReactions");
 
     if (this.site.mobileView) {
-      this.toggleProperty("isHovered");
+      this.set("isHovered", false);
```
jjaffeux, line 644: Toggle seems un-necessary here, if we reacted we want to close the menu 100%.

assets/javascripts/discourse/components/chat-message.js
```diff
@@ -230,7 +231,8 @@ export default Component.extend({
   @action
   handleClick() {
     if (this.site.mobileView) {
-      this.toggleProperty("isHovered");
+      this.onHoverMessage && this.onHoverMessage(this.message);
```
CvX, line 234: 🏌️ ⛳

```suggestion
      this.onHoverMessage?.(this.message);
```

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1222,6 +1223,14 @@ export default Component.extend({
     }
   },
 
+  @action
+  onHoverMessage(message) {
+    this.set(
+      "hoveredMessageId",
+      message?.id && message.id !== this.hoveredMessageId ? message.id : null
```
markvanlan, line 1230: This is difficult to read but.. if it works that's ok

markvanlan (approved):

assets/javascripts/discourse/components/chat-live-pane.js
jjaffeux: yep there are various ways to write it, I finally stopped on this one, I think that's okayish.

PR #718, opened 2022-03-21
FIX: Clipboard copy chat quotes breaks on Safari

martin-brennan:
Safari among the other browsers is doing the correct thing, and
throwing the following error when the user is clicking the
Quote button in chat isolated mode:

> NotAllowedError: the request is not allowed by the user agent or the
  platform in the current context, possibly because the user denied permission.

This is because the clipboard API is supposed to be used in a
short time period to respond to a user-initiated event. Luckily,
there is a way around this, which is to hand a promise off to
a ClipboardItem. This is done via a new clipboardCopyAsync function
created in core.

This commit uses the new function and rewrites other parts of the
chat quote code to work around its usage.

cf. https://github.com/discourse/discourse/pull/16232

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -1,5 +1,5 @@
 import Component from "@ember/component";
-import { clipboardCopy } from "discourse/lib/utilities";
+import { clipboardCopyAsync } from "discourse/lib/utilities";
 import { getOwner } from "discourse-common/lib/get-owner";
 import discourseComputed, {
   afterRender,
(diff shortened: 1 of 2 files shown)


test/javascripts/acceptance/chat-quoting-test.js
```diff
@@ -197,10 +197,6 @@ acceptance(
         exists("#reply-control.composer-action-createTopic"),
         "the composer does not open"
       );
-      assert.ok(
```
martin-brennan, line 200: Not really possible to test this now with the way the clipboard promise is structured, we skip copying to clipboard in tests anyway

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1,5 +1,5 @@
 import Component from "@ember/component";
-import { clipboardCopy } from "discourse/lib/utilities";
+import { clipboardCopyAsync } from "discourse/lib/utilities";
```
lis2, line 2: Maybe we should add some discourse compatibility, so the plugin will not crash on older versions of Discourse. Something like https://github.com/discourse/discourse-assign/blob/main/.discourse-compatibility

lis2 (approved):

PR #720, opened 2022-03-21
FEATURE: Remove move to topic

martin-brennan:
This feature does not exist in a form we are
happy with right now, we are removing it for
now and may come back to it later. Copying
chat messages one to one to posts is not what
we want. The alternative after this is
removed is to use chat quotes instead.

I will drop the post connections table in a
later PR


tgxworld (approved):
Tests are passing and this looks pretty low risk :)

PR #721, opened 2022-03-21
FIX: Do not autofocus chat composer when ? pressed

martin-brennan:
The ? key is used to show the keyboard shortcuts in the
app, we want this to work rather than autofocusing the chat
composer.

### DIFF
--- a/assets/javascripts/discourse/components/full-page-chat.js
+++ b/assets/javascripts/discourse/components/full-page-chat.js
@@ -55,31 +55,35 @@ export default Component.extend({
   },
 
   @bind
-  _autoFocusChatComposer(e) {
+  _autoFocusChatComposer(event) {
     if (
-      !e.key ||
-      e.key.length > 1 ||
-      e.metaKey ||
-      e.ctrlKey ||
-      e.code === "Space"
+      !event.key ||
+      // Handles things like Enter, Tab, Shift
+      event.key.length > 1 ||
+      // Don't need to focus if the user is beginning a shortcut.
+      event.metaKey ||
+      event.ctrlKey ||
+      // Space's key comes through as ' ' so it's not covered by event.key
+      event.code === "Space" ||
+      // ? is used for the keyboard shortcut modal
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/components/full-page-chat.js
```diff
@@ -55,31 +55,35 @@ export default Component.extend({
...
+      !event.key ||
+      // Handles things like Enter, Tab, Shift
+      event.key.length > 1 ||
+      // Don't need to focus if the user is beginning a shortcut.
+      event.metaKey ||
+      event.ctrlKey ||
+      // Space's key comes through as ' ' so it's not covered by event.key
+      event.code === "Space" ||
+      // ? is used for the keyboard shortcut modal
+      event.code === "Slash"
```
tgxworld, line 69: Just to confirm, is the code for `?` represented by `Slash`? On first read, I thought `Slash` would either mean a forward or back slash.

tgxworld (approved):
Just need to confirm one comment but other wise the fix looks great to me :+1:

assets/javascripts/discourse/components/full-page-chat.js
martin-brennan: @tgxworld yeah it's weird...this "code" property represents the physical keyboard key https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code. I would have used keyCode but I wanted to match the code style here and apparently that's deprecated anyway! https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

So yeah, the physical key is called Slash because that's what you get without pressing Shift `/`. I guess we could further narrow to only return if code is Slash and event.shiftKey at the same time, but IMO not worth it. Maybe if someone complains about `/` not focusing the composer we can revisit :D

tgxworld: Thank you for the detailed explanation :+1:

PR #725, opened 2022-03-21
FIX: ensures chat-composer-details uses available space

jjaffeux:


jjaffeux:
This fix avoids this case:

Screenshot 2022-03-21 at 10 37 35

CvX (approved):

PR #729, opened 2022-03-22
DEV: Drop chat_message_post_connections

martin-brennan:
Since 613ff279c3c396fb84ca2911efe68aeaaa739bc0 we
no longer need this table.

### DIFF
--- a/db/post_migrate/20220321235638_drop_chat_message_post_connections_table.rb
+++ b/db/post_migrate/20220321235638_drop_chat_message_post_connections_table.rb
@@ -3,14 +3,8 @@
 require 'migration/table_dropper'
 
 class DropChatMessagePostConnectionsTable < ActiveRecord::Migration[6.1]
-  DROPPED_TABLES ||= %i{
-    chat_message_post_connections
-  }
-
   def up
-    DROPPED_TABLES.each do |table|
-      Migration::TableDropper.execute_drop(table)
-    end
+    Migration::TableDropper.execute_drop("chat_message_post_connections")
   end
 
   def down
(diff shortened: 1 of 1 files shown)


db/post_migrate/20220321235638_drop_chat_message_post_connections_table.rb
```diff
@@ -0,0 +1,19 @@
...
+
+require 'migration/table_dropper'
+
+class DropChatMessagePostConnectionsTable < ActiveRecord::Migration[6.1]
+  DROPPED_TABLES ||= %i{
+    chat_message_post_connections
+  }
+
+  def up
+    DROPPED_TABLES.each do |table|
```
danielwaterworth, line 11: Why the loop, are you expecting DROPPED_TABLES to be populated elsewhere?

martin-brennan: No idea TBH, this just seems to be the pattern we have adopted for them, even for single tables:

https://github.com/discourse/discourse/blob/31bc756037834b531750167f029c734e12040db9/db/migrate/20201110110952_drop_github_user_infos.rb

Doesn't really make sense though you're right, will get rid of the loop and the const, it's not necessary

danielwaterworth (approved):

db/post_migrate/20220321235638_drop_chat_message_post_connections_table.rb
tgxworld: This is not well documented by we actually need the `DROPPED_TABLES` constants for backup/restore. See https://github.com/discourse/discourse/commit/7cb51d0e4069ef982919e9f005e6bb18da04ebed for more details. In core, we actually have a spec to catch this.

PR #730, opened 2022-03-22
FEATURE: Insert link with shortcut in composer

martin-brennan:
This commit adds the ctrl+l shortcut to open the
insert link modal in the chat composer (ctrl+k
was already taken by channel switcher)

image

This shortcut only works if the chat composer is
focused.

### DIFF
--- a/assets/javascripts/discourse/initializers/chat-keyboard-shortcuts.js
+++ b/assets/javascripts/discourse/initializers/chat-keyboard-shortcuts.js
@@ -101,7 +101,17 @@ export default {
       api.addKeyboardShortcut(
         `${mod}+l`,
         (event) => openInsertLinkModal(event),
-        { global: true }
+        {
+          global: true,
+          help: {
+            category: "chat",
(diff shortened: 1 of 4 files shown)


nattsw (approved):
👩‍🍳 😽

assets/javascripts/discourse/initializers/chat-keyboard-shortcuts.js
```diff
@@ -89,6 +98,11 @@ export default {
         (event) => modifyComposerSelection(event, "code"),
         { global: true }
       );
+      api.addKeyboardShortcut(
+        `${mod}+l`,
```
nattsw, line 102: I feel like I asked before.. do we register any chat shortcuts into a list?

martin-brennan: @nattsw you are right! I was looking at that help modal just this morning! Added a message for insert link now :)

PR #738, opened 2022-03-23
FIX: allow new duplicate reactions when max new reactions are reached

techAPJ:


lib/chat_message_reactor.rb
```diff
@@ -50,9 +50,10 @@ def validate_channel_status!
     )
   end
 
-  def validate_max_reactions!(react_action)
+  def validate_max_reactions!(react_action, emoji)
     if react_action == ADD_REACTION &&
-      @chat_message.reactions.count('DISTINCT emoji') >= MAX_REACTIONS_LIMIT
+      @chat_message.reactions.count('DISTINCT emoji') >= MAX_REACTIONS_LIMIT &&
+      !@chat_message.reactions.exists?(emoji: emoji)
```
markvanlan, line 56: Does this cause another query? It would be great to avoid, I can't say with certianty that it _doesn't_.. approving and maybe you can look?

markvanlan (approved):

lib/chat_message_reactor.rb
techAPJ: This will cause another query only when the condition `@chat_message.reactions.count('DISTINCT emoji') >= MAX_REACTIONS_LIMIT` is `true`. So it will not run for most of the cases. I am not sure if we can skip the query altogether because we need to check the new reaction emoji against already reacted emojis.

PR #739, opened 2022-03-23
UX: implements a dedicated mobile menu for messages

jjaffeux:
Co-Authored-By: Chapoi <101828855+chapoi@users.noreply.github.com>

### DIFF
--- a/assets/javascripts/discourse/components/chat-message-actions-mobile.js
+++ b/assets/javascripts/discourse/components/chat-message-actions-mobile.js
@@ -13,6 +13,6 @@ export default Component.extend({
   _addFadeIn() {
     document
       .querySelector(".chat-msgactions-backdrop")
-      ?.classList?.add("fade-in");
+      ?.classList.add("fade-in");
   },
 });
--- a/assets/javascripts/discourse/components/chat-message-actions-mobile.js
+++ b/assets/javascripts/discourse/components/chat-message-actions-mobile.js
@@ -1,0 +1,18 @@
+import Component from "@ember/component";
+import { later } from "@ember/runloop";
+
+export default Component.extend({
+  tagName: "",
+
+  didInsertElement() {
+    this._super(...arguments);
+
+    later(this._addFadeIn);
+  },
+
+  _addFadeIn() {
+    document
+      .querySelector(".chat-msgactions-backdrop")
+      ?.classList?.add("fade-in");
+  },
+});
--- a/assets/javascripts/discourse/templates/components/chat-message.hbs
+++ b/assets/javascripts/discourse/templates/components/chat-message.hbs
@@ -16,7 +16,7 @@
     {{#if selectingMessages}}
       {{input
         type="checkbox"
-        class="chat-message-selector"
+        class="chat-message-selector catch-click"
         checked=message.selected
         click=(action "toggleChecked")
       }}
--- a/assets/javascripts/discourse/templates/components/chat-message.hbs
+++ b/assets/javascripts/discourse/templates/components/chat-message.hbs
@@ -33,45 +33,13 @@
     {{else}}
       <div class={{chatMessageClasses}}>
         {{#if showActions}}
-          <div class="chat-msgactions-hover">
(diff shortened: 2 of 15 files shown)


jjaffeux:
Screenshot 2022-03-23 at 17 27 52

jjaffeux:
Kapture 2022-03-23 at 17 58 33

assets/javascripts/discourse/components/chat-message-actions-mobile.js
```diff
@@ -0,0 +1,18 @@
...
+  didInsertElement() {
+    this._super(...arguments);
+
+    later(this._addFadeIn);
+  },
+
+  _addFadeIn() {
+    document
+      .querySelector(".chat-msgactions-backdrop")
+      ?.classList?.add("fade-in");
```
pmusaraj, line 16: Almost not worth mentioning... 
```suggestion
      ?.classList.add("fade-in");
```

assets/javascripts/discourse/templates/components/chat-message.hbs
```diff
@@ -33,45 +33,13 @@
...
-                {{dropdown-select-box
-                  class="more-buttons"
-                  options=(hash icon="ellipsis-h")
-                  content=moreButtons
-                  onChange=(action "handleMoreButtons")
-                }}
-              {{/if}}
-            </div>
-          </div>
+          {{component (concat "chat-message-actions-" (if site.mobileView "mobile" "desktop"))
```
pmusaraj, line 36: It may be better to avoid `concat` here, purely to allow future devs to more easily find "chat-message-actions-desktop" and "chat-message-actions-mobile".

assets/stylesheets/mobile/chat-message.scss
```diff
@@ -0,0 +1,141 @@
...
+  bottom: 0;
+  left: 0;
+  right: 0;
+  height: calc(100% - var(--header-offset));
+  width: 100%;
+  z-index: z("usercard");
+  transition: background-color 0.2s ease;
+
+  &.fade-in {
+    background-color: rgba(0, 0, 0, 0.75);
```
pmusaraj, line 23: This is a personal preference, but I'd like to try with 0.25 or 0.35 opacity. The background darkening feels way too strong, especially since the header is excluded. 

In fact, curious if we can try some other funky things here, like a bit of blurring maybe? (I am aware we are diverging from our norms, but chat feels like a good place to try new things.)

assets/stylesheets/mobile/chat-message.scss
```diff
@@ -0,0 +1,141 @@
...
+    transition: top 1s linear;
+  }
+}
+
+.chat-msgactions-backdrop {
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  height: calc(100% - var(--header-offset));
```
pmusaraj, line 17: IMO this would be fine to cover the header, i.e. `height: 100%`.

pmusaraj (commented):
This looks good. 

I wonder whether this is better done as a long-press on mobile. I pulled the branch locally, and used my phone, and it feels like a simple tap is too light, it launches a big visual change a little suddenly. Is there a way to have this be invoked only on longpress?

Also, small bug, if you invoke the "Select" option, a click on other chat messages will retrigger the actions overlay. (For example, checking the checkbox next to another message).

assets/javascripts/discourse/templates/components/chat-message.hbs
jjaffeux: initially I had duplicated both components, but I think I prefer having to deal only with updating one set of properties

assets/stylesheets/mobile/chat-message.scss
jjaffeux: Im not a fan of blurry background, and I would want to avoid future discussions on the PR to be about this, but I guess we can try in the future.

I did change to 0.35 though, seems better I agreed :+1:

pmusaraj (approved):

PR #740, opened 2022-03-23
FIX: Enable quoting in all cases

CvX:
1. The computed property didn't have its dependencies specified, which meant the button would never appear if the first chat you tried to use the option was private. (it would cache computed value)
2. It doesn't make much sense to prevent users from quoting private chat - there are legit cases when that's needed.


jjaffeux:
I just saw this bug with @pmusaraj while looking at other stuff  :+1:

martin-brennan:
@CvX @jjaffeux we already discussed internally that we don't want quoting to be enabled in DMs, see /t/58331/90. So, we should disable selecting messages in DMs, I probably missed that when I remove Move to Topic. Thank you for picking up this bug in general though!

martin-brennan:
Also note this is not just enforced on a UI level, it's enforced in the controller:

https://github.com/discourse/discourse-chat/blob/3bc372e79fe7ba2bcaa2131808d295ae1cdf3043/app/controllers/chat_controller.rb#L331-L342

pmusaraj (approved):

martin-brennan (approved):

PR #743, opened 2022-03-24
FIX: Fix missing app event bindings

martin-brennan:
In  https://github.com/discourse/discourse/commit/1341baaebaf9932952cf3fd65448564473188525 in core
I added composerEventPrefix for composer app events,
but I only changed a few of them. Also using the
useUploadPlaceholders: false option to avoid any text
being inserted into the chat composer as result
of uploads, introduced in:

https://github.com/discourse/discourse/pull/16272


tgxworld (approved):
Approved first since this is urgent but can we follow up with tests?

tgxworld:
Tests have been added in https://github.com/discourse/discourse-chat/pull/743

PR #744, opened 2022-03-24
DEV: Add chat upload test

martin-brennan:
Previously skipped this one because it wasn't
working in CI, let's test it now and also test
for the composer not getting a placeholder for
the in progress upload

### DIFF
--- a/test/javascripts/acceptance/chat-test.js
+++ b/test/javascripts/acceptance/chat-test.js
@@ -17,7 +17,7 @@ import {
   triggerKeyEvent,
   visit,
 } from "@ember/test-helpers";
-import { skip, test } from "qunit";
+import { test } from "qunit";
 import {
   allChannels,
   chatChannels,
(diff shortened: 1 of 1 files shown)


tgxworld (approved):

test/javascripts/acceptance/chat-test.js
```diff
@@ -1515,18 +1515,32 @@ acceptance("Discourse Chat - image uploads", function (needs) {
     const appEvents = loggedInUser().appEvents;
     const done = assert.async();
 
-    appEvents.on("chat-composer:all-uploads-complete", () => {
+    appEvents.on("chat-composer:all-uploads-complete", async () => {
```
tgxworld, line 1518: Actually do you know if the test will block on this? I have a feeling that the test will move on after setting the listeners on the appEvents. That means by the time the assertions are called, it may not be the right screen anymore.

PR #745, opened 2022-03-24
FEATURE: add an option to disable channel-wide mention notifications

techAPJ:
Screen Shot 2022-03-24 at 16 07 47


config/locales/client.en.yml
```diff
@@ -92,6 +92,9 @@ en:
       only_chat_push_notifications:
         title: "Only send chat push notifications"
         description: "Block all non-chat push notifications from being sent"
+      ignore_channel_wide_mention:
+        title: "Ignore channel-wide mentions"
+        description: "Do not send notifications for channel-wide mentions like @here / @all"
```
markvanlan, line 97: The `like` makes it sound like there are more than just those. Maybe 
"Do not send notifications for channel-wide mentions (@here and @all)"

markvanlan (approved):

config/locales/client.en.yml
techAPJ: Agreed. I've made the copy change.

PR #752, opened 2022-03-28
DEV: Use new consistency-check-skipping notification type API

markvanlan:
Relies on new api added here - https://github.com/discourse/discourse/pull/16299


pmusaraj:
I don't think we need this anymore @markvanlan, closing.

PR #756, opened 2022-03-31
DEV: Add date button, use cleaner `applyLocalDates` import

pmusaraj:
Previously the date button was added in `discourse-local-dates`, see https://github.com/discourse/discourse/pull/16342 which removes it from core. 

This PR also 
 - fixes the date modal in chat, it previously was inserting `undefinedundefinedundefined` in the chat composer
 - adds an acceptance test
 - uses the new `applyLocalDates` wrapper from core


markvanlan (approved):
Nice nice

PR #759, opened 2022-04-01
FEATURE: Chat notification emails

markvanlan:
This PR adds `chat_message_email_statuses` which stores info on whether users need to be emailed for a particularly chat message. Records are created when chat messages are created and edited. 

We then have a `ChatMailer` class which runs via a job every 15 minutes, and enqueues a `user_email` job for each user with `unprocessed` `chat_message_email_statuses`.

Uses new param for `UserEmail` described here https://github.com/discourse/discourse/pull/16460

Screen Shot 2022-04-11 at 3 17 20 PM

Screen Shot 2022-04-11 at 3 18 48 PM


lib/extensions/user_notifications_extension.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    build_summary_for(user)
+    opts = {
+      from_alias: I18n.t('user_notifications.chat_summary.from', site_name: Email.site_title),
+      subject: I18n.t('user_notifications.chat_summary.subject', count: @email_statuses.count, email_prefix: @email_prefix, date: short_date(Time.now)),
+      add_unsubscribe_link: true,
+      unsubscribe_url: "#{Discourse.base_url}/email/unsubscribe/#{@unsubscribe_key}",
+    }
+
+    ChatMessageEmailStatus.where(id: @email_statuses.map(&:id)).update_all(status: ChatMessageEmailStatus::STATUSES[:processed])
```
pmusaraj, line 20: This feels like the wrong place to update all email status summaries. Could we do this in the job but _after_ having sent out the email? And maybe it would suffice to query by `user_id`, surely we want to mark as processed all existing email message statuses for said user?

app/models/chat_message_email_status.rb
```diff
@@ -0,0 +1,104 @@
...
+  def self.mentioned_statuses_for_message(chat_message)
+    self.where(chat_message: chat_message).where.not(type: TYPES[:regular])
+  end
+
+  def self.type_from_mention_info(mention_info)
+    return TYPES[:direct_mention] if mention_info.nil?
+    return TYPES[:group_mention] if mention_info[:is_group]
+    return TYPES[:global_mention] if [:all, :here].include?(mention_info[:identifier])
+  end
+end
```
pmusaraj, line 104: This model is missing annotations.

app/models/chat_channel.rb
```diff
@@ -118,6 +118,21 @@ def title_from_chatable
     end
   end
 
+  def email_title(user)
```
pmusaraj, line 121: We need a test for this.

app/models/chat_message_email_status.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  TYPES = {
+    regular: 0,
+    group_mention: 1,
+    global_mention: 2,
+    direct_mention: 3,
+  }
+
+  enum status: STATUSES
+  enum type: TYPES
```
pmusaraj, line 19: I'm on the fence about adding the `type` column at this time. If we're not using it yet, maybe it's best to leave it out for the current version?

config/locales/server.en.yml
```diff
@@ -104,3 +105,23 @@ en:
...
+    chat_summary:
+      deleted_user: "Deleted user"
+      description:
+        one: "You have a new chat message"
+        other: "You have new chat messages"
+      from: "%{site_name}"
+      subject:
+        one: "[%{email_prefix}] New chat message"
+        other: "[%{email_prefix}] New chat messages"
+      unsubscribe: "This email is sent from %{site_link}. Change your %{email_preferences_link}, or %{unsubscribe_link} to unsubscribe."
```
pmusaraj, line 119:
```suggestion
      unsubscribe: "This chat summary is sent from %{site_link} when you are away. Change your %{email_preferences_link}, or %{unsubscribe_link} to unsubscribe."
```

(Better matches equivalent email copy in core.)

pmusaraj (commented):
Looking generally quite good, the main comment is around the `type` column in `chat_message_email_status`.

lib/extensions/user_notifications_extension.rb
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @group_email_statuses = @email_statuses.group_by { |status| status.chat_message.chat_channel }
+    @group_email_statuses.each do |chat_channel, statuses|
+      @group_email_statuses[chat_channel] = statuses.sort_by(&:created_at)
+    end
+    @user = user
+
+    build_email(user.email, opts)
+    ChatMessageEmailStatus.where(user_id: user.id).update_all(status: ChatMessageEmailStatus::STATUSES[:processed])
```
markvanlan, line 27: @pmusaraj I moved this line to after the email is sent. I really don't want to enqueue ANOTHER job for this one query.. this is already happening in a job so it's not a performance issue.

lib/chat_mailer.rb
```diff
@@ -0,0 +1,34 @@
...
+  def self.mail_notifications
+    return unless SiteSetting.chat_enabled
+
+    instance = self.new
+    instance.mail_notifications
+  end
+
+  def mail_notifications
+    while users.exists?
+      users.limit(100).each do |user|
```
pmusaraj, line 13: Is this limit here for performance reasons?

markvanlan: Yeah I spoke with David about this, and this was his suggestion on how to make it performant and not a memory issue.

pmusaraj (approved):

PR #764, opened 2022-04-05
DEV: Extract uploads from chat-composer into a chat-composer-uploads component

martin-brennan:
This commit breaks up some of the tight coupling between the chat-composer component and uploads. It removes the use of `ComposerUppyUpload` which is more complex and heavily geared towards interactions with the core composer, including the inserting of upload placeholder text which we don't need at all for the chat composer. Instead we are using `UppyUploadMixin`, which needed some improvements made in https://github.com/discourse/discourse/pull/16383 to accommodate this change.

We cannot completely remove all coupling. The chat-composer needs to load in drafts and edit messages that may have uploads attached, and for that there is an app event introduced here called `chat-composer:load-uploads` that handles this preloading. This works the other way as well -- when uploads are complete in chat-composer-uploads or removed the chat-composer needs its internal state updated, for which I introduced an `uploadChanged` action.

The chat-composer-uploads component needs to bind the paste event for the chat-composer text area to allow for uploads to be pasted in directly.

Additionally, the uploads now shows "Processing..." instead of "Uploading..." when the preprocessing such as media optimization is still happening for each individual upload. Upload state is cleared up and cancelled properly if switching to editing a message while an upload is in progress.

The CSS for the chat-composer-upload(s) has also been extracted, and some improvements have been made to e.g. allow for scrolling on horizontal overflow.

image


assets/javascripts/discourse/components/chat-composer-uploads.js
```diff
@@ -0,0 +1,141 @@
...
+      );
+      if (!inProgressUpload?.processing) {
+        inProgressUpload?.set("processing", true);
+      }
+    });
+
+    this._onPreProcessComplete((file) => {
+      run(() => {
+        const inProgressUpload = this.inProgressUploads.find(
+          (upl) => upl.id === file.id
```
jjaffeux, line 113: findBy

assets/javascripts/discourse/components/chat-composer-uploads.js
```diff
@@ -0,0 +1,141 @@
...
+    this.setProperties({
+      uploads: [],
+      fileInputSelector: `#${this.fileUploadElementId}`,
+    });
+    this.appEvents.on("chat-composer:load-uploads", this, "_loadUploads");
+  },
+
+  didInsertElement() {
+    this._super(...arguments);
+    this.composerInputEl = document.querySelector(".chat-composer-input");
```
jjaffeux, line 29: maybe save the selector name as you reuse it multiple times

assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -217,25 +160,18 @@ export default Component.extend(TextareaTextManipulation, ComposerUploadUppy, {
       this,
       "_replyToMsgChanged"
     );
+    this.appEvents.off(
+      "upload-mixin:chat-composer-uploader:in-progress-uploads",
```
jjaffeux, line 164: Im mixed on the name of the event here, seems very tied to an implementation detail

assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -777,25 +698,8 @@ export default Component.extend(TextareaTextManipulation, ComposerUploadUppy, {
...
-  },
-
-  @discourseComputed("composerDisabled", "uploads.[]", "inProgressUploads.[]")
-  showUploadsContainer(composerDisabled, uploads, inProgressUploads) {
-    if (composerDisabled) {
-      return false;
-    }
-
-    return uploads?.length > 0 || inProgressUploads?.length > 0;
+  uploadChanged(uploads) {
```
jjaffeux, line 701: its somewhat unexpected that method name is singular and it takes an array as arg

assets/stylesheets/common/chat-composer-upload.scss
```diff
@@ -0,0 +1,71 @@
...
+      max-width: 100%;
+      max-height: 100%;
+    }
+  }
+
+  .data {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    line-height: 1.5em;
```
jjaffeux, line 35: dont we have some var for line heights?

assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -322,28 +249,63 @@ export default Component.extend(TextareaTextManipulation, ComposerUploadUppy, {
...
 
+  // the chat-composer needs to be able to set the internal list of uploads
+  // for chat-composer-uploads to preload in existing uploads for drafts
+  // and for when messages are being edited.
+  //
+  // the opposite is true as well -- when an upload is completed the chat-composer
+  // needs its internal state updated so drafts can be saved, which is handled
+  // by the uploadChanged action
+  _syncUploads(newUploads = []) {
+    const currentUploadIds = this._uploads.map((upl) => upl.id);
```
jjaffeux, line 284: mapBy

assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -322,28 +249,63 @@ export default Component.extend(TextareaTextManipulation, ComposerUploadUppy, {
...
+    const newUploadIds = newUploads.map((upl) => upl.id);
```
jjaffeux, line 285: mapBy

assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -47,33 +41,17 @@ export default Component.extend(TextareaTextManipulation, ComposerUploadUppy, {
...
-  composerModelContentKey: "value",
-  editorInputClass: ".chat-composer-input",
-  showCancelBtn: or("isUploading", "isProcessingUpload"),
-  uploadCancelled: false,
-  uploadProcessorActions: null,
-  uploadPreProcessors: null,
-  uploadMarkdownResolvers: null,
-  uploadType: "chat-composer",
-  uppyId: "chat-composer-uppy",
-  editorClass: alias("editorInputClass"),
```
jjaffeux, line 77: I dont see onValueChange initialized here

assets/javascripts/discourse/components/chat-composer-uploads.js
```diff
@@ -0,0 +1,141 @@
...
+        );
+        inProgressUpload?.set("processing", false);
+      });
+    });
+  },
+
+  @bind
+  _pasteEventListener(event) {
+    if (
+      document.activeElement !== document.querySelector(".chat-composer-input")
```
jjaffeux, line 123: use the constant mentionned sooner for the selector

assets/javascripts/discourse/components/chat-composer-uploads.js
```diff
@@ -0,0 +1,141 @@
...
+      this._pasteEventListener
+    );
+  },
+
+  uploadDone(upload) {
+    this.uploads.pushObject(upload);
+    this.onUploadChanged(this.uploads);
+  },
+
+  @discourseComputed("uploads.[]", "inProgressUploads.[]")
```
jjaffeux, line 47: wonder if you could direct track .length here?

assets/javascripts/discourse/components/chat-composer-uploads.js
```diff
@@ -0,0 +1,141 @@
...
+      const inProgressUpload = this.inProgressUploads.find(
+        (upl) => upl.id === file.id
+      );
+      if (!inProgressUpload?.processing) {
+        inProgressUpload?.set("processing", true);
+      }
+    });
+
+    this._onPreProcessComplete((file) => {
+      run(() => {
```
jjaffeux, line 111: A comment on why you  eed run here could be interesting

assets/javascripts/discourse/components/chat-composer-uploads.js
```diff
@@ -0,0 +1,141 @@
...
+    if (this.siteSettings.composer_media_optimization_image_enabled) {
+      this._useUploadPlugin(UppyMediaOptimization, {
+        optimizeFn: (data, opts) =>
+          this.mediaOptimizationWorker.optimizeImage(data, opts),
+        runParallel: !this.site.isMobileDevice,
+      });
+    }
+
+    this._onPreProcessProgress((file) => {
+      const inProgressUpload = this.inProgressUploads.find(
```
jjaffeux, line 102: findBy

assets/stylesheets/common/chat-composer-upload.scss
```diff
@@ -0,0 +1,71 @@
...
+    width: 50px;
+    height: 100%;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    margin: 0 1em 0 0;
+    border-radius: 8px;
+
+    .d-icon {
+      font-size: 3.2rem;
```
jjaffeux, line 22: we should use a var value here

jjaffeux (approved):

martin-brennan:
Thanks for the review @jjaffeux , hopefully will address this today :+1:

assets/javascripts/discourse/components/chat-composer.js
martin-brennan: Yes good catch, will fix!

assets/javascripts/discourse/components/chat-composer.js
martin-brennan: It's passed in from chat-live-pane:

https://github.com/discourse/discourse-chat/blob/e6aab40cb0124097faae65dd5475699acf96be3d/assets/javascripts/discourse/templates/components/chat-live-pane.hbs#L127-L127

assets/javascripts/discourse/components/chat-composer-uploads.js
martin-brennan: Yeah good point :+1:

assets/javascripts/discourse/components/chat-composer-uploads.js
martin-brennan: Hmm it might not be necessary, not sure why it is added, will remove.

assets/javascripts/discourse/components/chat-composer.js
martin-brennan: This is an event in UppyUploadMixin in core, it gets fired every time the `inProgressUploads` changes (so when one is added, successful, or removed). Maybe there is a clearer event name I can use, like `upload-status` or something? The alternative is having to hook into all the add/remove/success events from uppy directly, whereas for this composer we only care about the current state of inProgressUploads.

martin-brennan:
@jjaffeux Fixed all the minor things, and left a couple of replies :)

jjaffeux:
> @jjaffeux Fixed all the minor things, and left a couple of replies :)

Great thanks, its good to me 👍

PR #765, opened 2022-04-08
UI: implements a progress bar

jjaffeux:
Note this component is mostly all copied from the experimental discourse loading slider. Im not including test as I hope the following will happen rather soon:

- integrate loading slider in discourse-core
- share implementation of global progress bar loader and specific progress bar component in core
- remove this component


assets/javascripts/discourse/components/d-progress-bar.js
```diff
@@ -0,0 +1,99 @@
...
+
+    if (this.isLoading) {
+      this.start();
+    } else {
+      this.end();
+    }
+  },
+
+  get container() {
+    return document.getElementById(this.key);
```
pmusaraj, line 43: I am not seeing where `key` is set. It's probably me, but it seems odd.

assets/stylesheets/common/d-progress-bar.scss
```diff
@@ -0,0 +1,56 @@
...
+    transition: width var(--still-loading-duration) linear;
+    width: var(--still-loading-width);
+  }
+
+  &.done .d-progress-bar {
+    transition: width var(--done-duration) ease-out;
+    width: 100%;
+  }
+
+  body.discourse-hub-webview & {
```
pmusaraj, line 48: This is probably not needed. Discourse-loading-slider adds this class to the body, so this would have no effect here. (And if the progress bar is placed anywhere other than at the very top of the browser, it's not necessary.)

pmusaraj (approved):
I would also add comments at the top of the component and the stylesheet, noting that this is temporarily in chat, given it's going to be in core soon.

assets/javascripts/discourse/components/d-progress-bar.js
jjaffeux: It's when you defined the component, to have a uniq key if you have multiple bars

assets/stylesheets/common/d-progress-bar.scss
jjaffeux: Yes I saw this, mostly I find it easier to keep most of the stuff when copy/pasting stuff, it makes it easier to do update later, but I can remove it if you want

assets/javascripts/discourse/components/d-progress-bar.js
jjaffeux: {{d-progress-bar key="cats-loader" isLoading=(not isDone)}}

PR #769, opened 2022-04-11
FIX: makes direct_messages an index route

jjaffeux:
It makes more sense to have usernames as a query string than a URL param and it avoids issues with format and dots.


ZogStriP (approved):
👍

PR #771, opened 2022-04-11
DEV: sendMessage doesn’t need channel as param

jjaffeux:


jordanvidrine (approved):

pmusaraj (approved):
Do you think we can add a test to prevent regressions here?

jjaffeux:
> Do you think we can add a test to prevent regressions here?

Well it's always possible, but what we really need is a proper API, I will do it as part of the API to add buttons, and then I can add tests for this API.

PR #774, opened 2022-04-13
FIX: fixes a regression preventing indicator to show

jjaffeux:
This commit also improves tests.


tgxworld:
Do we need a higher level acceptance test here? I see changes have been made to the component test but it doesn't address the fix where the wrong argument was being passed to the component.

CvX (approved):

jjaffeux:
yes I'm mixed on this

> Do we need a higher level acceptance test here? I see changes have been made to the component test but it doesn't address the fix where the wrong argument was being passed to the component.

Yes I'm actually trying to write one, there's nothing on the subject yet, the feature has been added with no kind of acceptance test.

PR #787, opened 2022-04-19
FIX: Default channel setting not working

markvanlan:
This commit https://github.com/discourse/discourse-chat/commit/b8c5f7397c15126baed4363d7e19264c25ae56e9 introduced the site setting for a default channel. The issue is that the setting is a string in the JS, rather than an int (so that it can be left blank), and we were not casting the value to an integer when comparing.


martin-brennan (approved):
Ooh nice catch, thanks!

PR #791, opened 2022-04-22
FIX: Use @service router to fix chat quoting on mobile

martin-brennan:
In 6445af48bb74b6312f863b1001bdc46546b1e34b I moved some things
from chat-live-pane into a new chat-selection-manager component.
However I did not realise I made a breaking change when I used
the class syntax for the Component; in this syntax instead of
doing `router: service()` we need to do `@service router`. This
was not picked up by the mobile quoting test because it was commented/
skipped back in 8794f358389b366fbe25b5ff5eb90ad31528251f because
the mobile menu was added but we did not have a way to test the
touch support.

This commit fixes the router issue and also adds a way to test the
long press on mobile by triggering the long press on a message
immediately instead of 500ms later in testing mode. This must be triggered
immediately because there is no concept of "long press" there, the Ember
`tap` test helper does send the touchstart/touchend events correctly but immediately
see https://github.com/emberjs/ember-test-helpers/blob/master/API.md#tap

### DIFF
--- a/assets/javascripts/discourse/components/chat-message.js
+++ b/assets/javascripts/discourse/components/chat-message.js
@@ -1,4 +1,5 @@
 import bootbox from "bootbox";
+import { isTesting } from "discourse-common/config/environment";
 import Component from "@ember/component";
 import I18n from "I18n";
 import getURL from "discourse-common/lib/get-url";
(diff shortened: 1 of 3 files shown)


tgxworld (approved):

tgxworld:
Sorry do we have a test that catches the chat quoting regression?

martin-brennan:
> Sorry do we have a test that catches the chat quoting regression?

@tgxworld yes that was the one that previously was `skip`ped.

tgxworld:
@martin-brennan Oops sorry didn't notice that :) Changes look good.

PR #793, opened 2022-04-22
UX: display staged message when creating channel 

fzngagan:


CvX (approved):

jjaffeux:
We will have to change things in the near future, but Im fine with this solution as first improvement step :+1: thanks

PR #794, opened 2022-04-22
FIX: correctly acknowledge for deletion in unread_counts

jjaffeux:
Prior to this fix, if deleted message was the last message of a channel it would forever stay in unread state.

Note that this commit won't fix client side not updating unread state after without a refresh after last message has been deleted. But refresh or navigating to channel will now work where it wouldn't have before.

This commit also adds extensive specs for unread_counts.

### DIFF
--- a/spec/lib/chat_channel_fetcher_spec.rb
+++ b/spec/lib/chat_channel_fetcher_spec.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-
 describe DiscourseChat::ChatChannelFetcher do
   describe ".unread_counts" do
     fab!(:user_1) { Fabricate(:user) }
@@ -27,7 +26,7 @@ describe DiscourseChat::ChatChannelFetcher do
       context "has no unread messages" do
         it "returns the correct count" do
           unread_counts = subject.unread_counts([chat_channel], user_1)
-          expect(unread_counts[chat_channel.id]).to be_nil
+          expect(unread_counts[chat_channel.id]).to eq(0)
         end
       end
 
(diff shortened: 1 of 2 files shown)


spec/lib/chat_channel_fetcher_spec.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+require "rails_helper"
```
tgxworld, line 3:
```suggestion
```

Minor note, this is no longer required going forward per https://github.com/discourse/discourse/blob/3d62e335f7590f5f96e31b2903e8dde2f991e51e/.rspec#L2

spec/lib/chat_channel_fetcher_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+        it "returns the correct count" do
+          unread_counts = subject.unread_counts([chat_channel], user_1)
+          expect(unread_counts[chat_channel.id]).to eq(2)
+        end
+      end
+
+      context "has no unread messages" do
+        it "returns the correct count" do
+          unread_counts = subject.unread_counts([chat_channel], user_1)
+          expect(unread_counts[chat_channel.id]).to be_nil
```
tgxworld, line 31: Hmm the return value here feels odd IMO because `nil` isn't a count. Personally, I would prefer we return `0` instead of `nil`.

tgxworld (approved):
Some comments but mostly not related to the bug being addressed here. Code looks good :+1:

spec/lib/chat_channel_fetcher_spec.rb
jjaffeux: Yes I do agree it came as a surprise

jjaffeux: How do you feel about this change ?

[unreads_count should always return an integer](https://github.com/discourse/discourse-chat/pull/794/commits/b21233702c2d96ef9e533969715effe505ec6145)

ZogStriP (approved):

PR #797, opened 2022-04-26
FIX: Resize images within oneboxes

nattsw:
We never really resized images within oneboxes, so here's an attempt:

Reddit:
Screenshot 2022-04-26 at 1 53 54 PM

Existing collapser + Twitter:
Screenshot 2022-04-26 at 1 47 20 PM

Keeps max width without overflowing:
Screenshot 2022-04-26 at 2 29 29 PM

Tested in chrome and ff

### DIFF
--- a/assets/stylesheets/common/chat-message-images.scss
+++ b/assets/stylesheets/common/chat-message-images.scss
@@ -1,0 +1,30 @@
+$max_image_height: 150px;
+
+aside.onebox {
+  width: max-content;
+  max-width: -moz-fit-content;
+  max-width: fit-content;
+}
+
+// append selectors to set images to a
+// max height of $max_image_height
+.chat-message-collapser .onebox img:not(.ytp-thumbnail-image),
+.chat-message-collapser img.onebox,
+.chat-message-collapser .chat-uploads img,
+.chat-message-collapser p img,
+aside.onebox .onebox-body .aspect-image-full-size,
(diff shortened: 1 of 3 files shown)


assets/stylesheets/common/chat-message-images.scss
```diff
@@ -0,0 +1,25 @@
+$max_image_height: 150px;
+
+aside.onebox {
+  width: max-content;
+}
```
nattsw, line 5: This makes sure the border is resized according to the image

assets/stylesheets/common/chat-message-images.scss
```diff
@@ -0,0 +1,30 @@
+$max_image_height: 150px;
+
+aside.onebox {
+  width: max-content;
+  max-width: -moz-fit-content;
+  max-width: fit-content;
```
nattsw, line 6: https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content

assets/stylesheets/common/chat-message-images.scss
```diff
@@ -0,0 +1,30 @@
+$max_image_height: 150px;
+
+aside.onebox {
+  width: max-content;
```
nattsw, line 4: Makes it the onebox border snug against the image

martin-brennan (approved):
Nice :)

PR #798, opened 2022-04-26
FIX: following public channel doesn’t return channel

jjaffeux:
Prior to this change answering to an unfollowed public channel would create an error. This commit makes the result of the first step similar to allow for proper code sharing.

This commit also explains more precisely the different possible cases of sendMessage to prepare for future refactoring.

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -879,6 +879,13 @@ export default Component.extend({
     this.set("sendingloading", true);
     this._setDraftForChannel(null);
 
+    // TODO: all send message logic is due for massive refactoring
+    // This is all the possible case Im currently aware of
+    // - answering to a public channel where you are not a member yet (preview = true)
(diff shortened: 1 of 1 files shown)


assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -879,6 +879,13 @@ export default Component.extend({
     this.set("sendingloading", true);
     this._setDraftForChannel(null);
 
+    // TODO: all send message logic is due for massive refactoring
+    // This is all the possible case Im currently aware of
+    // - answering to a public channel where you are not a member yet (preview = true)
```
pmusaraj, line 884: Nitpicky, but I think we should refer to `posting` or `messaging` in this comment block (vs. `answering`).

pmusaraj (approved):

assets/javascripts/discourse/components/chat-live-pane.js
jjaffeux: funny I just said in call we should adopt common vocabulary :p  right will change

PR #800, opened 2022-04-26
FIX: Scope chat image/onebox styling to `.chat-message`

davidtaylorhq:
This styling was leaking into regular discourse posts and breaking onebox widths. Followup to 811595b8


nattsw (approved):

nattsw:
Thank you 🙏🏻

PR #804, opened 2022-04-27
FEATURE: Move chat messages to another channel

martin-brennan:
This commit introduces functionality for staff to move
messages from one chat channel to another. The original
messages are deleted and all references to the original
messages are updated to point to the new messages.

After the move we add a placeholder message to the original
channel indicating who moved the messages, how many were
moved, and where to, with a link to the first message in
the new channel.

image

image

image

Mobile (I made some changes so the buttons are centered and stacked):

image


app/controllers/chat_controller.rb
```diff
@@ -331,15 +331,39 @@ def dismiss_retention_reminder
...
+    @chat_channel = find_channel_and_check_access(params[:chat_channel_id])
 
     message_ids = params[:message_ids].map(&:to_i)
     markdown = ChatTranscriptService.new(@chat_channel, messages_or_ids: message_ids).generate_markdown
     render json: success_json.merge(markdown: markdown)
   end
 
+  def move_messages_to_channel
+    params.require(:message_ids)
+    params.require(:destination_channel_id)
```
tgxworld, line 343: Do we need to ensure `chat_channel_id` is present too?

martin-brennan: It's part of the route, so it should be there otherwise we won't hit this method :)

app/controllers/chat_controller.rb
```diff
@@ -331,15 +331,39 @@ def dismiss_retention_reminder
...
+
+    raise Discourse::InvalidAccess if !guardian.can_move_chat_messages?
+
+    @chat_channel = find_channel_and_check_access(params[:chat_channel_id])
```
tgxworld, line 347: I'm wondering if there is a need to use instance variables here.

martin-brennan: Good point, will remove.

app/controllers/chat_controller.rb
```diff
@@ -331,15 +331,39 @@ def dismiss_retention_reminder
...
+    @destination_channel = find_channel_and_check_access(params[:destination_channel_id])
+
+    message_ids = params[:message_ids].map(&:to_i)
+
+    begin
+      moved_messages = ChatMessageMoveService.new(
```
jjaffeux, line 353: We generally dont use the service terminology afaik.

Also the chat prefix is un necessary IMO as it's namespaced under DiscourseChat (I know other places do it in code currently but it will be dropped).

MoveMessageManager maybe?

Anyways feel free to keep service if you think that's the best fit here. Im just currently thinking of routes and naming in the plugin so thought I would raise this here.

app/controllers/chat_controller.rb
```diff
@@ -441,4 +465,11 @@ def create_action_whisper(topic, action)
         custom_fields: { "action_code_who" => current_user.username }
       )
   end
+
+  def find_channel_and_check_access(channel_id)
```
jjaffeux, line 469: Wonder if we shouldnt have some light service object for this Dont feel super good about this function.

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
```diff
@@ -0,0 +1,71 @@
...
+      this.set("channels", channels);
+    });
+  }
+
+  @computed("destinationChannelId")
+  get disableMoveButton() {
+    return isBlank(this.destinationChannelId);
+  }
+
+  @computed("channels.publicChannels")
```
jjaffeux, line 33: Cant we rely on chat.publicChannels instead of all this fetching/filtering logic?

assets/javascripts/discourse/components/chat-selection-manager.js
```diff
@@ -11,17 +13,27 @@ import getURL from "discourse-common/lib/get-url";
...
+  @service router;
+  @alias("currentUser.staff") showMoveMessages;
+
   @computed("selectedMessageIds.length")
   get anyMessagesSelected() {
     return this.selectedMessageIds.length > 0;
   }
 
+  @action
+  moveMessagesToChannel() {
```
jjaffeux, line 30: Not sure about the name here, sounds like we are actually moving stuff, when we are just showing Ui at that point

app/controllers/chat_controller.rb
martin-brennan: > We generally dont use the service terminology afaik.

I tried to come back with a rebuttal, but everything with `Service` in the name in core or in chat was made by me :laughing: 

> Also the chat prefix is un necessary IMO as it's namespaced under DiscourseChat

I actually don't have it namespaced. Will do that, and change to being called `DiscourseChat::MessageMover` (to echo `PostMover` in core)

jjaffeux: Or MessageMover as we have PostMover?

app/controllers/chat_controller.rb
martin-brennan: Yes that's fine, I only moved this into a function because it will be the second time I use it in this controller. Maybe we could have a class called `ChannelLookup`? Then we could have `ChannelLookup.find(id)` and it does the same thing as this function.

Not sure on the naming, any thoughts on what it might look like?

app/controllers/chat_controller.rb
martin-brennan: Yes I suggested the same above :)

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
martin-brennan: No, we cannot show the current channel in the list that you are moving messages from, that would be confusing.

assets/javascripts/discourse/components/chat-selection-manager.js
martin-brennan: Yes good point, will change to `openMoveMessageModal` :+1:

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
jjaffeux: I mean Im fine the reject part, but it could be a computed property on `chat.publicChannels` instead of all the fetching logic we have here.

```
@discourseComputed("chat.publichChannels")
avaiableChannels(publicChannels) {
  return publicChannels.reject(...);
}
```

app/controllers/chat_controller.rb
jjaffeux: I know we have ChannelFetcher already, didn't look much what it's doing at the moment, ChannelLookup sounds good to me

app/controllers/chat_controller.rb
jjaffeux: Sorry was on mobile didn't see your answer :p

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -699,6 +702,15 @@ export default Component.extend({
     }
   },
 
+  handleBulkDeleteMessage(data) {
```
pmusaraj, line 705: Bikeshed: deleting messages in bulk sounds like a pretty useful action on its own.

assets/javascripts/discourse/components/chat-selection-manager.js
```diff
@@ -11,17 +13,27 @@ import getURL from "discourse-common/lib/get-url";
...
 export default class AdminCustomizeColorsShowController extends Component {
   tagName = "";
-  @service router;
   chatChannel = null;
   selectedMessageIds = null;
   showChatQuoteSuccess = false;
   cancelSelecting = null;
 
+  @service router;
+  @alias("currentUser.staff") showMoveMessages;
```
pmusaraj, line 22: This can be a confusing use of `alias`, could we use `currentUser.staff` in the template directly?

assets/stylesheets/common/common.scss
```diff
@@ -17,6 +17,15 @@ $float-height: 530px;
   }
 }
 
+.chat-move-message-channel-chooser {
+  width: 100% !important;
```
pmusaraj, line 21: Can we avoid the `!important`?

assets/javascripts/discourse/components/chat-selection-manager.js
jjaffeux: Yes I missed this one, definitely not an alias anyways; but a reads if we keep it. I do agree just `currentUser.staff` is probably more explicit in this case.

assets/stylesheets/common/common.scss
martin-brennan: Yep will try

assets/javascripts/discourse/components/chat-selection-manager.js
martin-brennan: Okay will change to using the staff directly in the template

assets/javascripts/discourse/components/chat-live-pane.js
martin-brennan: Yes I brought that up yesterday internally about the "selected messages" actions, maybe we could have a bulk delete at some point

lib/chat_message_move_service.rb
```diff
@@ -0,0 +1,165 @@
...
+
+  def initialize(acting_user:, source_channel:, message_ids:)
+    @source_channel = source_channel
+    @acting_user = acting_user
+    @source_message_ids = message_ids
+    @source_messages = find_messages(@source_message_ids, source_channel)
+    @ordered_source_message_ids = @source_messages.map(&:id)
+  end
+
+  def move_to_channel(destination_channel)
```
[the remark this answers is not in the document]
martin-brennan: I wanted it this way because I figured that this class could eventually also be used for `move_to_topic` when we introduce that again -- one class many targets (even if eventually we refactor to have this parent class that calls out to a `MoveMessagesToChannel` and `MoveMessagesToTopic` classes

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
```diff
@@ -0,0 +1,55 @@
...
+  destinationChannelId = null;
+  selectedMessageIds = null;
+
+  @computed("destinationChannelId")
+  get disableMoveButton() {
+    return isBlank(this.destinationChannelId);
+  }
+
+  @discourseComputed("chat.publicChannels.[]")
+  availableChannels(publicChannels) {
```
tgxworld, line 23: Just curious, why do we define getters for the other computed properties but not this?

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
```diff
@@ -0,0 +1,55 @@
...
+    return publicChannels.reject((chan) => chan.id === this.sourceChannel.id);
+  }
+
+  get selectedMessageCount() {
```
tgxworld, line 27: Hmm does this not need to be a computed property?

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
jjaffeux: this might also cause issues on legacy: discourseComputed in native classes is not working well, use computed please

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
jjaffeux: You can also just use reads() for this

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
```diff
@@ -0,0 +1,55 @@
...
+      {
+        method: "PUT",
+        data: {
+          message_ids: this.selectedMessageIds,
+          destination_channel_id: this.destinationChannelId,
+        },
+      }
+    )
+      .then((response) => {
+        return this.router.transitionTo(
```
jjaffeux, line 44: I think you can return a promise from this with .promise I think

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
```diff
@@ -0,0 +1,55 @@
...
+  selectedMessageIds = null;
+
+  @computed("destinationChannelId")
+  get disableMoveButton() {
+    return isBlank(this.destinationChannelId);
+  }
+
+  @discourseComputed("chat.publicChannels.[]")
+  availableChannels(publicChannels) {
+    return publicChannels.reject((chan) => chan.id === this.sourceChannel.id);
```
jjaffeux, line 24: you can use rejectBy

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  get selectedMessageCount() {
+    return this.selectedMessageIds.length;
+  }
+
+  @action
+  moveMessages() {
+    return ajax(
```
jjaffeux, line 33: Just curious what happens if I call the same endpoint with same params multiple times?

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
martin-brennan: I don't think it's necessary to make it computed if the value will not change for the lifetime of the component, I will change to reads() though.

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
martin-brennan: Okay, but I don't think it really needs any return value since nothing is waiting on it. Will just remove return.

martin-brennan:
@tgxworld and @jjaffeux addressed second round of comments.

assets/javascripts/discourse/components/chat-message-move-to-channel-modal-inner.js
jjaffeux: Ok so this is what I thought. This is kinda unexpected for a PUT as it should more or less be idempotent. I don't want to go into a lengthy REST debate, but if you have an idea of something else than PUT here, that might be good

martin-brennan: I mean, I think it fits PUT as we are updating stuff. We don't exactly strictly adhere to REST in most places. I don't really care though happy to change to POST if you want.

martin-brennan: Also it is idempotent, you can't move the same messages twice. I intentionally throw an error too, it's not just an unhandled error: 

https://github.com/discourse/discourse-chat/blob/1d1707fd11fcf84d451f2fe80745b178b076c6d9/lib/message_mover.rb#L36-L38

jjaffeux: I just think maybe it should just not error if these message ids are already associated with this channel. But that will make code more complicated, I will live with this.

jjaffeux:
Minor UI issue on float:

Screenshot 2022-04-29 at 09 57 43

jjaffeux:
I had an error when moving a message from a channel to another:
Screenshot 2022-04-29 at 09 58 07
Screenshot 2022-04-29 at 09 58 17

martin-brennan:
Comments by @jjaffeux were addressed in the latest commit and for other places in chat with https://github.com/discourse/discourse-chat/pull/821

jjaffeux (approved):

PR #812, opened 2022-04-28
FIX: ensures collapsing is working on legacy

jjaffeux:
As part of this fix was related to the use of get/`@computed` in a non native class, I have refactored the whole component:

- native class
- tagless


CvX:
A reminder to use \`\` when writing about decorators, so that you don't mention random people on github 😝

CvX (approved):

PR #818, opened 2022-04-28
FIX: Don't hide the new messages indicator beside the channel header.

romanrizzi:
Previously, seeing new messages in a channel was unnatural when there was a large number on unread messages because we displayed the newest ones first, meaning you'll have to scroll back to see the conversation from the last you
read.

This change improves our scrolling logic by loading new messages from where you left the last time you visited the channel. Also, it'll load a couple of older messages to avoid leaving the "New Messages" indicator hidden beside the
channel header. We had to refactor the messages endpoint to make the direction you're scrolling explicit, which let us create the concept of landing in the middle of the conversation, where you can generally scroll both ways.

**Before:**
Screen Shot 2022-04-28 at 18 16 28
**After:**
Screen Shot 2022-04-28 at 18 14 43

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -31,8 +31,9 @@ const STICKY_SCROLL_LENIENCE = 4;
 const READ_INTERVAL = 1000;
 const PAGE_SIZE = 50;
 
-const PAST = "Past";
-const FUTURE = "Future";
+const PAST = "past";
+const FUTURE = "future";
+const BOTH = "both";
 
 let _chatMessageDecorators = [];
 
@@ -175,11 +176,17 @@ export default Component.extend({
 
     return this.chat.loadCookFunction(this.site.categories).then((cook) => {
       this.set("cook", cook);
+
       const findArgs = {
         channelId,
         targetMessageId: this.targetMessageId,
         pageSize: PAGE_SIZE,
       };
+
+      if (!this.targetMessageId) {
+        this._includeInmediateHistory(findArgs);
+      }
+
       return this.store
         .findAll("chat-message", findArgs)
         .then((messages) => {
@@ -233,8 +240,9 @@ export default Component.extend({
     const findArgs = {
       channelId: this.chatChannel.id,
       pageSize: PAGE_SIZE,
+      direction,
+      messageId,
     };
-    findArgs[`${loadingPast ? "before" : "after"}MessageId`] = messageId;
     const channelId = this.chatChannel.id;
 
     return this.store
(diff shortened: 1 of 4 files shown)


jjaffeux:
Can you add some visuals to the PR please @romanrizzi ?

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -410,6 +418,23 @@ export default Component.extend({
     return message.id || `staged-${message.stagedId}`;
   },
 
+  _getLastReadId() {
+    return this.currentUser.chat_channel_tracking_state[this.chatChannel.id]
+      ?.chat_message_id;
+  },
+
+  _includeInmediateHistory(findArgs) {
```
pmusaraj, line 426:
```suggestion
  _includeImmediateHistory(findArgs) {
```

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -410,6 +418,23 @@ export default Component.extend({
...
+    let lastReadId = this._getLastReadId();
```
pmusaraj, line 427:
```suggestion
    const lastReadId = this._getLastReadId();
```

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -418,22 +443,29 @@ export default Component.extend({
...
       this.messages.findIndex((m) => m.id === lastReadId) || 0;
-    const newestUnreadMessage = this.messages[indexOfLastReadyMessage + 1];
+    let newestUnreadScrollTarget = indexOfLastReadyMessage + 1;
+    let newestUnreadMessage = this.messages[newestUnreadScrollTarget];
 
     if (newestUnreadMessage) {
       newestUnreadMessage.set("newestMessage", true);
-      // We have the last read message from lookup, but now we need the index of the message,
-      // so that we can scroll to the message directly after it.
+      const previousMessagesToDisplay = 3;
```
pmusaraj, line 459: I'd be fine with reducing this to 2. To avoid the "new message" indicator being below the fold in the off chance that these 3 messages are long. (There's only so much we can do here without major goldplating, but 2 is a tiny bit less risky than 3.)

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -410,6 +418,23 @@ export default Component.extend({
...
+      ?.chat_message_id;
+  },
+
+  _includeInmediateHistory(findArgs) {
+    let lastReadId = this._getLastReadId();
+
+    if (lastReadId) {
+      // We are fetching the last read message, the two previous ones,
+      // and the next 47 to complete a page.
+      const offset = 47;
```
pmusaraj, line 432: Maybe we should use `50 - previousMessagesToDisplay` here. Actually, `50` should be moved to a page size constant as well since it is a param.

pmusaraj (approved):
Looks good!

jjaffeux (approved):

pmusaraj (approved):

PR #821, opened 2022-04-29
FIX: Do not assume name exists for channel

martin-brennan:
There were many places in the code where we (mainly I)
assumed that the channel would always have a name. This
is not the case and we should always be using the
`channel.title(user)` method to determine the channel
title.


lib/chat_transcript_service.rb
```diff
@@ -24,12 +24,14 @@ class ChatTranscriptBBCode
 
     def initialize(
       channel: nil,
+      acting_user: nil,
```
jjaffeux, line 27: I don't think you wanted to have all the acting_user stuff in this PR

martin-brennan: @jjaffeux yes I do...how else can I pass the user to the `.title(user)` method on the channel?

jjaffeux (approved):

PR #822, opened 2022-04-29
DEV: Introduce PR template

nattsw:
To minimize UI bugs /65598

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,24 @@
+<!-- Checklist for UI / stylesheet changes, delete if not applicable -->
+
+- [ ] chat-scoped -- core unaffected
+
+### View mode
+
+- [ ] docked (windowed/drawer)
+- [ ] isolated (full-screen)
+
+### Browsers
+
+- [ ] safari
+- [ ] chrome
+- [ ] firefox
+
+### Device
+
+- [ ] desktop – wide screen
+- [ ] mobile – `?mobile_view=1`
+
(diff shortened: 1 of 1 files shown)


jjaffeux:
It's ironic to have a flakey on this :D

jjaffeux:
Do we actually had a lot of cases of specific bugs in firefox/safari ?

jjaffeux:
Mostly I just think:

```
### Browsers

- [ ] safari
- [ ] chrome
- [ ] firefox

### Device

- [ ] desktop – wide screen
- [ ] mobile – `?mobile_view=1
```

Is general recommendation, not specific to chat

jjaffeux:
And even if I added it ember-cli/legacy is also not specific to chat

jjaffeux (approved):

jjaffeux:
Im liking the list but feels like a lot of it is part of general Discourse work, maybe we should backport part of it in discourse/discourse ?

It's ok to have an extended list here.  Maybe we also add the test part of discourse/discourse in this template too?

nattsw:
> a lot of cases of specific bugs in firefox/safari ?

I believe the one that I know of for Safari is related to [scrolling](https://github.com/discourse/discourse-chat/pull/472), and [copy/pasting](https://github.com/discourse/discourse-chat/pull/718).

For Firefox, my earlier style for `fit-content` required something moz-specific which I just randomly decided to test and it indeed failed.

nattsw:
Thank you

PR #824, opened 2022-04-29
FIX: loads populate only on development env

jjaffeux:


davidtaylorhq (approved):

lib/tasks/chat.rake
```diff
@@ -1,21 +1,23 @@
 # frozen_string_literal: true
 
-chat_task = Rake::Task['dev:populate']
-chat_task.enhance {
-  SiteSetting.chat_enabled = true
-  DiscourseDev::PublicChannel.populate!
-  DiscourseDev::DirectChannel.populate!
-  DiscourseDev::Message.populate!
-}
+if ENV['ALLOW_DEV_POPULATE'] == '1'
```
davidtaylorhq, line 3:
```suggestion
if Discourse.allow_dev_populate?
```

https://github.com/discourse/discourse/blob/14f61c57847b6d0a861653e2f479947ad9b68962/lib/discourse.rb#L1026-L1028

plugin.rb
```diff
@@ -122,6 +119,12 @@ def self.allowed_group_ids
   load File.expand_path('../app/jobs/scheduled/update_user_counts_for_chat_channels.rb', __FILE__)
   load File.expand_path('../app/services/chat_publisher.rb', __FILE__)
 
+  if ENV['ALLOW_DEV_POPULATE'] == '1'
```
davidtaylorhq, line 122:
```suggestion
if Discourse.allow_dev_populate?
```

https://github.com/discourse/discourse/blob/14f61c57847b6d0a861653e2f479947ad9b68962/lib/discourse.rb#L1026-L1028

pmusaraj (approved):

PR #827, opened 2022-05-02
FIX: Don't hide the new messages indicator beside the channel header

romanrizzi:
Reintroduces discourse/discourse-chat#818, which was reverted on discourse/discourse-chat#826.


jjaffeux:
What is the difference so far?

romanrizzi:
I'm still trying to reproduce the issue locally 😞

jjaffeux:
> I'm still trying to reproduce the issue locally 😞

At the very least I think you need two browser windows with lots of messages (rake chat:populate:messages) and then type a message

jjaffeux (approved):

jjaffeux:
Ok let's try this

PR #830, opened 2022-05-04
FEATURE: Chat message bookmarks

martin-brennan:
**Note: Do not merge until the complete polymorphic bookmark move has been made in core.**

Registers the `ChatMessage` bookmarkable using the core API and introduces the UI to bookmark individual chat messages, along with reminder notification functionality for chat message bookmarks.

**Also Note: The design of this is not complete yet, but the rest of the PR is still fine to review!**. I will add UI/JS tests once the design is nailed down as things may move around.

Chat bookmarks in user list:

image

Bookmarks within the channel:

image
image


jjaffeux:
Just saying we can merge with incomplete design and iterate later, dont block yourself on this martin.

lib/chat_channel_fetcher.rb
```diff
@@ -16,6 +16,43 @@ def self.structured(guardian)
     }
   end
 
+  def self.all_secured_channel_ids(guardian)
```
jjaffeux, line 19: This should have tests I think

assets/javascripts/discourse/templates/components/chat-message.hbs
```diff
@@ -114,6 +114,14 @@
 
               {{format-chat-date message details}}
 
+              {{#if bookmark}}
```
jjaffeux, line 117: I imagine we have the same kind of logic in core, this could be a {{bookmark-icon bookmark=bookmark}} component, which would also be easy to test

jjaffeux (approved):

assets/javascripts/discourse/components/chat-message.js
```diff
@@ -764,6 +792,35 @@ export default Component.extend({
...
+      this.bookmark ||
+        Bookmark.create({
+          bookmarkable_type: "ChatMessage",
+          bookmarkable_id: this.message.id,
+          user_id: this.currentUser.id,
+        }),
+      {
+        onAfterSave: (savedData) => {
+          const bookmark = Bookmark.create(savedData);
+          this.set("message.bookmark", savedData);
```
jjaffeux, line 807: it's possible you currently can't do differently due to chat design, but this bookmark and message.bookmark at the same time seems like a smell to me.

jjaffeux (approved):

jjaffeux:
Overall looks super good 👏👏👏

assets/javascripts/discourse/components/chat-message.js
martin-brennan: I just did it this way because I didn't want to override `message.bookmark` which comes down from the serializer, but thinking about it more I can probably just do that without problems :+1:

assets/javascripts/discourse/templates/components/chat-message.hbs
martin-brennan: Actually in core all of the icons and this kind of logic for bookmarks are set in widgets. However this is a good idea, I will come back to it after merging the polymorphic bookmarks in core.

lib/chat_channel_fetcher.rb
martin-brennan: It is tested in a way via the bookmarkable query tests, but yes I will add dedicated tests for this section because it is important.

martin-brennan:
> Just saying we can merge with incomplete design and iterate later, dont block yourself on this martin.

Thanks @jjaffeux I will make your changes and I will hopefully merge this week after https://github.com/discourse/discourse/pull/16729 in core

lib/chat_channel_fetcher.rb
martin-brennan: Just checked and this channel fetcher is pretty badly tested :( or maybe mostly indirectly tested. Will add a todo to myself to add more.

jjaffeux: Yes fetcher is very low on tests, don't need to go aggressive on this, but at least when we add a new method it should be tested directly and not indirectly via another test.

jjaffeux (approved):

PR #837, opened 2022-05-05
DEV: Allow category group mods to moderate chat channels

martin-brennan:
This allows category group moderators to moderate chat
channels via the `guardian.can_moderate_chat?` extension.
Practically this now allows category moderators to do
the following on category channels:

* Move messages to another channel
* Delete/restore/and see deleted chat messages

It also fixes some inconsistency with `can_delete_self` and
`can_delete_others`, which we had guardian methods for
and meta/details hash to attach the result to, but we were
setting a hardcoded result in the JS code.

Doing this necessitated introducing a dynamic `generateChatView`
function for the chat fixtures, because what is in these meta properties
is dependent on the current user. Tests were updated to use this
new function instead of the static `chatView`.

### DIFF
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -344,10 +344,10 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
     params.require(:message_ids)
     params.require(:destination_channel_id)
 
-    raise Discourse::InvalidAccess if !guardian.can_move_chat_messages?
-
     chat_channel = DiscourseChat::ChatChannelFetcher.find_with_access_check(params[:chat_channel_id], guardian)
     destination_channel = DiscourseChat::ChatChannelFetcher.find_with_access_check(params[:destination_channel_id], guardian)
+
+    raise Discourse::InvalidAccess if !guardian.can_move_chat_messages?(chat_channel)
 
     message_ids = params[:message_ids].map(&:to_i)
 
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -345,13 +345,11 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
     params.require(:destination_channel_id)
 
     chat_channel = DiscourseChat::ChatChannelFetcher.find_with_access_check(params[:chat_channel_id], guardian)
+    raise Discourse::InvalidAccess if !guardian.can_move_chat_messages?(chat_channel)
     destination_channel = DiscourseChat::ChatChannelFetcher.find_with_access_check(params[:destination_channel_id], guardian)
 
-    raise Discourse::InvalidAccess if !guardian.can_move_chat_messages?(chat_channel)
-
-    message_ids = params[:message_ids].map(&:to_i)
-
     begin
+      message_ids = params[:message_ids].map(&:to_i)
       moved_messages = DiscourseChat::MessageMover.new(
         acting_user: current_user, source_channel: chat_channel, message_ids: message_ids
       ).move_to_channel(destination_channel)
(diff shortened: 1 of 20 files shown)


martin-brennan:
Noting that because `chatView` is used by the fixtures and now `can_delete_self`, `can_delete_others`, and `can_moderate` are all stored on the meta property there, we need some way to easily switch these booleans based on the logged in `needs.user`...might be able to just used `loggedInUser()` and add a function that can generate a `chatView` for fixtures based on the current user and other opts...

app/controllers/chat_controller.rb
```diff
@@ -344,11 +344,11 @@ def move_messages_to_channel
     params.require(:message_ids)
     params.require(:destination_channel_id)
 
-    raise Discourse::InvalidAccess if !guardian.can_move_chat_messages?
-
     chat_channel = DiscourseChat::ChatChannelFetcher.find_with_access_check(params[:chat_channel_id], guardian)
     destination_channel = DiscourseChat::ChatChannelFetcher.find_with_access_check(params[:destination_channel_id], guardian)
 
+    raise Discourse::InvalidAccess if !guardian.can_move_chat_messages?(chat_channel)
```
tgxworld, line 350: I would move this just right after `chat_channel = ....`. That way, we don't fetch `destination_channel` for nothing if the guardian check fails.

spec/lib/guardian_extensions_spec.rb
```diff
@@ -147,6 +147,18 @@
           expect(staff_guardian.can_moderate_chat?(channel.chatable)).to eq(true)
           expect(guardian.can_moderate_chat?(channel.chatable)).to eq(false)
         end
+
+        context "when enable_category_group_moderation is true" do
+          before { SiteSetting.enable_category_group_moderation = true }
+
+          it "returns true if the regular user is part of the reviewable_by_group for the category" do
+            mods = Fabricate(:group)
+            GroupUser.create(user: user, group: mods)
```
tgxworld, line 156: Perhaps we can just add the user to the group directly? If I am not wrong, we run some special callbacks when adding a user to a group so just manually creating the `GroupUser` record might lead to surprises .

```suggestion
            mods.add(user)
```

tgxworld: What do you think about testing this as part of the request spec for chat_controller which I feel will help us to achieve higher coverage since the controller and the guardian is being tested at the same time.

test/javascripts/acceptance/chat-flagging-test.js
```diff
@@ -29,6 +29,7 @@ const chatSettled = async () => {
 };
 
 acceptance("Discourse Chat - Flagging test", function (needs) {
+  let defaultChatView;
```
tgxworld, line 32: Should we reset this variable after each test?

tgxworld (approved):
Some comments but nothing blocking :) Changes look good to me :+1:

test/javascripts/acceptance/chat-flagging-test.js
martin-brennan: I think it's fine as is, it's kind of a weird reason we use this (because we are accessing the chat message IDs from the view) in this test, and it's only used within this acceptance function. The hoops we have to jump through because we don't have let/context in JS! :laughing:

spec/lib/guardian_extensions_spec.rb
martin-brennan: Yep definitely to both of these, we already have a `can_moderate` test for the controller that I can add this to :+1:

test/javascripts/acceptance/chat-flagging-test.js
tgxworld: Ooo icic. I wasn't sure if `generateChatView` would be returning different ids each time. All good here :)

PR #838, opened 2022-05-05
FEATURE: Chat notification emails

romanrizzi:
Reintroduces feature originally added in #759 and later reverted in #803.  We didn't revert migrations, so it's not necessary to add them in this commit.

his PR is similar to https://github.com/discourse/discourse-chat/pull/759 but doesn't use a new table to track which mentions and private messages need to be processed using a new table.

### Light mode:

Screen Shot 2022-05-20 at 12 00 00

### Dark mode:

Screen Shot 2022-05-20 at 11 59 35

### DIFF
--- a/app/models/chat_channel.rb
+++ b/app/models/chat_channel.rb
@@ -122,6 +122,21 @@ class ChatChannel < ActiveRecord::Base
     end
   end
 
+  def email_title(user)
+    if direct_message_channel?
+      usernames = (chatable.users.map(&:username) - [user.username]).map { |username| "@#{username}" }
+      if usernames.count > 5
+        return I18n.t("chat.channel.title_for_large_group_dm", users: usernames.take(5).join(", "), count: usernames.count - 5)
+      end
+
+      if usernames.count > 1
+        return usernames.join(", ")
+      end
+    end
+
+    title(user)
+  end
+
   def change_status(acting_user, target_status)
     return if !ChatChannel.statuses.include?(target_status.to_s)
     return if !Guardian.new(acting_user).can_change_channel_status?(self, target_status)
--- a/app/models/chat_channel.rb
+++ b/app/models/chat_channel.rb
@@ -122,21 +122,6 @@ class ChatChannel < ActiveRecord::Base
     end
   end
 
-  def email_title(user)
-    if direct_message_channel?
-      usernames = (chatable.users.map(&:username) - [user.username]).map { |username| "@#{username}" }
-      if usernames.count > 5
-        return I18n.t("chat.channel.title_for_large_group_dm", users: usernames.take(5).join(", "), count: usernames.count - 5)
-      end
-
-      if usernames.count > 1
-        return usernames.join(", ")
-      end
-    end
-
-    title(user)
-  end
-
   def change_status(acting_user, target_status)
     return if !ChatChannel.statuses.include?(target_status.to_s)
     return if !Guardian.new(acting_user).can_change_channel_status?(self, target_status)
--- a/assets/javascripts/discourse/controllers/preferences-chat.js
+++ b/assets/javascripts/discourse/controllers/preferences-chat.js
@@ -12,9 +12,20 @@ const chatAttrs = [
   "only_chat_push_notifications",
   "ignore_channel_wide_mention",
   "chat_sound",
+  "chat_email_frequency",
+];
+
+const emailFrequencyOptions = [
+  { name: I18n.t(`chat.email_frequency.never`), value: "never" },
+  { name: I18n.t(`chat.email_frequency.when_away`), value: "when_away" },
 ];
 
 export default Controller.extend({
+  init() {
+    this._super(...arguments);
+    this.set("emailFrequencyOptions", emailFrequencyOptions);
+  },
+
   @discourseComputed
   chatSounds() {
     return Object.keys(CHAT_SOUNDS).map((value) => {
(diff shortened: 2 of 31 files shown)


assets/javascripts/discourse/controllers/preferences-chat.js
```diff
@@ -12,9 +12,20 @@ const chatAttrs = [
   "only_chat_push_notifications",
   "ignore_channel_wide_mention",
   "chat_sound",
+  "chat_email_frequency",
+];
+
+const emailFrequencyOptions = [
```
jjaffeux, line 18:
```suggestion
const EMAIL_FREQUENCY_OPTIONS = [
```

app/controllers/chat_controller.rb
```diff
@@ -143,7 +143,26 @@ def edit_message
 
   def update_user_last_read
     set_channel_and_chatable
-    set_user_last_read
+    UserChatChannelMembership
+      .where(user: current_user, chat_channel: @chat_channel)
+      .update_all(last_read_message_id: params[:message_id])
```
ZogStriP, line 148: I know this is 99% of the time supposed to update only 1 row, but there are a couple of addition checks we usually add to avoid anything terrible from happening.

The first one is to ensure we're updating the rows that needs to be updated by ensuring the value we're updating is actually different from the one stored in the database.

The second one is to ensure we're only ever updating 1 row using `LIMIT 1`.

Not sure if we can easily write that in ActiveRecord, but in SQL it should look something like this

```sql
  UPDATE user_chat_channel_membership
     SET last_read_message_id = @params[:message_id]
   WHERE user_id = @current_user.id
     AND chat_channel_id = @chat_channel.id
     AND last_read_message_id <> @params[:message_id]
   LIMIT 1
```

app/controllers/chat_controller.rb
```diff
@@ -143,7 +143,26 @@ def edit_message
...
+
+    chat_mentions = ChatMention
+      .joins(:notification)
+      .joins(:chat_message)
+      .where(user: current_user)
+      .where(chat_message: { chat_channel_id: @chat_channel.id })
+      .where(notification: { read: false })
+
+    chat_mentions.each do |chat_mention|
+      chat_mention.notification.update(read: true)
```
ZogStriP, line 158: Is this going to do 1 request per mention? 🤔

app/models/chat_channel.rb
```diff
@@ -122,6 +122,21 @@ def title_from_chatable
     end
   end
 
+  def email_title(user)
+    if direct_message_channel?
+      usernames = (chatable.users.map(&:username) - [user.username]).map { |username| "@#{username}" }
```
ZogStriP, line 127: Should we sort the list of users like we do in "normal" email notification?

config/locales/client.en.yml
```diff
@@ -70,6 +70,11 @@ en:
         public_available: "There are public channels available for you to follow."
         start_direct_message: "You can also start a personal chat with one or more users."
         title: "Get started by joining a channel"
+      email_frequency:
+        description: "We'll only email you if we haven't seen you in the last 10 minutes."
```
ZogStriP, line 74: Doesn't the job run every 15 minutes?

```suggestion
        description: "We'll only email you if we haven't seen you in the last 15 minutes."
```

lib/chat_mailer.rb
```diff
@@ -0,0 +1,41 @@
...
+      user.user_option.update(last_emailed_for_chat: Time.now)
+    end
+  end
+
+  def users
+    User
+      .distinct
+      .joins(:user_option)
+      .where(user_options: { chat_enabled: true, chat_email_frequency: UserOption.chat_email_frequencies[:when_away] })
+      .where('user_options.last_emailed_for_chat IS NULL OR user_options.last_emailed_for_chat < ?', 5.minutes.ago)
```
ZogStriP, line 23: Why 5 minutes ago?

app/models/chat_channel.rb
romanrizzi: We should use the existing `title` method and sort by username there instead. It was recently improved in #859.

app/controllers/chat_controller.rb
romanrizzi: This logic has a couple of bugs. I fixed them in #862 and will remove this change from here.

app/controllers/chat_controller.rb
romanrizzi: Fixed in #862

lib/chat_mailer.rb
romanrizzi: We should rely on `user#last_seen_at` here. No need to keep this field.

config/locales/client.en.yml
romanrizzi: This will be tied to the `user#last_seen_at` and the `email_time_window_mins` setting (default to 10 minutes). Maybe we could run the job every 10 minutes instead 15.

romanrizzi:
We're not including the unsubscribe link for now. We need to add a core API for that.

config/locales/server.en.yml
```diff
@@ -114,3 +114,22 @@ en:
...
+  user_notifications:
+    chat_summary:
+      deleted_user: "Deleted user"
+      description:
+        one: "You have a new chat message"
+        other: "You have new chat messages"
+      from: "%{site_name}"
+      subject:
+        one: "[%{email_prefix}] New chat message"
+        other: "[%{email_prefix}] New chat messages"
```
pmusaraj, line 127:
```suggestion
        one: "%{email_prefix} new chat message"
        other: "%{email_prefix} new chat messages"
```

pmusaraj (approved):
Looks great!

We should wait for Monday next week to merge and maybe for @jjaffeux to have a chance to review.

config/locales/server.en.yml
romanrizzi: Wouldn't the subject look weird with this change? e.g. `Discourse Meta new chat messages` instead of `[Discourse Meta] New chat messages`

PR #841, opened 2022-05-08
FIX: ensures mentions are correctly highlighted

jjaffeux:
Prior to this fix mentions were highlighted when inserted the first time,  but if the message was changed (cooked) without destroying the message component, then it wouldn't get decorated again.

This commit also uses the decorating API to apply the classes.

### DIFF
--- a/assets/javascripts/discourse/components/chat-message.js
+++ b/assets/javascripts/discourse/components/chat-message.js
@@ -16,9 +16,25 @@ import { inject as service } from "@ember/service";
 import { popupAjaxError } from "discourse/lib/ajax-error";
 import { prioritizeNameInUx } from "discourse/lib/settings";
 import { Promise } from "rsvp";
+import { addChatMessageDecorator } from "discourse/plugins/discourse-chat/discourse/components/chat-live-pane";
 
-const HERE = "here";
-const ALL = "all";
+const HERE = "@here";
+const ALL = "@all";
+
+addChatMessageDecorator(function (chatMessage) {
+  if (!this.currentUser) {
+    return;
+  }
+
+  const highlightable = [`@${this.currentUser.username}`, HERE, ALL];
+
+  chatMessage.querySelectorAll(".mention").forEach((node) => {
+    const mention = node.textContent.trim();
+    if (highlightable.includes(mention)) {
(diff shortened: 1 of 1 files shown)


jjaffeux:
Screenshot 2022-05-09 at 11 14 32

Screenshot 2022-05-09 at 11 14 35


Lies.

nattsw (approved):

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -78,6 +78,25 @@ export default {
...
+      api.decorateChatMessage(function (chatMessage) {
+        if (!this.currentUser) {
+          return;
+        }
+
+        const highlightable = [
+          `@${this.currentUser.username}`,
+          "@here",
+          "@all",
+        ];
```
nattsw, line 91: Btw this means that other users won't be highlighted?

nattsw: Also noted that `here` and `all` are keywords but hardcoded here.

nattsw (commented):
Couple of comments about highlighting

assets/javascripts/discourse/initializers/chat-setup.js
jjaffeux: Yes I too find the naming a little bit weird as ".mention" will already be highlighted and ".highlighted" is basically "even more highlighting".

This is out of the scope of this PR but we could maybe refactor to have better naming here.

jjaffeux: > Also noted that `here` and `all` are keywords but hardcoded here.

what do you suggest instead?

nattsw: > as ".mention" will already be highlighted and ".highlighted" is basically "even more highlighting"

Got it.

> what do you suggest instead?

Personally I was thinking something along the lines of `MENTION_KEYWORDS = ["here", "all"]`, mostly because in some other places like slack, there's also `channel` and potentially other configurable terms/aliases.

jjaffeux: Works for me, will do the change.

nattsw (approved):
Thanks!

PR #842, opened 2022-05-09
REFACTOR: ChatNotifier mention types expansion.

romanrizzi:
When we want to notify users of a chat message, the `ChatNotifier` is our only source of truth. This service selects and groups the users by mention types, expanding direct, group, and channel-wide mentions. It also filters
unreachable users we can't notify because they don't have chat access or are not participating if the channel is private.

 It doesn't send notifications itself but generates a "map" and gives it to a scheduled job. This structure includes all the users to notify and the mention type metadata, which we need in order to ship different data in the
notifications.

The current implementation has a bug that includes group members who aren't participating in a private conversation if the message mentions a group. The scheduled job has an additional membership check, so it doesn't cause any
problems right now, but if another component uses this map to do other things, there's a risk it will leak private messages.

It happens because the application of the expansion rules and the map generation are happening in two different points, and we don't apply the `can_see_chat_channel?` filter during the later. The duplication makes the  logic more
error-prone, so this PR refactor the class to do both in a single step, avoiding losing information or missing filters. It also adds some unit tests to complement the acceptance ones we already have.

### DIFF
--- a/app/jobs/regular/create_chat_mention_notifications.rb
+++ b/app/jobs/regular/create_chat_mention_notifications.rb
@@ -52,32 +52,36 @@ module Jobs
     end
 
     def build_payload_for(membership, identifier_type:)
-      translation_prefix = @is_direct_message_channel ?
-        "discourse_push_notifications.popup.direct_message_chat_mention" :
-        "discourse_push_notifications.popup.chat_mention"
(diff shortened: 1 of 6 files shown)


pmusaraj (approved):
Looks generally good to me, thanks for adding the additional unit specs.

PR #843, opened 2022-05-09
FIX: Don't hide the new messages indicator beside the channel header

romanrizzi:
Reverts discourse/discourse-chat#835

### Isolated:
Screen Shot 2022-05-09 at 15 30 54

### Drawer:
Screen Shot 2022-05-09 at 15 32 42

### Mobile:
Screen Shot 2022-05-09 at 15 33 55

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -31,9 +31,11 @@ const MAX_RECENT_MSGS = 100;
 const STICKY_SCROLL_LENIENCE = 4;
 const READ_INTERVAL = 1000;
 const PAGE_SIZE = 50;
+const MESSAGES_ABOVE_NEW_MESSAGE_INDICATOR = 2;
 
-const PAST = "Past";
-const FUTURE = "Future";
+const PAST = "past";
+const FUTURE = "future";
+const BOTH = "both";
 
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -447,6 +473,22 @@ export default Component.extend({
     }
   },
 
+  scrollToSeparator(indicatorClass) {
+    const separatorEl = this._scrollerEl.querySelector(
+      `.chat-message-separator.${indicatorClass}`
+    );
+    const aboveIndicatorOffset = this.element.clientHeight / 7;
+
+    if (separatorEl) {
```
jjaffeux, line 482: This is weird to me because you access the DOM outside of after render block to check if you should do the operation. All of this should be in afterRender block

jjaffeux (approved):

assets/javascripts/discourse/components/chat-live-pane.js
romanrizzi: I fixed it on the latest commit.

romanrizzi:
I realized  this also affected highlighted messages, so ended up moving all the scrolling logic inside the `scrollToMessage` function instead of having a separate one for the separator.

danielwaterworth (approved):

PR #848, opened 2022-05-11
REFACTOR: chat-message-reaction

jjaffeux:
- tagLess
- native class
- better support for mobile with long press
- tests
- correctly shows emoji name in title
- uses a real button instead of role="button"

### DIFF
--- a/assets/javascripts/discourse/templates/components/chat-message-reaction.hbs
+++ b/assets/javascripts/discourse/templates/components/chat-message-reaction.hbs
@@ -1,5 +1,29 @@
-{{replace-emoji emojiString}}
+<button
+  {{on "click" (action "handleClick")}}
+  {{on "touchstart" (action "handleTouchstart")}}
+  {{on "touchend" (action "handleTouchend")}}
+  {{on "mouseover" (action "handleMouseover")}}
+  {{on "mouseout" (action "handleMouseout")}}
+  tabindex="0"
+  class={{concat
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse/templates/components/chat-message-reaction.hbs
```diff
@@ -1,5 +1,30 @@
...
+  {{on "click" (action "handleClick")}}
+  {{on "touchstart" (action "handleTouchstart")}}
+  {{on "touchend" (action "handleTouchend")}}
+  {{on "mouseover" (action "handleMouseover")}}
+  {{on "mouseout" (action "handleMouseout")}}
+  tabindex="0"
+  class={{concat
+    this.class
+    " chat-message-reaction "
+    this.emoji
```
nbianca, line 12: Does `this.emoji` include emoji modifiers such as `t3` in `:woman:t3:`? I ask this just to make sure that the generated class name is always valid and does not contain `:`.

nbianca (approved):
:+1:

assets/javascripts/discourse/templates/components/chat-message-reaction.hbs
jjaffeux: good question but AFAIK a `:` in a class name is possible: https://codepen.io/jjaffeux/pen/dydXJQv

PR #852, opened 2022-05-12
REFACTOR: chat-message-info

jjaffeux:
Not this commit will also allow primary group flair to show next to username.

### View mode

- [ ] isolated (full-screen)

### Browsers

- [ ] safari
- [ ] chrome
- [ ] firefox

### Device

- [ ] desktop – wide screen

### Ember

- [ ] ember-cli
- [ ] legacy

### DIFF
--- a/app/serializers/chat_message_serializer.rb
+++ b/app/serializers/chat_message_serializer.rb
@@ -13,7 +13,7 @@ class ChatMessageSerializer < ApplicationSerializer
     :edited,
     :reactions
 
-  has_one :user, serializer: BasicUserSerializer, embed: :objects
+  has_one :user, serializer: UserCardSerializer, embed: :objects
   has_one :chat_webhook_event, serializer: ChatWebhookEventSerializer, embed: :objects
   has_one :in_reply_to, serializer: ChatInReplyToSerializer, embed: :objects
   has_many :uploads, serializer: UploadSerializer, embed: :objects
--- a/assets/javascripts/discourse/templates/components/chat-message-info.hbs
+++ b/assets/javascripts/discourse/templates/components/chat-message-info.hbs
@@ -1,0 +1,41 @@
+<div class="chat-message-info">
+  {{#if @message.chat_webhook_event}}
+    {{#if @message.chat_webhook_event.username}}
+      <span class="chat-message-info__username {{this.usernameClasses}}">
+        {{@message.chat_webhook_event.username}}
+      </span>
+    {{/if}}
+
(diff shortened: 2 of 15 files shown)


jjaffeux:
Screenshot 2022-05-13 at 00 10 34

assets/javascripts/discourse/templates/components/chat-message-info.hbs
```diff
@@ -0,0 +1,41 @@
...
+        {{user-avatar-flair user=@message.user}}
+      </span>
+    </span>
+  {{/if}}
+
+  <span class="chat-message-info__date">
+    {{format-chat-date @message @details}}
+  </span>
+
+  {{#if (or @message.reviewable_id (eq @message.user_flag_status 0))}}
```
martin-brennan, line 30: I find this a little hard to follow, is it `if the message has a reviewable id AND the user flag status equals 0` or is it `if the message has a reviewable id OR the user flag status equals 0`?

martin-brennan (approved):
Looking really nice, love the new separation, makes chat-message a lot leaner :+1:

app/serializers/chat_message_serializer.rb
```diff
@@ -13,7 +13,7 @@ class ChatMessageSerializer < ApplicationSerializer
     :edited,
     :reactions
 
-  has_one :user, serializer: BasicUserSerializer, embed: :objects
+  has_one :user, serializer: UserCardSerializer, embed: :objects
```
davidtaylorhq, line 16: The UserCardSerializer includes a data from a *lot* of tables, so will be much slower. Do we need all of it?

To avoid N+1s, we'll need to make sure all the data is preloaded like

https://github.com/discourse/discourse/blob/b65ecf69871656155c381343cd84de919cbbf803/app/controllers/users_controller.rb#L112-L120

jjaffeux: Yep this is why I pointed to it in code-review. No we don't need all of it, we just need the flair stuff. But In the end that would mean a custom user serializer no?

davidtaylorhq: Yeah either a custom user serializer, or you could wrap up the user in a ChatPosterSerializer (similar to core's TopicPosterSerializer)

https://github.com/discourse/discourse/blob/e599b5b08f4323d50fb2dd584dbb06333d60b995/app/serializers/topic_poster_serializer.rb

jjaffeux: Will look at this, was looking for something like this, thanks

PR #854, opened 2022-05-13
FIX: Immediately queue notification jobs

pmusaraj:
Waiting 3 seconds here means that stuff gets queued and either sent in wrong order or some items would be skipped. 

PR also renames the two similar notifications jobs for consistency.


app/jobs/regular/chat_notify_watching.rb
```diff
@@ -54,7 +54,7 @@ def send_notifications(membership)
         MessageBus.publish("/chat/notification-alert/#{user.id}", payload, user_ids: [user.id])
       end
 
-      if membership.mobile_notifications_always? && !online_user_ids.include?(user.id)
```
pmusaraj, line 57: This check was redundant, we're exiting early in line 35 on the same condition.

romanrizzi (approved):

martin-brennan:
Thanks for changing this, it also will fix this edge case in moving messages between channels, I will remove the TODO now:

https://github.com/discourse/discourse-chat/blob/e61702554cf695a742cf9263e0c06aa0f0478201/lib/message_mover.rb#L105-L109

PR #855, opened 2022-05-13
UX: Tweak the archived channel UI

CvX:
Before / After

Screen Shot 2022-05-13 at 20 39 32

Screen Shot 2022-05-13 at 20 39 17

### DIFF
--- a/assets/stylesheets/common/common.scss
+++ b/assets/stylesheets/common/common.scss
@@ -1204,6 +1204,12 @@ body.has-full-page-chat {
       text-align: right;
       padding-right: 1em;
     }
+
+    .chat-channel-status {
+      .d-icon {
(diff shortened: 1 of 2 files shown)


jjaffeux:
How is longer text looking when in mobile or float?

jjaffeux:
:hugs: ?

CvX:
The longer text was _before_! 😃 My thinking was:

1. Displaying "Archive channel completed" (grammatically dubious btw) for the rest of the time is a bit weird, as it looks like a just-in-time message but is actually a description for the archived channel 🤷 
2. The name "destination topic" is uncommon and makes me think of "destination vacation" 🏝 😂
3. The period after the label "Channel is archived" was unnecessary

jjaffeux (approved):

PR #859, opened 2022-05-16
FIX: Add channel ID attribute to chat quotes

martin-brennan:
We have run across various complications when trying to
link to a channel by channel title (see /t/66638 and
https://github.com/discourse/discourse-chat/pull/857 for
examples) so instead we are just going to add a channel
ID to the chat quotes instead for easier linking.

This commit also improves DM channel titles for chat quotes,
which were not taken into account in the original design,
but was later allowed in 1f10fc6c4cf5642d21419eff30dc986d0cd33a7c.

### DIFF
--- a/app/models/direct_message_channel.rb
+++ b/app/models/direct_message_channel.rb
@@ -8,15 +8,25 @@ class DirectMessageChannel < ActiveRecord::Base
     users.include?(user)
   end
 
-  def chat_channel_title_for_user(chat_channel, user)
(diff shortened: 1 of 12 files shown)


jjaffeux (approved):

jjaffeux:
@martin-brennan test failure looks legit

martin-brennan:
Thanks @jjaffeux forgot to update again after last change, should go through this time (thanks for approval too :+1: )

PR #864, opened 2022-05-18
FIX: composer disabled state was incorrect

jjaffeux:


jjaffeux:
The issue:

image

New behavior:

- disabled:
Screenshot 2022-05-18 at 15 49 24

-enabled:
Screenshot 2022-05-18 at 15 49 31

PR #867, opened 2022-05-18
FIX: reduces margin before dots of replying indicator

jjaffeux:


assets/stylesheets/common/chat-replying-indicator.scss
```diff
@@ -22,6 +22,8 @@
   .chat-replying-indicator__wave {
     flex: 0 0 auto;
     display: inline-flex;
+    // negative margin to account for linebreak space
+    margin-left: -3px;
```
davidtaylorhq, line 26: This feels kinda hacky... is the space always guaranteed to be 3px for all font sizes? 🤔 

In core, we use hbs comments to avoid the linebreak altogether.... but that also feels pretty hacky

https://github.com/discourse/discourse/blob/ddb5d8815817dc448cf17042dcc8994db5691a00/plugins/discourse-presence/assets/javascripts/discourse/templates/components/topic-presence-display.hbs#L11-L12

davidtaylorhq: This solves the problem for me locally:

```diff
diff --git a/assets/stylesheets/common/chat-replying-indicator.scss b/assets/stylesheets/common/chat-replying-indicator.scss
index f222a08..a95446e 100644
--- a/assets/stylesheets/common/chat-replying-indicator.scss
+++ b/assets/stylesheets/common/chat-replying-indicator.scss
@@ -3,6 +3,8 @@
 }
 
 .chat-replying-indicator {
+  display: inline-flex;
+
   color: var(--primary-medium);
   padding-bottom: unquote(
     "max(calc(env(safe-area-inset-bottom) - 0.75em), 0.25em)"
```
(this was previously done via 698b4a0aa82a195acf8b02a2d5c7b413eb1d1ad3, but recent refactoring changed the html/class structure)

davidtaylorhq: Ok, if you prefer the `-3px` to the `display: inline-flex;` solution, then happy to approve :+1:

davidtaylorhq (approved):

assets/stylesheets/common/chat-replying-indicator.scss
jjaffeux: no I changed to inline, weird :D

PR #871, opened 2022-05-20
FIX: minor fixes to msg-actions

jjaffeux:
- ensures menu is disappearing correctly when interacting with another message
- ensures mobile menu is not showing focus state


jjaffeux:
Mostly this commit avoids this case:
Screenshot 2022-05-20 at 13 37 48

PR #872, opened 2022-05-20
REFACTOR: move draft logic its own service

fzngagan:
I took some inspiration for the syntax from https://github.com/discourse/discourse/blob/a8163a5c0cde06220ad71b2f96a304c28631a38a/app/assets/javascripts/discourse/app/services/presence.js#L246
Tried replacing init with constructor but the injected stuff can't be accessed i.e. this.currentUser

### Ember

- [x] ember-cli
- [x] legacy

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -65,6 +65,7 @@ export default Component.extend({
   targetMessageId: null,
 
   chat: service(),
+  chatDraftHandler: service(),
   router: service(),
   chatComposerPresenceManager: service(),
 
@@ -148,7 +149,7 @@ export default Component.extend({
             this._startLastReadRunner();
           });
       } else {
-        this.set("draft", this.chat.getDraftForChannel("draft"));
+        this.set("draft", this.chatDraftHandler.getDraftForChannel("draft"));
       }
     }
     this.currentUserTimezone = this.currentUser?.resolvedTimezone(
@@ -196,7 +197,7 @@ export default Component.extend({
   },
 
   loadDraftForChannel(channelId) {
-    this.set("draft", this.chat.getDraftForChannel(channelId));
+    this.set("draft", this.chatDraftHandler.getDraftForChannel(channelId));
   },
 
   _fetchMoreMessages(direction) {
@@ -1194,7 +1195,7 @@ export default Component.extend({
         user: draft.replyToMsg.user,
       };
     }
-    this.chat.setDraftForChannel(this.chatChannel, draft);
+    this.chatDraftHandler.setDraftForChannel(this.chatChannel, draft);
     this.set("draft", draft);
   },
 
--- a/assets/javascripts/discourse/services/chat-draft-handler.js
+++ b/assets/javascripts/discourse/services/chat-draft-handler.js
@@ -1,0 +1,49 @@
+import Service from "@ember/service";
+import discourseDebounce from "discourse-common/lib/debounce";
+import { ajax } from "discourse/lib/ajax";
+
+export default class ChatDraftHandler extends Service {
+  init() {
+    super.init(...arguments);
+
+    this.draftStore = {};
+    if (this.currentUser.chat_drafts) {
+      this.currentUser.chat_drafts.forEach((draft) => {
+        this.draftStore[draft.channel_id] = JSON.parse(draft.data);
+      });
+    }
+  }
+
+  _saveDraft(channelId, draft) {
+    const data = { channel_id: channelId };
+    if (draft) {
+      data.data = JSON.stringify(draft);
+    }
+
+    ajax("/chat/drafts", { type: "POST", data });
+  }
+
+  setDraftForChannel(channel, draft) {
+    if (
+      draft &&
+      (draft.value || draft.uploads.length > 0 || draft.replyToMsg)
+    ) {
+      this.draftStore[channel.id] = draft;
+    } else {
+      delete this.draftStore[channel.id];
+      draft = null; // _saveDraft will destroy draft
+    }
+
+    discourseDebounce(this, this._saveDraft, channel.id, draft, 2000);
+  }
+
+  getDraftForChannel(channelId) {
+    return (
+      this.draftStore[channelId] || {
+        value: "",
+        uploads: [],
+        replyToMsg: null,
+      }
+    );
+  }
+}
--- a/assets/javascripts/discourse/services/chat.js
+++ b/assets/javascripts/discourse/services/chat.js
@@ -63,13 +63,6 @@ export default Service.extend({
       this._subscribeToChannelEdits();
       this._subscribeToChannelStatusChange();
       this.presenceChannel = this.presence.getChannel("/chat/online");
-      this.draftStore = {};
-
-      if (this.currentUser.chat_drafts) {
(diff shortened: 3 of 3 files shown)


jjaffeux:
It needs tests :)

assets/javascripts/discourse/services/chat-draft-handler.js
```diff
@@ -0,0 +1,49 @@
...
+  _saveDraft(channelId, draft) {
+    const data = { channel_id: channelId };
+    if (draft) {
+      data.data = JSON.stringify(draft);
+    }
+
+    ajax("/chat/drafts", { type: "POST", data });
+  }
+
+  setDraftForChannel(channel, draft) {
```
jjaffeux, line 26: we don't need to have draft in the name of the function here, given we are already in draft service, same of others

assets/javascripts/discourse/services/chat-draft-handler.js
```diff
@@ -0,0 +1,49 @@
+import Service from "@ember/service";
+import discourseDebounce from "discourse-common/lib/debounce";
+import { ajax } from "discourse/lib/ajax";
+
+export default class ChatDraftHandler extends Service {
+  init() {
+    super.init(...arguments);
+
+    this.draftStore = {};
```
jjaffeux, line 9: should be prefixed with _

assets/javascripts/discourse/services/chat-draft-handler.js
```diff
@@ -0,0 +1,49 @@
...
+      delete this.draftStore[channel.id];
+      draft = null; // _saveDraft will destroy draft
+    }
+
+    discourseDebounce(this, this._saveDraft, channel.id, draft, 2000);
+  }
+
+  getDraftForChannel(channelId) {
+    return (
+      this.draftStore[channelId] || {
```
jjaffeux, line 42: extract the right part into some const; eg: `DEFAULT_DRAFT`

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1194,7 +1195,7 @@ export default Component.extend({
         user: draft.replyToMsg.user,
       };
     }
-    this.chat.setDraftForChannel(this.chatChannel, draft);
+    this.chatDraftHandler.setDraftForChannel(this.chatChannel, draft);
```
jjaffeux, line 1198: Why do we need to pass the draft here when it's supposed to be hold in the handler?

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -196,7 +197,7 @@ export default Component.extend({
   },
 
   loadDraftForChannel(channelId) {
-    this.set("draft", this.chat.getDraftForChannel(channelId));
+    this.set("draft", this.chatDraftHandler.getDraftForChannel(channelId));
```
jjaffeux, line 200: Do we need to create a local property here? Can't we just always the draft from the handler?

assets/javascripts/discourse/services/chat-draft-handler.js
```diff
@@ -0,0 +1,49 @@
...
+      draft &&
+      (draft.value || draft.uploads.length > 0 || draft.replyToMsg)
+    ) {
+      this.draftStore[channel.id] = draft;
+    } else {
+      delete this.draftStore[channel.id];
+      draft = null; // _saveDraft will destroy draft
+    }
+
+    discourseDebounce(this, this._saveDraft, channel.id, draft, 2000);
```
jjaffeux, line 37: I feel like this debounce logic should be in the composer, this could be different based on context

assets/javascripts/discourse/services/chat-draft-handler.js
fzngagan: Done

assets/javascripts/discourse/services/chat-draft-handler.js
fzngagan: Done

assets/javascripts/discourse/components/chat-live-pane.js
fzngagan: Currently I've created a computed prop like the following.

```
  @discourseComputed("chatChannel.id")
  draft(channelId) {
    return this.chatDraftHandler.getForChannel(channelId);
  }
```

But its not working. I'm investigating why.

fzngagan: Thanks for the pointer, its really helpful. But in this case, its an old school factory component.

What actually fixes it though, is if I make it a local property as it was earlier and set it it in `didReceiveAttrs`

jjaffeux: Sorry was on mobile, I didn't have context, and usually this is the issue :p

assets/javascripts/discourse/services/chat-draft-handler.js
```diff
@@ -0,0 +1,48 @@
...
+
+    this._draftStore = {};
+    if (this.currentUser.chat_drafts) {
+      this.currentUser.chat_drafts.forEach((draft) => {
+        this._draftStore[draft.channel_id] = JSON.parse(draft.data);
+      });
+    }
+  }
+
+  @bind
```
jjaffeux, line 23: what is the bind doing here?

jjaffeux: If this is for the debounce, I would recommend you have a debounce handler in chat-live-pane which would call this instead of directly debouncing the call to the draft handler

assets/javascripts/discourse/services/chat-draft-handler.js
```diff
@@ -0,0 +1,48 @@
...
+  setForChannel({ channelId, draft }) {
```
jjaffeux, line 24: I'm still not sure about the name here, also we say channel but we want a channelId, so it's kinda misleading.

I also think I would prefer a signature `(channelId, draft)`  same for get `(channelId)`

assets/javascripts/discourse/services/chat-draft-handler.js
fzngagan: No, actually I wasn't able to access `this._draftStore` in these methods. I looked around and figured that `@bind` fixed it.

assets/javascripts/discourse/services/chat-draft-handler.js
fzngagan: I started with simple `get` and `set` but quickly realised that it overrides ember's `get` and `set` so its out of question. 

I totally agree that signature should be `(channelId, draft)` and debounce method should be in `chat-live-pane`.

assets/javascripts/discourse/components/chat-live-pane.js
fzngagan: I've made it a local property. The probably reason for computed prop not working is, it needs to be set at a point in time but computed is set as soon as its read the first time.

I can dig further if needed.

assets/javascripts/discourse/services/chat-draft-handler.js
fzngagan: I realized that bind isn't needed and removed it. But I promise things weren't working earlier without it. Not sure what that was.

jjaffeux:
I have seen the update, will look in next days

jjaffeux:
Closing this for now, thanks for the work though, but I can't focus on this ATM.

PR #878, opened 2022-05-23
DEV: allows plugins to enable markdown features

jjaffeux:
Usage in plugin.rb:

```ruby
if respond_to?(:discourse_chat)
  discourse_chat&.enable_markdown_feature('discourse-video')
end
```


davidtaylorhq (approved):
Nice!

PR #880, opened 2022-05-24
DEV: Delete old Tag channel code

martin-brennan:
We decided a while back not to allow tag channels in
chat, this commit deletes/cleans up all the old code
referring to Tag chatables. Only Topic and Category
can be used as a chatable (and DirectMessageChannel too
but that's a special case.)


jjaffeux (approved):

CvX:
What about the existing data? Should there be a migration that removes tag-related rows?

(I think on dev somebody removed those manually? there was a tag-based channel just a couple of hours ago)

jjaffeux:
> What about the existing data? Should there be a migration that removes tag-related rows?
> 
> (I think on dev somebody removed those manually? there was a tag-based channel just a couple of hours ago)

I would prefer we do it in two steps, let's remove the data in few weeks. Even if plugin is still experimental, let's give people a chance to see tags channels are not here anymore, and react if they are bothered?

martin-brennan:
@CvX I deleted the ones internally on dev, they were empty already. I checked the main sites we have using chat and none of them had tag channels. If we eventually find more then we can delete them, but I don't think anyone really was using chat but us when we removed the option to create them.

PR #883, opened 2022-05-24
FEATURE: Add a button to switch back to small chat

oblakeerickson:
This feature, in addition to adding a button to get back to small chat
from being in the full-page chat, adds *memory* to the window size.
This means that if you prefer to open chat in full-size when using the
sidebar it will remember which preference you have set when opening
channels.

See: https://meta.discourse.org/t/221388

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -25,6 +25,7 @@ import { spinnerHTML } from "discourse/helpers/loading-spinner";
 import { decorateGithubOneboxBody } from "discourse/initializers/onebox-decorators";
 import highlightSyntax from "discourse/lib/highlight-syntax";
 import { applyLocalDates } from "discourse/lib/local-dates";
+import { defaultHomepage } from "discourse/lib/utilities";
 
 const MAX_RECENT_MSGS = 100;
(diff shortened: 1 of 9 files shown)


martin-brennan (commented):
This looks good but I just have a couple of questions:

1. This only remembers the preference as long as you don't refresh, right? Should we maybe sync the store value to a user option so it's remembered always even through refreshes (and in which case maybe we surface it in the user preference UI, though that may be a little extreme)
2. This does not interfere with isolated chat behaviour right?

oblakeerickson:
> 1. This only remembers the preference as long as you don't refresh, right?

It saves it in local storage, so it should remember on refresh. It will be per device though.

> and in which case maybe we surface it in the user preference UI

The design comments from Sam were to avoid having a user preference for this.

martin-brennan:
@oblakeerickson ah I see I didn't realise KeyValueStore saved to localstorage :+1: Thanks for the clarification, will approve :)

oblakeerickson:
> why do we need `chatWindowFullPage` still if we have a dedicated `chat-window-store`?

well... I added it as a wrapper for the chat-window-store inside of chat.js because I couldn't get it the store service to work inside of the chat-header-icon widget. /shrug

PR #888, opened 2022-05-25
PERF: Update all decorators to use decorateChatMessage api

davidtaylorhq:
Some transformations were being implemented in the `chat-live-pane` component, rather than using the `chat-message` decoration system. This means that they were often being applied multiple times to the same message.

This commit refactors all the decorators from chat-live-pane so that they use the standard API. This means they can take advantage of the perf improvements from 43ce6e0c65


nbianca (approved):

jjaffeux:
It was on my todo-list, not only was it applied multiple times for those not using chat api, or more exactly not defined in chat message but it was applying them each time there was an update on the whole list.

PR #891, opened 2022-05-26
FIX: better handling of non existing messages

jjaffeux:
Before this fix, two conditions (AFAIK) could lead to data corruption:

- delete_old_chat_message job which could delete messages (and not just trash). There was no dependent behavior on related models

- some endpoint would not check if a message related to an impacted channel, eg:

```
PUT /chat/42/read/666.json
```

Where 666 is a message id NOT belonging to the channel with id 42. Moreover, there was no check server side before updating the `last_read_message_id` of a membership that the provided `message_id` was belonging to this channel. As a result we would sometimes store an incorrect `last_read_message_id` which is then used for messages lookup and ends up loading the messages of another channel. We attempt to reduce the chances for the frontend to make an incorrect request.

Note that this issue has been here for quite a long time but has been made more visible by https://github.com/discourse/discourse-chat/commit/17f1e4213a58670265d3ccba959082a7b0442d7a#diff-8903d1d4f083bdb0595643bf27bd4a5d3dfafd49a912bb061c42135971e6aa8dR172 as it's making a bigger use of `last_read_message_id` than before.

---

This commit both attempts to avoid future corrupting and also fixes current corrupted data. This commit also drops a table `user_chat_channel_last_reads` which appears to not be used anymore but is holding references to messages, so better drop it.

### DIFF
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -154,8 +154,9 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
       raise Discourse::InvalidParameters.new(:message_id)
     end
 
-    chat_message = ChatMessage.find_by(chat_channel_id: @chat_channel.id, id: params[:message_id])
-    raise Discourse::NotFound unless chat_message
+    unless ChatMessage.exists?(chat_channel_id: @chat_channel.id, id: params[:message_id])
+      raise Discourse::NotFound
+    end
 
     membership.update!(last_read_message_id: params[:message_id])
 
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -149,6 +149,14 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
 
     membership = UserChatChannelMembership.find_by(user: current_user, chat_channel: @chat_channel, following: true)
     raise Discourse::NotFound if membership.nil?
+
+    if membership.last_read_message_id && params[:message_id].to_i < membership.last_read_message_id
+      raise Discourse::InvalidParameters.new(:message_id)
+    end
+
+    chat_message = ChatMessage.find_by(chat_channel_id: @chat_channel.id, id: params[:message_id])
+    raise Discourse::NotFound unless chat_message
+
     membership.update!(last_read_message_id: params[:message_id])
 
     Notification
--- a/app/jobs/scheduled/delete_old_chat_messages.rb
+++ b/app/jobs/scheduled/delete_old_chat_messages.rb
@@ -15,7 +15,10 @@ module Jobs
       ChatMessage
         .in_public_channel
         .created_before(SiteSetting.chat_channel_retention_days.days.ago)
-        .destroy_all
+        .in_batches(of: 200).each do |relation|
+          destroyed_ids = relation.destroy_all.pluck(:id)
+          reset_last_read_message_id(destroyed_ids)
+        end
     end
 
     def delete_dm_channel_messages
@@ -24,11 +27,20 @@ module Jobs
       ChatMessage
         .in_dm_channel
         .created_before(SiteSetting.chat_dm_retention_days.days.ago)
-        .destroy_all
+        .in_batches(of: 200).each do |relation|
+          destroyed_ids = relation.destroy_all.pluck(:id)
+          reset_last_read_message_id(destroyed_ids)
+        end
     end
 
     def valid_day_value?(setting_name)
       (SiteSetting.public_send(setting_name) || 0).positive?
+    end
+
+    def reset_last_read_message_id(ids)
+      UserChatChannelMembership
+        .where(last_read_message_id: ids)
+        .update_all(last_read_message_id: nil)
     end
   end
 end
--- a/db/post_migrate/20220526135414_remove_corrupted_last_read_message_id.rb
+++ b/db/post_migrate/20220526135414_remove_corrupted_last_read_message_id.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 class RemoveCorruptedLastReadMessageId < ActiveRecord::Migration[7.0]
-  def change
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+
+  def up
     # Delete memberships for deleted channels
     execute <<~SQL
       DELETE FROM user_chat_channel_memberships uccm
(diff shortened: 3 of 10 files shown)


app/controllers/chat_controller.rb
```diff
@@ -149,6 +149,9 @@ def update_user_last_read
 
     membership = UserChatChannelMembership.find_by(user: current_user, chat_channel: @chat_channel, following: true)
     raise Discourse::NotFound if membership.nil?
+    chat_message = ChatMessage.find_by(chat_channel_id: @chat_channel.id, id: params[:message_id])
+    raise Discourse::NotFound unless chat_message
+
```
romanrizzi, line 154: We should probably check that `params[:message_id]`is greater than `membership#last_read_message_id` to prevent the frontend from accidentally "unreading" messages. What do you think?

jjaffeux: ok :+1:

jjaffeux: However we should do it in the future IMO otherwise it might conflict with resolving the problem here as it might always consider the last_read_message_id is greater given it's an incorrect one

davidtaylorhq: For the record (because I was confused 😉 ).

It looks like this check was added, and the problem with existing channels has been handled via the post-deploy migration :+1: Marking as resolved

app/jobs/scheduled/delete_old_chat_messages.rb
```diff
@@ -12,23 +12,35 @@ def execute(args = {})
     def delete_public_channel_messages
       return unless valid_day_value?(:chat_channel_retention_days)
 
-      ChatMessage
+      destroyed_ids = ChatMessage
```
davidtaylorhq, line 15: This should probably be done in batches... otherwise we could end up trying to load 10000s of messages into memory in one go. 

(but I realise that problem was not introduced by this PR, so feel free to ignore for now)

db/post_migrate/20220526135414_remove_corrupted_last_read_message_id.rb
```diff
@@ -0,0 +1,38 @@
+class RemoveCorruptedLastReadMessageId < ActiveRecord::Migration[7.0]
+  def change
```
davidtaylorhq, line 2: We shouldn't use ActiveRecord in migrations (to make sure the migration continues to work forever, regardless of future changes/removals of ActiveRecord models).

So I think this needs to be re-written in raw SQL statements 😓

davidtaylorhq: Totally untested, but I'm imagining something like this:

```ruby
# Delete memberships for deleted channels
execute <<~SQL
  DELETE FROM user_chat_channel_memberships
  LEFT JOIN chat_channels ON chat_channels.id = user_chat_channel_memberships.chat_channel_id
  WHERE chat_channels.id IS NULL
SQL

# Delete messages for deleted channels
execute <<~SQL
  DELETE FROM chat_messages
  LEFT JOIN chat_channels ON chat_channels.id = chat_messages.chat_channel_id
  WHERE chat_channels.id IS NULL
SQL

# Reset highest_channel_message_id if the message does not exist, or is part of a different channel
execute <<~SQL
  WITH highest_channel_message_id AS (
    SELECT chat_channel_id, max(chat_messages.id) as highest_id
    FROM chat_messages
    GROUP_BY chat_channel_id
  )
  UPDATE user_chat_channel_memberships uccm
  SET uccm.last_read_message_id = highest_channel_message_id.highest_id
  FROM highest_channel_message_id
  WHERE uccm.last_read_message_id IS NOT NULL
  AND uccm.last_read_message_id NOT IN (
    SELECT id FROM chat_messages WHERE chat_messages.chat_channel_id = uccm.chat_channel_id
  )
SQL
```

jjaffeux: Oh yes I re

> We shouldn't use ActiveRecord in migrations (to make sure the migration continues to work forever, regardless of future changes/removals of ActiveRecord models).
> 
> So I think this needs to be re-written in raw SQL statements 😓

Oh yes I remember rewriting many migrations in automation to raw sql... thanks

app/jobs/scheduled/delete_old_chat_messages.rb
jjaffeux: probably a good time to fix it too though... Given it's ran every day, maybe we limit to 10000 and do a find_each ?

davidtaylorhq: Sounds good to me :+1: Even with find_each, make sure it's not run inside a transaction because that'll cause ActiveRecord to keep all the modified models in memory. (I had that fun with https://github.com/discourse/discourse/commit/67b23c0e22ed851ec7924709b53e416356980682)

jjaffeux:
@davidtaylorhq @martin-brennan this is ready for review.

Few notes:
- I wonder if we have to destroy `uploads` related to `chat_uploads`, it's possible it's used at other places right?
- Technically chat_drafts can old a reference to delete message in data column where the json has a `replyToMsg`, I doubt it will cause any issue, so I didn't handle it

db/post_migrate/20220531105951_drop_user_chat_channel_last_reads.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class DropUserChatChannelLastReads < ActiveRecord::Migration[7.0]
+  def up
+    # usage has been dropped in https://github.com/discourse/discourse-chat/commit/1c110b71b28411dc7ac3ab9e3950e0bbf38d7970
+    # but apparently table never got dropped
+    drop_table :user_chat_channel_last_reads
```
davidtaylorhq, line 7: Does this work? I suspect you'll need to add a `DROPPED_TABLES` constant like https://github.com/discourse/discourse/blob/main/spec/fixtures/db/post_migrate/drop_table/20990309014013_drop_email_logs_table.rb

davidtaylorhq (approved):

db/post_migrate/20220531105951_drop_user_chat_channel_last_reads.rb
jjaffeux: I saw we were doing this at lots of places in core 🤷‍♂️

martin-brennan: I think we should definitely be using `TableDropper` as per https://github.com/discourse/discourse/commit/6a3c8fe69c16ad7360046f145db6689c18e91005

I am sure I have gotten pulled up on this by automated rules, it should be stopped by https://github.com/discourse/discourse/blob/9ac85d61632dae6020a627e02aa40c93e9b5086a/lib/migration/safe_migrate.rb so I think maybe this would break after merge when we run the plugin tests?

So yeah should follow the pattern here https://github.com/discourse/discourse/blob/9ac85d61632dae6020a627e02aa40c93e9b5086a/db/migrate/20201110110952_drop_github_user_infos.rb . Maybe the places doing it in core were done before this rule was added?

spec/migrations/remove_corrupted_last_read_message_id_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+      before do
+        Bookmark.register_bookmarkable(ChatMessageBookmarkable)
+      end
+
+      fab!(:message_1) { Fabricate(:chat_message) }
+      fab!(:message_2) { Fabricate(:chat_message) }
+      let!(:bookmark_1) { Fabricate(:bookmark, bookmarkable: message_1) }
+      let!(:bookmark_2) { Fabricate(:bookmark, bookmarkable: message_2) }
+
+      it 'destroys ChatMessage of this message' do
```
martin-brennan, line 139:
```suggestion
      it 'destroys ChatMessage bookmarks of this message' do
```

martin-brennan (approved):
Looks great thanks for doing this, my main comment is about the drop table migration. Also agree with @david, no need to delete `Upload` records since they get cleaned up by core if they are dangling anyway :+1:

db/post_migrate/20220531105951_drop_user_chat_channel_last_reads.rb
martin-brennan: Maybe the rails 7 change possibly messed something up with our freedom patch? Since it should be stopping this https://github.com/discourse/discourse/blob/9ac85d61632dae6020a627e02aa40c93e9b5086a/lib/freedom_patches/safe_migrations.rb, see my other comment too, for some reason it did not get added inline here.

app/controllers/chat_controller.rb
```diff
@@ -149,6 +149,14 @@ def update_user_last_read
 
     membership = UserChatChannelMembership.find_by(user: current_user, chat_channel: @chat_channel, following: true)
     raise Discourse::NotFound if membership.nil?
+
+    if membership.last_read_message_id && params[:message_id].to_i < membership.last_read_message_id
+      raise Discourse::InvalidParameters.new(:message_id)
+    end
+
+    chat_message = ChatMessage.find_by(chat_channel_id: @chat_channel.id, id: params[:message_id])
```
ZogStriP, line 157: Do we use `chat_message` anywhere else? If not, should we instead do an existence check rather than loading the model from the database?

jjaffeux: No indeed we don't, this should be enough: https://github.com/discourse/discourse-chat/pull/891/commits/693b54a034188860c88b72b8c15502a2c6f9b086

ZogStriP (approved):
LGTM 👍

PR #893, opened 2022-05-27
FIX: Serialize dates using ISO8601

nbianca:
The date was not showed correctly in Safari because it could not
recognize the date format.

### View mode

- [ ] isolated (full-screen)

### Browsers

- [ ] safari
- [ ] chrome
- [ ] firefox

### Device

- [ ] desktop – wide screen

### Ember

- [ ] ember-cli
- [ ] legacy

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -263,6 +263,7 @@ after_initialize do
       args[:avatar_url] = message.user.avatar_template_url.gsub('{size}', '20')
       args[:cooked] = message.cooked
       args[:created_at] = message.created_at
+      args[:created_at_str] = message.created_at.iso8601
     end
 
     Mustache.render(DiscourseChat.onebox_template, args)
--- a/lib/onebox/templates/discourse_chat.mustache
+++ b/lib/onebox/templates/discourse_chat.mustache
@@ -33,7 +33,7 @@
 {{/cooked}}
 
 {{#cooked}}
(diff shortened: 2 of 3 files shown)


plugin.rb
```diff
@@ -263,6 +263,7 @@ def self.onebox_template
       args[:avatar_url] = message.user.avatar_template_url.gsub('{size}', '20')
       args[:cooked] = message.cooked
       args[:created_at] = message.created_at
+      args[:created_at_str] = message.created_at.iso8601
```
jjaffeux, line 266: This solution seems unexpected, can you elaborate please?

We show dates in safari at other places without doing this, no?

nbianca: The message I was seeing in console was:

```
Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
```

The previous date format was `2022-05-27 18:21:45 UTC`. The right kind of date looks like `2022-05-27T18:21:45Z`.

We do not normally have to do this because Rails serializers do this work. We also call iso8601 some other places too, where we serialize the data ourselves.

jjaffeux (approved):

PR #895, opened 2022-05-30
FIX: Message order consistency

martin-brennan:
In https://meta.discourse.org/t/moving-chat-messages-to-a-new-channel-jumbled-them-all-up/227301
we were having issues with moved messages between channels
being jumbled up. During investigations it was found that
the ChatController /messages route ordered messages by ID
but everything else, including the message mover, ordered by
created_at.

This commit makes things consistent by changing the /messages
route to also order by created_at, then ID. This should make it so the
order is always consistent in the UI and also for moved messages.

### DIFF
--- a/lib/message_mover.rb
+++ b/lib/message_mover.rb
@@ -57,7 +57,7 @@ class DiscourseChat::MessageMover
   private
 
   def find_messages(message_ids, channel)
-    ChatMessage.where(id: message_ids, chat_channel_id: channel.id).order(:created_at)
+    ChatMessage.where(id: message_ids, chat_channel_id: channel.id).order(created_at: :asc)
   end
 
   def create_temp_table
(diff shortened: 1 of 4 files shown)


lib/message_mover.rb
```diff
@@ -84,10 +84,29 @@ def bulk_insert_movement_metadata
...
+    moved_message_ids = DB.query_single(<<~SQL, query_args)
       INSERT INTO chat_messages(chat_channel_id, user_id, message, cooked, cooked_version, created_at, updated_at)
-      SELECT :destination_channel_id, user_id, message, cooked, cooked_version, CLOCK_TIMESTAMP(), CLOCK_TIMESTAMP()
+      SELECT :destination_channel_id,
+             user_id,
+             message,
+             cooked,
+             cooked_version,
+             :future_start::timestamp + (
+                ROW_NUMBER() OVER (ORDER BY chat_messages.created_at) * interval '1 second' / 2
```
martin-brennan, line 107: `0.5` second intervals is fairly arbitrary, I just thought we don't want it _too_ long because if a lot of messages are moved at once we start getting into the "minutes into the future" territory which is a little confusing.

martin-brennan:
Failure unrelated:

```
Test Errors
----------------------------------------------
  Test Failed: defaults
    Assertion Failed:
    at Object.eval (unit/lib/chat-window-store-test:8:21)
    at Object.eval (discourse/tests/helpers/qunit-helpers:225:17)
    at processModule (qunit:1192:16)
    at module$1 (qunit:1217:4)
    at discourseModule (discourse/tests/helpers/qunit-helpers:194:25)
    at eval (unit/lib/chat-window-store-test:4:37)
    at obj (discourse-loader:371:31)
```

CvX:
Adding 0.5s - can we do `order(:created_at, :id)` instead? Wouldn't that solve the problem?

Also, creating messages 5 seconds into the future does not sound like a good idea. We don't want to deal with stuff like a unread message separator saying "Tomorrow".

martin-brennan:
> Adding 0.5s - can we do order(:created_at, :id) instead? Wouldn't that solve the problem?

Hmm you may be right, perhaps we can try that as a first pass and do the other part later if needed.

> Also, creating messages 5 seconds into the future does not sound like a good idea. We don't want to deal with stuff like a unread message separator saying "Tomorrow".

I get your point but I would think it highly unlikely that this would happen or that anyone would see this. I could just space out the messages by 0.001 seconds :P Anyway, will remove that part for now and just do the order change.

martin-brennan:
@CvX I have made those changes, can you please take another look :)

martin-brennan:
Unrelated test failure:

```
Test Errors
----------------------------------------------
  Test Failed: Chat float open to DM channel with unread messages with sidebar off
    Assertion Failed: failed, expected argument to be truthy, was: false
      Expected: true, Actual: false
    at eval (acceptance/chat-test:924:21)
    at Object.eval (discourse/tests/helpers/qunit-helpers:381:9)
    at processModule (qunit:1192:16)
    at module$1 (qunit:1217:4)
    at acceptance (discourse/tests/helpers/qunit-helpers:377:25)
    at eval (acceptance/chat-test:859:32)
    at obj (discourse-loader:371:31)
```

CvX (approved):

PR #900, opened 2022-05-31
FIX: prevents destroyed/deleted chatable to crash admin page

jjaffeux:
Prior to this fix, destroying a chatable would cause a 500 on /admin/plugins/chat page as the chatable would be found and the serializer couldn't display its data.


CvX (approved):

jjaffeux:
I was adding code for handling a topics being trashed and category being destroyed, but we might not need it for topics in the near future, and category needs something special, will handle in a different PR.

martin-brennan (approved):

martin-brennan:
Merged to fix meta

PR #901, opened 2022-06-01
hide collapse button on mobile and isolate

oblakeerickson:

### View mode

- [x] isolated (full-screen)

### Browsers

- [x] safari
- [x] chrome
- [x] firefox

### Device

- [x] desktop – wide screen
- [x] mobile – `?mobile_view=1`

### Ember

- [x] ember-cli
- [x] legacy

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -48,7 +48,7 @@ export default Component.extend({
   stickyScroll: true,
   stickyScrollTimer: null,
   showChatQuoteSuccess: false,
-  showCompressBtn: false,
+  showCloseFullScreenBtn: false,
 
   editingMessage: null, // ?Message
   replyToMsg: null, // ?Message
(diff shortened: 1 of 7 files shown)


assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -97,6 +98,11 @@ export default Component.extend({
     );
 
     this.appEvents.on("chat:cancel-message-selection", this, "cancelSelecting");
+
+    this.set(
+      "showCompressBtn",
+      !this.currentUser.chat_isolated && !this.site.mobileView
```
tgxworld, line 104: I think we should add a test for this logic added here.

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -97,6 +98,11 @@ export default Component.extend({
     );
 
     this.appEvents.on("chat:cancel-message-selection", this, "cancelSelecting");
+
+    this.set(
+      "showCompressBtn",
```
tgxworld, line 103:
```suggestion
      "showCloseFullScreenBtn",
```

Minor suggestion because I feel like `compress` doesn't really describe what the button actually does on click.

oblakeerickson: Ya I like that better too.

pmusaraj (approved):

PR #905, opened 2022-06-03
FIX: ensures we unsubscribe from /chat-reply in draft mode

jjaffeux:
Before this fix, being in another channel and switching to dm creator would not reset replying-indicator state and you would see incorrect data.


jjaffeux:
Test failure is unrelated

tgxworld:
Is there a test we can add/update here? I see that the presence indicator is something that is tested in chat.

jjaffeux:
> Is there a test we can add/update here? I see that the presence indicator is something that is tested in chat.

Yes, it's a little bit awkward but that should be enough to ensure this line is not removed inadvertently: [test](https://github.com/discourse/discourse-chat/pull/905/commits/5aece0260294cccd0f4358a4de1bdeb249c7dbb7)

tgxworld (approved):
Looks good to me :)

PR #907, opened 2022-06-03
FIX: ensures we attempt to fill the current pane with messages

jjaffeux:
Not only does it give more context on tall screens, but also prevents a bug where you couldn't load previous messages as there wouldn't be any scroll event given the pane is not overflowing.


oblakeerickson (approved):

jjaffeux:
@oblakeerickson thanks :+1: I will wait for review of @pmusaraj too as he has some thoughts on the subject

pmusaraj (approved):
Looks good! 

(Saw some jQuery used in `isElementInViewport`, PR to remove in https://github.com/discourse/discourse/pull/17000.)

PR #913, opened 2022-06-07
FIX: Onebox should only show active users

oblakeerickson:
A onebox for a chat channel should only show users that are active, not
staged, and not suspended in the members list.

Maybe as a future feature we should boot users from the channel list that don't
belong to a certain group or that are no longer activated. Not really
sure, it's possible they should re-join the group if they are no longer
suspended.

### DIFF
--- a/app/jobs/scheduled/update_user_counts_for_chat_channels.rb
+++ b/app/jobs/scheduled/update_user_counts_for_chat_channels.rb
@@ -15,8 +15,8 @@ module Jobs
       new_count = chat_channel
         .user_chat_channel_memberships
         .joins(:user)
-        .where(user: { active: true, suspended_till: nil, staged: false })
         .where(following: true)
+        .merge(User.activated.not_suspended.not_staged)
         .count
 
       return if current_count == new_count
--- a/app/jobs/scheduled/update_user_counts_for_chat_channels.rb
+++ b/app/jobs/scheduled/update_user_counts_for_chat_channels.rb
@@ -15,7 +15,7 @@ module Jobs
       new_count = chat_channel
         .user_chat_channel_memberships
         .joins(:user)
-        .where(user: {active: true, suspended_till: nil, staged: false})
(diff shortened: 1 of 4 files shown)


app/jobs/scheduled/update_user_counts_for_chat_channels.rb
```diff
@@ -15,8 +15,8 @@ def set_user_count(chat_channel)
       new_count = chat_channel
         .user_chat_channel_memberships
         .joins(:user)
+        .where(user: { active: true, suspended_till: nil, staged: false })
```
martin-brennan, line 18: Hmm I think we should still use the `merge`s here to reuse scopes from core, we almost have all the ones we need:

https://github.com/discourse/discourse/blob/a258630790d7c976632d857aa9eaf6e09fd71d4e/app/models/user.rb#L221-L222

Would just need to add a `:not_staged` scope to core and use that as well. Alternatively, add a new scope that is `active_not_suspended` or something that does all 3,

spec/plugin_spec.rb
```diff
@@ -184,6 +188,12 @@ def topic_view
 
     context "regular" do
       it "renders channel" do
```
martin-brennan, line 190: I think we should update this description to say "renders channel, excluding inactive, staged, and suspended users"

plugin.rb
```diff
@@ -233,7 +233,7 @@ def self.onebox_template
 
     users = chat_channel
       .user_chat_channel_memberships
-      .includes(:user)
+      .includes(:user).where(user: { active: true, suspended_till: nil, staged: false })
```
martin-brennan, line 236: Same scopes comment here

app/jobs/scheduled/update_user_counts_for_chat_channels.rb
oblakeerickson: Okay, sounds good. I was wondering if we had an existing scope for this, I guess I'll create a new one then.

CvX (approved):

PR #914, opened 2022-06-07
FIX: Undefined uploads error when attempting to cloneJSON

martin-brennan:
Sometimes `uploads` are not present, which means that
the change from 0f4816c43e285d3e5e330742ff2a62e045db596b sometimes
errors (e.g. when inserting gifs via discourse-gifs). This
fixes the issue by providing default values. As in the
referenced commit, current coupling prevents this from being
easily testable.

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -703,7 +703,7 @@ export default Component.extend({
         message: data.chat_message.message,
         cooked: data.chat_message.cooked,
         excerpt: data.chat_message.excerpt,
-        uploads: cloneJSON(data.chat_message.uploads),
+        uploads: cloneJSON(data.chat_message.uploads || []),
         edited: true,
       });
     }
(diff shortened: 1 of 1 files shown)


martin-brennan:
Test fail unrelated:

```
Test Errors
----------------------------------------------
  Test Failed: Chat float open to DM channel with unread messages with sidebar off
    Assertion Failed: failed, expected argument to be truthy, was: false
      Expected: true, Actual: false
    at eval (acceptance/chat-test:972:21)
    at Object.eval (discourse/tests/helpers/qunit-helpers:383:9)
    at processModule (qunit:1192:16)
    at module$1 (qunit:1217:4)
    at acceptance (discourse/tests/helpers/qunit-helpers:379:25)
    at eval (acceptance/chat-test:907:32)
    at obj (discourse-loader:[371](https://github.com/discourse/discourse-chat/runs/6766097456?check_suite_focus=true#step:23:372):31)
```

CvX (approved):

PR #915, opened 2022-06-07
FIX: Make sure chat uploads have correct URL in template

martin-brennan:
This commit fixes non-image chat uploads not working since
e9cb7a172dfd1f413626a4590e7f373447370086 when we removed
the `resolveAllShortUrls` calls in the chat message decorator. We can
just use the `upload.url` directly on the `href`.

`data-orig-X` attributes must remain, as they are used for quoting.


martin-brennan:
Failing test seems to be an unrelated flaky.

jjaffeux (approved):

PR #917, opened 2022-06-08
DEV: correct silence / silience typo

SamSaffron:
Note, not migrating site setting, accepting risk since we are in early beta.


CvX (approved):
What a _sili_ typo 🤪

PR #922, opened 2022-06-08
DEV: Fix flakey chat window tests

oblakeerickson:
Reset chat window state seems to help in these tests. They are passing
locally now at least.

### DIFF
--- a/test/javascripts/unit/lib/chat-window-store-test.js
+++ b/test/javascripts/unit/lib/chat-window-store-test.js
@@ -4,6 +4,7 @@ import { test } from "qunit";
 discourseModule("Discourse Chat | Unit | chat-window-store", function (hooks) {
   hooks.beforeEach(function () {
     this.chatWindowStore = this.container.lookup("service:chat-window-store");
+    this.chatWindowStore.set("fullPage", false);
   });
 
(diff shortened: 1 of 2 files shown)


test/javascripts/unit/lib/chat-window-store-test.js
```diff
@@ -4,6 +4,7 @@ import { test } from "qunit";
 discourseModule("Discourse Chat | Unit | chat-window-store", function (hooks) {
   hooks.beforeEach(function () {
     this.chatWindowStore = this.container.lookup("service:chat-window-store");
+    this.chatWindowStore.set("fullPage", false);
```
tgxworld, line 7: Usually, we want to avoid cleaning up state in a `beforeEach` callback since it does not address the problem where certain tests are leaking state. Will moving this to the `afterEach` callback resolve the flakiness seen here?

oblakeerickson: Ah okay. Let's try and see.

oblakeerickson: Ah okay. Let's try and see.

tgxworld (approved):

PR #928, opened 2022-06-13
FIX: N1 in chat channels controller index querying chat_channel_archives

martin-brennan:
We were getting N1 queries with the archive record for
a channel when listing all public channels. This is mitigated
by:

  a) only trying to query for the archive record if the channel
     status is archived and
  b) including the channel archive record when querying all public
     chat channels

### DIFF
--- a/app/serializers/chat_channel_serializer.rb
+++ b/app/serializers/chat_channel_serializer.rb
@@ -54,7 +54,7 @@ class ChatChannelSerializer < ApplicationSerializer
   end
 
   def include_archive_status?
-    scope.is_staff? && archive.present?
+    scope.is_staff? && object.archived? && archive.present?
   end
 
   def archive_completed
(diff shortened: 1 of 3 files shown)


app/serializers/chat_channel_serializer.rb
```diff
@@ -50,11 +50,11 @@ def chatable
   end
 
   def archive
-    object.chat_channel_archive
+    @archive ||= object.chat_channel_archive
```
tgxworld, line 53: The memoization here is not required since AR record memoizes the association internally once loaded.

tgxworld (approved):
Looks good to me :+1:

app/serializers/chat_channel_serializer.rb
martin-brennan: Yeah good point, wasn't totally sure so I threw it in anyway, will remove

PR #929, opened 2022-06-13
DEV: creates membership when reacting

jjaffeux:
With this change reacting to a preview channel or a draft channel, will create a membership, the reaction and then navigate to the actual channel.

To make this change easier/cleaner I also made two other changes:
- rely on `chat.activeChannel` instead of recreating  an intermediate variable in full-page-chat and topic-chat-float
- simplify the `switchChannel` code, I suspect it might cause few regressions, but those are not tested ATM and I will fix/tests them as they come up

It also creates extensive specs for ChatMessageReactor which had none.


ZogStriP (approved):
LGTM 👍

PR #930, opened 2022-06-14
FIX: Chat quote interfering with oneboxes

martin-brennan:
A oneline change. Chat quotes render their inner content with the
chat markdown engine with a limited subset of rules, using a
html_raw markdown-it token. However when I first added this I
pushed the token to the state with the wrong level. I had this:

```javascript
// rendering chat message content with limited markdown rule subset
const token = state.push("html_raw", "", 0);
token.content = customMarkdownCookFn(content);
state.push("html_raw", "", -1);
```

But needed this:

```javascript
// rendering chat message content with limited markdown rule subset
const token = state.push("html_raw", "", 1);
token.content = customMarkdownCookFn(content);
state.push("html_raw", "", -1);
```

Spot the difference? (I simply changed the 0 to 1 on the first line).
This was causing the token "level" to go into negatives, which then
broke our onebox.js logic in core. If the previous token level is 0,
then we do a big onebox, otherwise we do an inline onebox because
we assume the link is lower in the HTML heirarchy:

https://github.com/discourse/discourse/blob/0ae7b43018238a883326da950f1e3bbf7e1f87a6/app/assets/javascripts/pretty-text/engines/discourse-markdown/onebox.js#L21-L22

So the level was -1 after a chat quote since the html_raw token
was pushed incorrectly, which meant that any oneboxes after the quote
would render inline.

### DIFF
--- a/spec/integration/post_chat_quote_spec.rb
+++ b/spec/integration/post_chat_quote_spec.rb
@@ -139,4 +139,65 @@ describe "chat bbcode quoting in posts" do
       </div>
     COOKED
   end
+
+  it "correctly renders inline and non-inline oneboxes combined with chat quotes" do
+    full_onebox_html = <<~HTML.chomp
+      <aside class="onebox wikipedia" data-onebox-src="https://en.wikipedia.org/wiki/Hyperlink" dir="ltr">
+        <header class="source">
+          <svg class="fa d-icon d-icon-fab-wikipedia-w svg-icon svg-string" xmlns="http://www.w3.org/2000/svg">
+            <use href="#fab-wikipedia-w"></use>
+          </svg>
+          <a href="https://en.wikipedia.org/wiki/Hyperlink" target="_blank" rel="nofollow ugc noopener" tabindex="-1">en.wikipedia.org</a>
+        </header>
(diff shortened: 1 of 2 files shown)


spec/integration/post_chat_quote_spec.rb
```diff
@@ -139,4 +139,65 @@
       </div>
     COOKED
   end
+
+  it "correctly renders inline and non-inline oneboxes combined with chat quotes" do
```
jjaffeux, line 143: does it need such a complex onebox to demonstrate the issue?

jjaffeux: if possible I would write the minimum length possible code that would reproduce. Less to maintain in the future.

jjaffeux (approved):
Thanks for fixing it, if you can address the comment easily please do it, otherwise let's just merge this.

spec/integration/post_chat_quote_spec.rb
martin-brennan: Yes true, I ripped out the extra data from the example onebox, I had just copied it from the problem case internally.

PR #937, opened 2022-06-15
DEV: allows to refresh only notifications and not full page

jjaffeux:
- introduces chat#refreshTrackingState
- makes each tracking state an EmberObject to make computation easier
- reset lastForcedRefreshAt when leaving tab


pmusaraj (commented):
This seems like a good change, though there are test/linting issues. 

Does "allows to refresh only notifications and not full page" have user-facing benefits?

pmusaraj (approved):

PR #946, opened 2022-06-21
FIX: do not list bookmarks of trashed messages

jjaffeux:
Messages are trashed and not destroyed, as a result `dependent: :destroy` has no effect on bookmarks and we have to make sure we don't attempt to list these bookmarks.

Before this fix the following repro would cause a 500 when reaching for `/u/:username/bookmarks.json`:

- Open a DM with any user
- Send a message
- Bookmark the message
- Delete the message

```
NoMethodError at /u/tomtom/bookmarks.json
=========================================

undefined method `chat_channel' for nil:NilClass

plugins/discourse-chat/app/serializers/user_chat_message_bookmark_serializer.rb, line 11
----------------------------------------------------------------------------------------

    6     def title
    7       fancy_title
    8     end
    9
   10     def fancy_title
>  11       @fancy_title ||= chat_message.chat_channel.title(scope.user)
   12     end
   13
   14     def cooked
   15       chat_message.cooked
   16     end
```

### DIFF
--- a/lib/chat_message_bookmarkable.rb
+++ b/lib/chat_message_bookmarkable.rb
@@ -19,6 +19,7 @@ class ChatMessageBookmarkable < BaseBookmarkable
     user.bookmarks_of_type("ChatMessage")
       .joins(
         "INNER JOIN chat_messages ON chat_messages.id = bookmarks.bookmarkable_id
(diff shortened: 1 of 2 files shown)


nattsw (approved):
🚀

PR #948, opened 2022-06-23
FIX: Clean up deleted chat message bookmarks

martin-brennan:
The Bookmark.cleanup! method is called on a schedule in
core, and calls the cleanup_deleted method on each
registered bookmarkable. This commit adds that method
to ChatMessageBookmarkable, and also a missing validate_before_create
method to the bookmarkable to ensure the bookmark can be
created for the message.


martin-brennan:
Unrelated frontend test failures.

jjaffeux (approved):
LGTM thanks ❤️

PR #954, opened 2022-06-25
FIX: Typo in the chat_channel subject translation

romanrizzi:


romanrizzi:
Failure is unrelated. Same tests are also failing on main.

CvX (approved):

PR #959, opened 2022-06-28
FIX: The create channel's security settings link should work with nested categories.

romanrizzi:


jjaffeux (approved):

jjaffeux:
linting needs fixing please

PR #971, opened 2022-06-29
UX: Improve category hints when creating a channel.

romanrizzi:
We'll display which groups will have access to the channel as part of the hint when choosing a category. The hints route is a staff-only route.

https://user-images.githubusercontent.com/5025816/176479132-52aa69a0-b399-4327-b6da-1f9efc79149b.mov


assets/javascripts/discourse/controllers/create-channel-modal.js
```diff
@@ -18,6 +18,11 @@ export default Controller.extend(ModalFunctionality, {
   categoryId: null,
   name: "",
   description: "",
+  categoryHintText: null,
+
+  onShow() {
+    this.categoryHintText = this._defaultHint();
```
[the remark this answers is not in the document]
jjaffeux: also it should probably be using set

assets/javascripts/discourse/controllers/create-channel-modal.js
```diff
@@ -126,4 +116,38 @@ export default Controller.extend(ModalFunctionality, {
...
+  },
+
+  _updateCategoryHint(category) {
+    if (category) {
+      const fullSlug = this._buildCategorySlug(category);
+
+      return ajax(`/chat/hints/${category.id}.json`).then((response) => {
+        const groupHints = response.hints;
+
+        if (groupHints && groupHints.length > 0) {
```
jjaffeux, line 134:
```suggestion
        if (groupHints?.length > 0) {
```

app/controllers/hints_controller.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class DiscourseChat::HintsController < ApplicationController
+  def access_by_category
+    hints = Group
+      .joins(:category_groups)
+      .where(category_groups: { category_id: params[:category_id] })
+      .pluck(:name).map! { |name| "@#{name}" }
```
jjaffeux, line 8: why map! and not map here ?

romanrizzi: `map!` mutates the array while `map` returns a copy. No difference besides that.

If you prefer map, I can change it.

assets/javascripts/discourse/controllers/create-channel-modal.js
jjaffeux, line 24: Shouldn't this use set ?

plugin.rb
```diff
@@ -533,6 +534,9 @@ def self.onebox_template
     put '/chat_channels/:chat_channel_id/change_status' => 'chat_channels#change_status'
     delete '/chat_channels/:chat_channel_id' => 'chat_channels#destroy'
 
+    # hints controller. Only used by staff members, we don't want to leak category permissions.
+    get '/hints/:category_id' => 'hints#access_by_category', format: :json, constraints: StaffConstraint.new
```
jjaffeux, line 538: I have been thinking quite a long time about this and I think the way forward is:

(you can use chat_channels for now, I want to go dash for everything in the future)

- POST/PUT /chat/api/draft-chat-channels/
- which return a draft id, a chatable type, chatable id, name, description, and hints in a meta key
- when you choose a category for the first time it creates the draft-channel, if you change the category (and have a draft id) you PUT it to get the new hints

In the future everything will be in the same screen (creating DM and creating public channels) and we will have these URLs: /chat/draft-channels/new and /chat/draft-channels/:id/edit

Let me know what you think of this plan

romanrizzi: I understand the benefit of having a draft for a topic or post, where you don't want to lose what you wrote or pause and later come back at it, but I don't see much value in having the same for channels. Maybe I'm missing something, but creating a channel should be quick and easy, even if you lost the form values and have to start over.

app/controllers/api/category_chatables_controller.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class DiscourseChat::Api::CategoryChatablesController < ApplicationController
+  def permissions
+    permissions = Group
+      .joins(:category_groups)
+      .where(category_groups: { category_id: params[:id] })
+      .pluck(:name).map! { |name| "@#{name}" }
+
+    render json: { permissions: permissions }
```
jjaffeux, line 10: do we need the root key here?

assets/javascripts/discourse/controllers/create-channel.js
```diff
@@ -2,47 +2,39 @@ import Controller from "@ember/controller";
...
 export default Controller.extend(ModalFunctionality, {
   chat: service(),
   category: null,
   categoryId: null,
   name: "",
   description: "",
   categorySelected: notEmpty("category"),
+  categoryHintText: null,
+
+  onShow() {
```
jjaffeux, line 23: do you reset it on close somewhere?

assets/javascripts/discourse/controllers/create-channel.js
```diff
@@ -93,4 +86,38 @@ export default Controller.extend(ModalFunctionality, {
       return category.slug;
     }
   },
+
+  _defaultHint() {
```
jjaffeux, line 90: Maybe this could be a const at top of file

assets/javascripts/discourse/templates/modal/create-channel.hbs
```diff
@@ -8,15 +8,11 @@
...
-  <div class="create-channel-hint">
-    {{html-safe
-      (i18n
-        "chat.create_channel.choose_category.hint"
-        link=categoryHint.link
-        category=categoryHint.category
-      )
-    }}
-  </div>
+  {{#if categoryHintText}}
```
jjaffeux, line 11: now that we name it permissions in API, we should maybe reflect the change here?

test/javascripts/acceptance/create-channel-test.js
```diff
@@ -125,12 +129,33 @@ acceptance("Discourse Chat - Create channel modal", function (needs) {
...
+    await click(".edit-channels-dropdown .select-kit-header-wrapper");
+    await click("li[data-value='openCreateChannelModal']");
+
+    assert.strictEqual(
+      query(".create-channel-hint a").innerText,
+      "category security settings"
+    );
+    assert.ok(query(".create-channel-hint a").href.includes("/categories"));
+
+    await click(".category-chooser .select-kit-header-wrapper");
```
jjaffeux, line 153: you have a selectKit API for this look for `selectKit()` in core codebase

assets/javascripts/discourse/controllers/create-channel.js
```diff
@@ -93,4 +86,38 @@ export default Controller.extend(ModalFunctionality, {
...
+      link: "/categories",
+      category: "category",
+    });
+  },
+
+  _updateCategoryHint(category) {
+    if (category) {
+      const fullSlug = this._buildCategorySlug(category);
+
+      return ajax(`/chat/api/category-chatables/${category.id}/permissions.json`).then((response) => {
```
jjaffeux, line 101: could we add this in ChatApiClient please?

assets/javascripts/discourse/controllers/create-channel.js
romanrizzi: Yep. See line 76.

spec/requests/api/category_chatables_controller_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe DiscourseChat::Api::CategoryChatablesController do
+  describe '#access_by_category' do
+    fab!(:admin) { Fabricate(:admin) }
```
jjaffeux, line 7: I would move this in as admin as we only use it here

spec/requests/api/category_chatables_controller_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      it "returns an empty list when the category doesn't have group permissions" do
+        category_2 = Fabricate(:category)
+
+        get "/chat/api/category-chatables/#{category_2.id}/permissions.json"
+
+        expect(response.parsed_body).to be_empty
+      end
+    end
+
```
jjaffeux, line 37: maybe we need regular user check too?

assets/javascripts/discourse/controllers/create-channel.js
```diff
@@ -93,4 +94,38 @@ export default Controller.extend(ModalFunctionality, {
       return category.slug;
     }
   },
+
+  _updatePermissionsHint(category) {
+    if (category) {
+      const fullSlug = this._buildCategorySlug(category);
+
+      return ChatApi.categoryPermissions(category.id).then((response) => {
```
jjaffeux, line 102: just replace response by groupHints here maybe ?

PR #973, opened 2022-06-29
FIX: correctly displays unread indicator in titles

jjaffeux:


CvX (approved):

jjaffeux:
Screenshot 2022-06-29 at 18 26 29
Screenshot 2022-06-29 at 18 27 36
Screenshot 2022-06-29 at 18 25 41

CvX:
Hm, so we're back to the design where indicators are next to labels? 😃

iirc, it was like that originally. Then we decided it's too close to avatars (that use green as status indicators) so we moved them all the way to the right (+ a margin). We're reverting that change? cc: @discourse/designers (gotta try the new team mention 😛)

jordanvidrine:
> Hm, so we're back to the design where indicators are next to labels? 😃
> 
> 
> 
> iirc, it was like that originally. Then we decided it's too close to avatars (that use green as status indicators) so we moved them all the way to the right (+ a margin). We're reverting that change? cc: @discourse/designers (gotta try the new team mention 😛)

I'm not seeing any change in the placement of the green/blue unread indicator here.

jjaffeux:
really ? I tried how it was before my PR and it was next to label, I guess it was not  in non-sidebar... what a hell...

jjaffeux:
Really we can't have such big difference between sidebas/non sidebars, it makes 0 sense

PR #974, opened 2022-06-30
DEV: drops isolated mode

jjaffeux:
Note this commit also simplifies the code to detect if currently browsing on full page chat or not

### DIFF
--- a/assets/javascripts/discourse/templates/preferences/chat.hbs
+++ b/assets/javascripts/discourse/templates/preferences/chat.hbs
@@ -28,18 +28,6 @@
 <div class="control-group chat-setting">
   <label class="controls">
     {{input
-      id="user_chat_isolated"
-      type="checkbox"
-      checked=model.user_option.chat_isolated
-    }}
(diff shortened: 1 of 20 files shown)


assets/javascripts/discourse/templates/preferences/chat.hbs
```diff
@@ -25,18 +25,6 @@
...
-<div class="control-group chat-setting">
-  <label class="controls">
-    {{input
-      id="user_chat_isolated"
-      type="checkbox"
-      checked=model.user_option.chat_isolated
-    }}
-    {{i18n "chat.isolate.title"}}
-  </label>
-  <span class="control-instructions">{{i18n "chat.isolate.description"}}</span>
```
gschlager, line 37: You might want to remove those strings from the locale file as well.

jjaffeux: done

gschlager (approved):
LGTM

PR #977, opened 2022-07-01
FEATURE: integrate chat-sidebar to core-sidebar

lis2:
Mount channels section into core-sidebar via API

Related PR - https://github.com/discourse/discourse/pull/17296


assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -129,6 +132,82 @@ export default {
           }
         });
       });
+      api.addSidebarSection((BaseSectionHeader, BaseSectionLink) => {
+        return class extends BaseSectionHeader {
+          @tracked sectionLinks = A([]);
+
+          constructor() {
+            super(...arguments);
+            this.chatService = container.lookup("service:chat");
```
tgxworld, line 141: If this is a service, you can do something like `@service chat` instead of having to look it up.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -129,6 +132,82 @@ export default {
...
+                        channelTitle: channel.title,
+                      };
+                    }
+                    get title() {
+                      return channel.title;
+                    }
+                    get text() {
+                      return channel.title;
+                    }
+                  })()
```
tgxworld, line 165: We don't really need to define the class over and over again here. We can create a generic class and pass the `channel` in when initialising the class.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -124,6 +127,156 @@ export default {
...
+
+            get name() {
+              return this.channel.chatable_id;
+            }
+
+            get route() {
+              return "chat.channel";
+            }
+
+            get model() {
```
jjaffeux, line 146: Couldn't we just pass channel here ?

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -124,6 +127,156 @@ export default {
...
+                  } else {
+                    channel.unread_count = trackingState.unread_count;
+                    channel.unread_mentions = trackingState.unread_mentions;
+                  }
+                  newSectionLinks.push(new SidebarChatSectionLink({ channel }));
+                });
+              });
+              this.sectionLinks = newSectionLinks;
+            }
+
```
jjaffeux, line 232: Not sure name/title/text are descriptive enough, have a hard time understanding what it would be from here. Also why are they all the same "chat.chat_channels" ?

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -124,6 +127,156 @@ export default {
...
+              return this.channel.unread_count > 0 && "circle";
+            }
+
+            get suffixCssClass() {
+              return this.channel.unread_mentions > 0 ? "urgent" : "unread";
+            }
+          };
+
+          const SidebarChatSection = class extends BaseCustomSidebarSection {
+            @tracked sectionLinks = A([]);
```
jjaffeux, line 184: What is `A()` bringing here?

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -124,6 +127,156 @@ export default {
...
+
+            constructor({ sidebar }) {
+              super(...arguments);
+
+              this.sidebar = sidebar;
+              this.chatService = container.lookup("service:chat");
```
jjaffeux, line 190: We can't use @service here ? And if we can't shouldn't we load our own appEvents ?

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -124,6 +127,156 @@ export default {
...
+              this.sidebar.appEvents.on(
+                "chat:refresh-channels",
+                this._refreshChannels
+              );
+              this.sidebar.appEvents.on(
+                "chat:navigated-to-full-page",
+                this._refreshChannels
+              );
+              this.currentUser = getOwner(this).lookup("current-user:main");
```
jjaffeux, line 199: We should use the same pattern with container as above

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -124,6 +127,156 @@ export default {
...
+                this._refreshChannels
+              );
+            }
+
+            @bind
+            _refreshChannels() {
+              const newSectionLinks = [];
+              this.chatService.getChannels().then((channels) => {
+                channels.publicChannels.forEach((channel) => {
+                  const trackingState =
```
jjaffeux, line 218: Im not sure why we need to have this logic here? This should probably be something we only do in core-chat when we refresh channels. Also channel is an ember object (should be at least) we shouldn't set properties like a POJO.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -124,6 +127,156 @@ export default {
...
+            get title() {
+              return I18n.t("chat.chat_channels");
+            }
+
+            get text() {
+              return I18n.t("chat.chat_channels");
+            }
+
+            get actions() {
+              const currentUser = getOwner(this).lookup("current-user:main");
```
jjaffeux, line 246: we already did init this.current user in constructor I think

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -124,6 +127,156 @@ export default {
...
+              const actions = [
```
jjaffeux, line 247: It feels slightly odd in the API that everything is a class and then suddenly it becomes a class POJO

assets/javascripts/discourse/services/chat.js
```diff
@@ -650,6 +650,7 @@ export default Service.extend({
         dmChatChannel.set("last_message_sent_at", new Date());
         this.reSortDirectMessageChannels();
       }
+      this.appEvents.trigger("chat:refresh-channels");
```
jjaffeux, line 653: Could you elaborate more on these two please?

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -124,6 +127,156 @@ export default {
...
+
+            get route() {
+              return "chat.channel";
+            }
+
+            get model() {
+              return {
+                chatChannel: this.channel,
+                channelId: this.channel.id,
+                channelTitle: this.channel.chatable.slug,
```
jjaffeux, line 150: where is this used?

assets/javascripts/discourse/connectors/after-sidebar/chat-sidebar.hbs
```diff
@@ -1 +1 @@
-{{channels-list onSelect=(action "switchChannel")}}
+{{channels-list onSelect=(action "switchChannel") hidePublicChannels="true"}}
```
jjaffeux, line 1: Can we use a boolean please?

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -124,6 +127,156 @@ export default {
...
+              super(...arguments);
+
+              this.sidebar = sidebar;
+              this.chatService = container.lookup("service:chat");
+              this.sidebar.appEvents.on(
+                "chat:refresh-channels",
+                this._refreshChannels
+              );
+              this.sidebar.appEvents.on(
+                "chat:navigated-to-full-page",
```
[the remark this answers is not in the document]
lis2: This was added to update state of sidebar. When you click on a link, it is just making transition. Goal here was to call _refereshChannels so when you click link, an indicator that you have unread messages will be removed. Do you have a better approach in mind?

assets/javascripts/discourse/initializers/chat-setup.js
lis2: Yeah, it feels weird. It is because the route required `channelId` and `channelTitle` 

https://github.com/discourse/discourse-chat/blob/main/assets/javascripts/discourse/chat-route-map.js#L5

And after model requires `chatChannel` to set active channel

https://github.com/discourse/discourse-chat/blob/main/assets/javascripts/discourse/routes/chat-channel.js#L41

assets/javascripts/discourse/initializers/chat-setup.js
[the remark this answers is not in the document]
jjaffeux: I see, not yet, but we will have one at some point 👍

assets/javascripts/discourse/initializers/chat-setup.js
jjaffeux: I see, I will soon be able to remove this afterModel

assets/javascripts/discourse/initializers/chat-setup.js
lis2: Service didn't work. Own appEvents is loaded 👍

assets/javascripts/discourse/initializers/chat-setup.js
lis2: I thought it was necessary to correctly track, but looks like `[]` work as well. Removed

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +127,157 @@ export default {
...
+            get prefixIcon() {
+              return "hashtag";
+            }
+
+            get prefixIconColor() {
+              return this.channel.chatable.color;
+            }
+
+            get prefixIconBadge() {
+              return this.channel.chatable.read_restricted && "lock";
```
tgxworld, line 173: The expected return value is a string here so if `read_restricted` is `false`, we might break that expectation.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +127,157 @@ export default {
...
+            }
+
+            get suffixIcon() {
+              return this.channel.unread_count > 0 && "circle";
```
tgxworld, line 177: Same as https://github.com/discourse/discourse-chat/pull/977/files#r919573671

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +127,157 @@ export default {
...
+            }
+          };
+
+          const SidebarChatSection = class extends BaseCustomSidebarSection {
+            @tracked sectionLinks = [];
+
+            constructor({ sidebar }) {
+              super(...arguments);
+
+              this.sidebar = sidebar;
```
tgxworld, line 191: This should not be required in the plugin and should be set by `BaseCustomSidebarSection` instead.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +127,157 @@ export default {
...
+              this.chatService = container.lookup("service:chat");
+              this.currentUser = container.lookup("current-user:main");
```
tgxworld, line 193: This should not be required since we can access the `currentUser` from `this.sidebar.currentUser`.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +127,157 @@ export default {
...
+              this.appEvents = container.lookup("service:appEvents");
```
tgxworld, line 194: Same as https://github.com/discourse/discourse-chat/pull/977/files#r919574279

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +127,157 @@ export default {
...
+              this.appEvents.on("chat:refresh-channels", this._refreshChannels);
+              this.appEvents.on(
+                "chat:navigated-to-full-page",
+                this._refreshChannels
+              );
+            }
+
+            willDestroy() {
+              this._super(...arguments);
```
tgxworld, line 203:
```suggestion
              super(...arguments);
```
```suggestion
              this._super(...arguments);
```

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +127,157 @@ export default {
...
+                    channel.unread_mentions = trackingState.unread_mentions;
+                  }
+                  newSectionLinks.push(new SidebarChatSectionLink({ channel }));
+                });
+              });
+              this.sectionLinks = newSectionLinks;
+            }
+
+            get name() {
+              return I18n.t("chat.chat_channels");
```
tgxworld, line 235: This is meant to be the name of the section so that we can target it with code. As such, it is not supposed to be translated.

assets/javascripts/discourse/initializers/chat-setup.js
lis2: I tried to follow this convention:
https://github.com/discourse/discourse/blob/main/app/assets/javascripts/discourse/app/templates/components/sidebar/topics-section.hbs#L12

- name - component identifier which just setup CSS class https://github.com/discourse/discourse/blob/main/app/assets/javascripts/discourse/app/templates/components/sidebar/section-link.hbs#L3
- title - text for link
- text - text for "alt" attribute

You are right that name was wrong, I changed it to `chat-channels`

I can rename those attributes for example:
name -> identifier
title
text -> alternativeText

In that case, I would like to change core to have some consistency. Does it sound like a good idea to you?

assets/javascripts/discourse/services/chat.js
lis2: This one is observing `this.messageBus.subscribe(`/chat/${channel.id}/new-messages`, (busData) => {`

When a new message is available in chat, I trigger refresh channels. It will update sidebar to show "unread" indicator

assets/javascripts/discourse/initializers/chat-setup.js
lis2: This is that one -> https://github.com/discourse/discourse-chat/pull/977#discussion_r919512212

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +127,159 @@ export default {
...
+      api.addSidebarSection(
+        (BaseCustomSidebarSection, BaseCustomSidebarSectionLink) => {
+          const SidebarChatSectionLink = class extends BaseCustomSidebarSectionLink {
+            constructor({ channel }) {
+              super(...arguments);
+              this.channel = channel;
+            }
+
+            get name() {
+              return this.channel.chatable_id;
```
tgxworld, line 141: I would recommend a combination of `chatable_type` and `chatable_id` here to properly differentiate between the links.

assets/javascripts/discourse/initializers/chat-setup.js
lis2: I fixed it, problem was that publicChannels `unread_mentions` and `unread_count` properties were getting out of sync from trackinState. Public channels had initial value and only trackingState was updated. I changed service to update those numbers as well. After that, I could completely drop this logic.

assets/javascripts/discourse/initializers/chat-setup.js
jjaffeux: If it's used for alt attribute, I would just call it `alt`. Im fine with this but I don't know everything about sidebar.

tgxworld: I wouldn't worry too much about this because the parent class is document and anyone that implements the parent's class interface will have to read the docs.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +129,371 @@ export default {
...
+            }
+          };
+
+          return SidebarChatSection;
+        }
+      );
+
+      api.addSidebarSection(
+        (BaseCustomSidebarSection, BaseCustomSidebarSectionLink) => {
+          const SidebarChatSectionLink = class extends BaseCustomSidebarSectionLink {
```
tgxworld, line 300: I think we can improve the constant name here to reflect the type of section link this is. Otherwise, it is going to be confusing trying to differentiate this sidebar section from the one above when they share the same constant name.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +129,371 @@ export default {
...
+            get hoverValue() {
+              return "times";
+            }
+
+            get hoverAction() {
+              return () => {
+                this.chatService.unfollowChannel(this.channel);
+              };
+            }
+            get hoverTitle() {
```
tgxworld, line 389:
```suggestion
            }

            get hoverTitle() {
```

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +129,372 @@ export default {
...
+                this._refreshPms
+              );
+              this.sidebar.appEvents.on(
+                "chat:navigated-to-full-page",
+                this._refreshPms
+              );
+              this._refreshPms();
+            }
+
+            teardown() {
```
tgxworld, line 421:
```suggestion
            willDestory() {
```

I removed the `teardown` function in https://github.com/discourse/discourse/pull/17550

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +129,372 @@ export default {
         });
       });
     });
+
+    withPluginApi("1.3.0", (api) => {
```
tgxworld, line 133: I think it'll also be nicer from an organisation perspective if we move the sidebar specific stuff into another initializer.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +129,372 @@ export default {
...
+                this._refreshChannels
+              );
+              this.chatService.appEvents.on(
+                "chat:navigated-to-full-page",
+                this._refreshChannels
+              );
+              this._refreshChannels();
+            }
+
+            teardown() {
```
tgxworld, line 224:
```suggestion
            willDestroy() {
```

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -125,6 +129,372 @@ export default {
...
+              return this.channel.unread_mentions > 0 ? "urgent" : "unread";
+            }
+
+            get currentWhen() {
+              return (
+                this.chatService.router.currentRouteName?.startsWith(
+                  "chat.channel"
+                ) && this.channel.id === this.chatService.activeChannel?.id
+              );
+            }
```
tgxworld, line 203: I'm not sure if we need this since the `LinkTo` component should automatically take care of adding the `.active` class to the link.

lis2: Sidebar link was not active when channel was changed via keyboard shortcut and switch channel modal. This is because switch channel is using params to transitionTo

https://github.com/discourse/discourse-chat/blob/main/assets/javascripts/discourse/services/chat.js#L462

Then active state is calculated here:

https://github.com/discourse/discourse-chat/blob/main/assets/javascripts/discourse/components/chat-channel-row.js#L23

In LinkTo we pass model and active was not correctly detected.

That's why I tried to use `currentWhen` with logic stolen from original sidebar and it solved problem

tgxworld: We need to move away from the hacks of the old sidebar. This means that any link present in the sidebar has to be a valid Ember route.

assets/stylesheets/sidebar-extensions.scss
```diff
@@ -142,6 +142,11 @@
       padding-right: 0;
     }
   }
+  .sidebar-section-chat-channels {
+    summary:hover svg.d-icon {
```
jjaffeux, line 146: It should have a class most likely

test/javascripts/acceptance/core-sidebar-test.js
```diff
@@ -0,0 +1,195 @@
...
+import { directMessageChannels } from "discourse/plugins/discourse-chat/chat-fixtures";
+
+acceptance("Discourse Chat - Core Sidebar", function (needs) {
+  needs.user({ experimental_sidebar_enabled: true, has_chat_enabled: true });
+  needs.settings({
+    chat_enabled: true,
+    enable_experimental_sidebar: true,
+  });
+  needs.pretender((server, helper) => {
+    let directChannels = directMessageChannels.mapBy("chat_channel");
```
tgxworld, line 18: Do we need to clone the fixture here to avoid mutating it?

test/javascripts/acceptance/core-sidebar-test.js
```diff
@@ -0,0 +1,195 @@
...
+  });
+
+  test("Public channels section", async function (assert) {
+    await visit("/");
+
+    assert.strictEqual(
+      query(
+        ".sidebar-section-chat-channels .sidebar-section-header-text"
+      ).textContent.trim(),
+      "Channels",
```
tgxworld, line 62:
```suggestion
      I18n.t("chat.chat_channels"),
```

test/javascripts/acceptance/core-sidebar-test.js
```diff
@@ -0,0 +1,195 @@
...
+      "displays correct channels section title"
+    );
+    const links = queryAll(".sidebar-section-chat-channels a");
```
tgxworld, line 65: Instead of targeting `a`, is there a more specific class we can target instead?

test/javascripts/acceptance/core-sidebar-test.js
```diff
@@ -0,0 +1,195 @@
...
+        .eq(0)
+        .find(".sidebar-section-link-prefix svg.prefix-badge")[0]
+        .classList.contains("d-icon-lock"),
+      true,
+      "displays lock badge for restricted channel"
+    );
+    assert.strictEqual(
+      links[0].textContent.trim(),
+      "dev",
+      "displays dev name in a link"
```
tgxworld, line 85:
```suggestion
      "displays channel name in the link"
```

test/javascripts/acceptance/core-sidebar-test.js
```diff
@@ -0,0 +1,195 @@
...
+  });
+
+  test("Direct messages section", async function (assert) {
+    await visit("/");
+
+    assert.strictEqual(
+      query(
+        ".sidebar-section-chat-dms .sidebar-section-header-text"
+      ).textContent.trim(),
+      "Personal chat",
```
tgxworld, line 149: Let's use the `I18n` function here instead so that the test is not brittle to translation changes.

assets/javascripts/discourse/initializers/chat-sidebar.js
```diff
@@ -0,0 +1,423 @@
...
+                      this.chatService.router.currentRouteName.startsWith("")
+                    ) {
+                      this.chatService.router.transitionTo(
+                        "chat.draft-channel"
+                      );
+                    } else {
+                      this.appEvents.trigger(
+                        "chat:open-view",
+                        DRAFT_CHANNEL_VIEW
+                      );
```
tgxworld, line 402: I think this logic is important to test as well.

assets/javascripts/discourse/initializers/chat-sidebar.js
```diff
@@ -0,0 +1,423 @@
...
+              super(...arguments);
+
+              if (container.isDestroyed) {
+                return;
+              }
+              this.chatService = container.lookup("service:chat");
+              this.sidebar.appEvents.on(
+                "chat:refresh-channels",
+                this._refreshPms
+              );
```
tgxworld, line 341: In the tests, I don't think we tested this logic here.

assets/javascripts/discourse/initializers/chat-sidebar.js
```diff
@@ -0,0 +1,423 @@
...
+                actions.push({
+                  id: "openCreateChannelModal",
+                  title: I18n.t("chat.channels_list_popup.create"),
+                  action: () => {
+                    showModal("create-channel");
+                  },
+                });
+              }
+              return actions;
+            }
```
tgxworld, line 188: I would test these actions too.

tgxworld (approved):
@lis2 The PR looks good to me and I'm happy for you to merge it as it is. However, I do think we can improve the test coverage further but I think we can follow up in another PR.

PR #983, opened 2022-07-01
FIX: correctly handles chat-channel-row active state

jjaffeux:


CvX (approved):
(what's with the tests failing? does the presence stuff need to be stubbed out?)

CvX (approved):

PR #988, opened 2022-07-04
FEATURE: Automatically add users to a channel.

romanrizzi:
When creating a category channel, you have the option to configure it to add users automatically, which means all group members that have access to that category will join. Also, you can later enable or disable this option when
editing a channel.

Additionally, users will join the channel after gaining access to the category, or removed after losing it.

Screen Shot 2022-07-06 at 10 51 51
Screen Shot 2022-07-06 at 10 51 20
Screen Shot 2022-07-06 at 10 51 39


app/controllers/chat_channels_controller.rb
```diff
@@ -81,9 +59,18 @@ def create
     chatable = chatable_type.constantize.find_by(id: params[:id])
     raise Discourse::NotFound unless chatable
 
-    chat_channel = ChatChannel.create!(chatable: chatable, name: params[:name], description: params[:description], user_count: 1)
+    auto_join_users = ActiveRecord::Type::Boolean.new.deserialize(params[:auto_join_users]) || false
+
+    chat_channel = ChatChannel.create!(
+      chatable: chatable, name: params[:name], description: params[:description],
```
jjaffeux, line 65: One key per row please, I think it's much more readable

assets/javascripts/discourse/controllers/create-channel.js
```diff
@@ -15,7 +14,7 @@ export default Controller.extend(ModalFunctionality, {
   categoryId: null,
   name: "",
   description: "",
-  categorySelected: notEmpty("category"),
```
jjaffeux, line 18: why is this removed?

app/controllers/api/chat_channels_controller.rb
```diff
@@ -13,13 +13,23 @@ def update
     end
 
     editable_params = params.permit(*CHAT_CHANNEL_EDITABLE_PARAMS)
+    editable_params.delete(:auto_join_users) if !chat_channel.category_channel?
     editable_params.each do |k, v|
       editable_params[k] = nil if editable_params[k].blank?
     end
+
+    if ActiveRecord::Type::Boolean.new.deserialize(editable_params[:auto_join_users])
```
jjaffeux, line 21: How about we just cancel the previous jobs ?

app/jobs/regular/auto_manage_channel_memberships.rb
```diff
@@ -0,0 +1,78 @@
...
+    REMOVE = 'remove'
+
+    def execute(args)
+      channel = ChatChannel.find_by(id: args[:chat_channel_id])
+      mode = args[:mode]
+
+      return if channel.nil?
+      return if !channel.category_channel?
+      return if !channel.auto_join_users?
+      return if ![JOIN, REMOVE].include?(mode)
```
jjaffeux, line 15: maybe we do this check first?

assets/javascripts/discourse/controllers/create-channel.js
romanrizzi: Accidentally removed it during rebase

assets/javascripts/discourse/templates/modal/create-channel-modal.hbs
```diff
@@ -0,0 +1,92 @@
+{{#d-modal-body title="chat.create_channel.title"}}
```
jjaffeux, line 1: `create-channel` please, otherwise it will generate `-modal-modal` as class name

assets/javascripts/discourse/templates/modal/create-channel.hbs
```diff
@@ -18,6 +18,11 @@
     }}
   </div>
 
+  <label class="create-channel-label">
+    {{input type="checkbox" checked=autoJoinChannel}}
+    {{i18n "chat.settings.enable_auto_join_users"}}
```
jjaffeux, line 23:
```suggestion
    <span>{{i18n "chat.settings.enable_auto_join_users"}}</span>
```

app/controllers/api/chat_channels_controller.rb
```diff
@@ -13,13 +13,23 @@ def update
     end
 
     editable_params = params.permit(*CHAT_CHANNEL_EDITABLE_PARAMS)
+    editable_params.delete(:auto_join_users) if !chat_channel.category_channel?
```
jjaffeux, line 16: Im thinking maybe we go with:

```ruby
CHAT_CHANNEL_EDITABLE_PARAMS = [:name, :description]
CATEGORY_CHAT_CHANNEL_EDITABLE_PARAMS = [:auto_join_users]

def update
  # ...
  editable_params = editable_params(params, chat_channel)
  # ...
end

def editable_params(params, chat_channel)
  list = CHAT_CHANNEL_EDITABLE_PARAMS
  if chat_channel.category_channel?
    list = list + CATEGORY_CHAT_CHANNEL_EDITABLE_PARAMS
  end
  params.permit!(*list)
end
```

app/models/chat_channel.rb
```diff
@@ -27,6 +27,34 @@ class ChatChannel < ActiveRecord::Base
 
   validates :name, length: { maximum:  Proc.new { SiteSetting.max_topic_title_length } }, presence: true, allow_nil: true
 
+  def join(user)
```
jjaffeux, line 30: I you move this logic to `models/chat_channel.rb` there should be specs for this now

assets/javascripts/discourse/controllers/create-channel.js
```diff
@@ -60,6 +59,7 @@ export default Controller.extend(ModalFunctionality, {
       id: this.categoryId,
       name: this.name,
       description: this.description,
+      auto_join_users: this.autoJoinChannel,
```
jjaffeux, line 62: Shouldn't we keep the same naming pattern here? `this.autoJoinUsers` ?

assets/javascripts/discourse/templates/modal/create-channel-modal.hbs
```diff
@@ -0,0 +1,92 @@
...
+    <div class="create-channel-hint">
+      {{html-safe
+        (i18n
+          "chat.create_channel.choose_category.hint"
+          link=categoryHint.link
+          category=categoryHint.category
+        )
+      }}
+    </div>
+  {{else if (eq type "topic")}}
```
jjaffeux, line 31: should be removed

assets/javascripts/discourse/templates/modal/create-channel-modal.hbs
```diff
@@ -0,0 +1,92 @@
...
+  </label>
+  {{input
+    class="create-channel-description-input"
+    type="textarea"
+    value=description
+  }}
+
+  {{#if entitySelected}}
+    <div class="chat-channel-title">
+      {{#if (and (eq type "topic") topic)}}
```
jjaffeux, line 60: should be removed

spec/jobs/regular/auto_join_channel_batch_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+    memberships = UserChatChannelMembership.where(user: @user)
+    expect(memberships).to be_empty
+  end
+
+  it 'does nothing if the channel chatable is not a category' do
+    same_id = 99
+    another_category = Fabricate(:category, id: same_id)
+    another_cgroup = Fabricate(:category_group, category: another_category, group: chatters_group)
+    topic = Fabricate(:topic, id: same_id)
+    @channel.update!(chatable: topic)
```
jjaffeux, line 93: would probably be better to test with direct message channel as chatable now that topic are gone. I'm sorry you ended up in the middle of remove topic :/

app/controllers/api/chat_channels_controller.rb
jjaffeux, line 21: do we actually need to check it's a valid boolean here ? We might just check you attempted to change auto_join_users ?

app/jobs/regular/auto_join_channel_batch.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Jobs
+  class AutoJoinChannelBatch < ::Jobs::Base
+    def execute(args)
+      return "starts_at or ends_at missing" if args[:starts_at].blank? || args[:ends_at].blank?
+      return "End is higher than start" if args[:ends_at] < args[:starts_at]
+
+      channel = ChatChannel.find_by(id: args[:chat_channel_id])
```
jjaffeux, line 9: We can probably just do `ChatChannel.find(...)`  and remove next line, given we want to fail on not found

jjaffeux: Or maybe even better: `ChatChannel.find_by!(id: ..., chatable_type: 'Category', auto_join_users: true)`   ?

That way we can remove all the checks after

app/jobs/regular/auto_remove_channel_batch.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Jobs
+  class AutoRemoveChannelBatch < ::Jobs::Base
+    def execute(args)
+      return "starts_at or ends_at missing" if args[:starts_at].blank? || args[:ends_at].blank?
+      return "End is higher than start" if args[:ends_at] < args[:starts_at]
+
+      channel = ChatChannel.find_by(id: args[:chat_channel_id])
```
jjaffeux, line 9: Same comment than two others about find_by!

assets/javascripts/discourse/templates/modal/create-channel-modal.hbs
```diff
@@ -0,0 +1,92 @@
...
+        <span class="topic-chat-badge">
+          {{d-icon "far-comments"}}
+        </span>
+        <span class="topic-chat-name">{{replace-emoji name}}</span>
+      {{else if (and (eq type "category") category)}}
+        <span
```
jjaffeux, line 66: I think this is removed in the template now

assets/javascripts/discourse/templates/modal/create-channel-modal.hbs
jjaffeux: You probably want to double check this, I think there's been multiple issues in the merge

app/jobs/regular/auto_remove_channel_batch.rb
romanrizzi: Answered above!

app/controllers/api/chat_channels_controller.rb
romanrizzi: Updating the `auto_join_users` attribute will make all batches to return early, so there's no need to cancel.

Also, we don't queue the batch jobs directly, but queue a different job that does it for us. It means we'll introduce a race condition when we attempt to cancel them. I'm ok with adjusting the rate limiter if it's too restrictive

romanrizzi: We only want to perform the rate limiting when enabling the option because we trigger the auto-join then. When setting it to false, it's just a regular update, so it should be ok to bypass the rate limiter.

app/controllers/chat_channels_controller.rb
```diff
@@ -28,37 +28,15 @@ def show
...
-        .find_by!(
-          user_id: current_user.id,
-          chat_channel_id: params[:chat_channel_id]
-        )
-      membership.update!(following: false)
-      new_user_count = [(membership.chat_channel.user_count || 0) - 1, 0].max
-      membership.chat_channel.update!(user_count: new_user_count)
-      render_serialized(membership, UserChatChannelMembershipSerializer, root: false)
-    end
+    membership = @chat_channel.remove(current_user)
```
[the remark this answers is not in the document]
jjaffeux: Yes I agree and maybe I should have made my comment on the counterpart "join", it's weird that the opposite of join is remove.

app/jobs/regular/auto_join_channel_batch.rb
jjaffeux: `find_by!` will return super early, if you mean you just don't want this to create a not found error, you can just do `find_by` and then `return if !channel`

romanrizzi: Yeah, I meant rising an error `find_by` + `return if !channel` is the way to go IMO.

app/jobs/regular/auto_join_channel_batch.rb
```diff
@@ -0,0 +1,52 @@
...
+        chat_channel_id: channel.id,
+        start: args[:starts_at],
+        end: args[:ends_at],
+        suspended_until: Time.zone.now,
+        last_seen_at: 3.months.ago,
+        channel_category: channel.chatable_id,
+        mode: UserChatChannelMembership.join_modes[:automatic]
+      }
+
+      records_created = DB.exec(<<~SQL, query_args)
```
jjaffeux, line 27: Im curious how does the `ON CONFLICT` plays with this `records_created` count? Do we get a + 1 on conflict?

assets/javascripts/discourse/components/chat-channel-settings-view.js
```diff
@@ -84,6 +99,17 @@ export default class ChatChannelSettingsView extends Component {
     );
   }
 
+  @computed(
+    "siteSettings.max_chat_auto_joined_users",
```
jjaffeux, line 103: siteSettings is a POJO it has no effect in computed, you can use it directly in the function

app/jobs/regular/auto_manage_channel_memberships.rb
```diff
@@ -0,0 +1,61 @@
...
+        processed += batch.size
+      end
+    end
+
+    private
+
+    def auto_join_query(channel)
+      category = channel.chatable
+
+      users = User
```
jjaffeux, line 36: We should exclude system and bot from this

app/jobs/regular/auto_join_channel_batch.rb
romanrizzi: Good catch. Not sure, will take a look

romanrizzi: I opted for removing the `ON CONCLICT`. If the membership exists and following is false, it's because the user left the channel. It doesn't make sense to add them again.

jjaffeux: agreed

spec/plugin_spec.rb
```diff
@@ -221,4 +222,68 @@
...
+          chatters_group.add(user)
+          apply_permissions_and_trigger_event(another_category, chatters_group_permission)
+
+          membership = UserChatChannelMembership.find_by(user: user, chat_channel: @channel)
+
+          expect(membership).to be_nil
+        end
+      end
+
+      def apply_permissions_and_trigger_event(category, permissions)
```
jjaffeux, line 284: not sure it adds much value to have this as a separate function

spec/plugin_spec.rb
```diff
@@ -221,4 +222,68 @@
...
     end
   end
+
+  describe 'auto-joining users to a channel' do
+    fab!(:chatters_group) { Fabricate(:group) }
+    fab!(:category) { Fabricate(:category) }
+    fab!(:user) { Fabricate(:user, last_seen_at: 15.minutes.ago) }
+
+    before do
+      @channel = Fabricate(:chat_channel, auto_join_users: true, chatable: category)
```
jjaffeux, line 232: you forgot this ivar

config/locales/client.en.yml
```diff
@@ -210,15 +210,20 @@ en:
           other: "%{count} members"
 
       create_channel:
+        auto_join_users:
+          public_category_warning: "%{category} is a public category. Automatically add every user to this channel?"
```
jjaffeux, line 214: ?

```suggestion
          public_category_warning: "%{category} is a public category. Automatically add every users to this channel?"
```

app/jobs/regular/auto_manage_channel_memberships.rb
```diff
@@ -0,0 +1,62 @@
...
+        .real
```
jjaffeux, line 37: Maybe we also add: `.activated.not_suspended.not_staged`

jjaffeux (approved):

jjaffeux (approved):

jjaffeux:
Ok :heart: Let's try this tomorrow

spec/jobs/regular/auto_manage_channel_memberships_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+
+    it 'does nothing when the chatable is not a category' do
+      same_id = 99
+      another_category = Fabricate(:category, id: same_id)
+      dm_channel = Fabricate(:direct_message_channel, id: same_id)
+      channel.update!(chatable: dm_channel)
+
+      assert_batches_enqueued(channel, 0)
+    end
```
CvX, line 26: @romanrizzi do you remember why those two have to have the same id in this test? Specifying an id when creating a record makes this test flaky - you're never guaranteed that given id is available.

romanrizzi: It's to make sure the `chatable_type` is used when finding the channel we'll auto-join users to. We can remove it if it's causing problems.

PR #989, opened 2022-07-04
DEV: remove topic-enable

jjaffeux:


app/controllers/chat_channels_controller.rb
```diff
@@ -64,20 +64,19 @@ def unfollow
   def create
     params.require([:type, :id, :name])
     guardian.ensure_can_create_chat_channel!
-    raise Discourse::InvalidParameters unless ["topic", "category"].include?(params[:type].downcase)
+    raise Discourse::InvalidParameters unless ["category"].include?(params[:type].downcase)
```
ZogStriP, line 67: Could just be an == now that there's only 1 value

jjaffeux: Yes sure

db/post_migrate/20220701195731_convert_chatable_topics_to_categories.rb
```diff
@@ -0,0 +1,34 @@
...
+      WHERE cc.chatable_type = 'Topic'
+      SQL
+
+    # soft delete all posts small actions
+    DB.exec(
+      "UPDATE posts SET deleted_at = :deleted_at, deleted_by_id = :deleted_by WHERE action_code IN (:action_codes)",
+      action_codes: ['chat.enabled', 'chat.disabled'],
+      deleted_at: Time.zone.now,
+      deleted_by: Discourse.system_user
+    )
```
gschlager, line 22: I'd rather use an explicit ID instead of relying on Rails magic, but the current SQL works, so it's up to you. :man_shrugging: 

```suggestion
    DB.exec(
      "UPDATE posts SET deleted_at = :deleted_at, deleted_by_id = :deleted_by_id WHERE action_code IN (:action_codes)",
      action_codes: ['chat.enabled', 'chat.disabled'],
      deleted_at: Time.zone.now,
      deleted_by_id: Discourse::SYSTEM_USER_ID
    )
```

app/controllers/chat_channels_controller.rb
```diff
@@ -64,20 +64,19 @@ def unfollow
...
     existing_args = {
       chatable_type: chatable_type,
       chatable_id: params[:id]
     }
-    existing_args[:name] = params[:name] unless creating_topic_channel
+    existing_args[:name] = params[:name]
     exists = ChatChannel.exists?(existing_args)
 
     if exists
-      translation_key = creating_topic_channel ? "channel_exists_for_topic" : "channel_exists_for_category"
```
gschlager, line 80: Can the `channel_exists_for_topic` translation be removed from the locale file or is it used somewhere else?

jjaffeux: I think we can remove it yes

app/controllers/chat_channels_controller.rb
```diff
@@ -64,20 +64,19 @@ def unfollow
...
+      translation_key = "channel_exists_for_category"
       raise Discourse::InvalidParameters.new(I18n.t("chat.errors.#{translation_key}"))
```
gschlager, line 80:
```suggestion
      raise Discourse::InvalidParameters.new(I18n.t("chat.errors.channel_exists_for_category"))
```

gschlager (approved):
LGTM :+1:

app/controllers/chat_channels_controller.rb
```diff
@@ -64,21 +64,19 @@ def unfollow
...
 
-    creating_topic_channel = params[:type].downcase === "topic"
-    chatable_type = creating_topic_channel ? "Topic" : "Category"
+    chatable_type = "Category"
     existing_args = {
       chatable_type: chatable_type,
       chatable_id: params[:id]
     }
-    existing_args[:name] = params[:name] unless creating_topic_channel
+    existing_args[:name] = params[:name]
```
ZogStriP, line 75: This line could be moved into the `existing_args` declaration right above now that there aren't any conditions

app/controllers/chat_controller.rb
```diff
@@ -456,7 +447,6 @@ def preloaded_chat_message_query
 
   def find_chatable
     chatable_class = case params[:chatable_type].downcase
```
ZogStriP, line 449: Do we still need this `case`? 🤔

ZogStriP (approved):
LGTM 👍

PR #990, opened 2022-07-05
FIX: Allow user to direct message themself again and fix DM inconsistencies

martin-brennan:
At some time in the past few months (possibly #758)
the ability for users to create a new DM channel with themselves
was broken. There was inconsistency around the
chat direct message controller sometimes accepting
an array of usernames and sometimes accepting
a comma separated string of usernames. They are
both needed - the former for POST and the latter
for GET. The test for POST in ruby did not pass the
usernames as an array so it was not catching the
error which was preventing the creation of the channel,
which was caused by a duplication of the current user
in the users array.

This commit also changes the `create!` method argument
for `DirectMessageChannelCreator` to have a
`target_user` argument, which will make it easier to
add mute/ignore checks involving the user creating
the message and who they are targeting, since an `acting_user`
argument will be needed for that change.

It also addresses some inconsistencies with how the usernames
are passed to the controller from JS, and adds missing specs
for `DirectMessageChannelCreator`

### DIFF
--- a/app/controllers/direct_messages_controller.rb
+++ b/app/controllers/direct_messages_controller.rb
@@ -6,7 +6,7 @@ class DiscourseChat::DirectMessagesController < DiscourseChat::ChatBaseControlle
   def create
     guardian.ensure_can_chat!(current_user)
     users = users_from_usernames(current_user, params)
-    chat_channel = DiscourseChat::DirectMessageChannelCreator.create!(acting_user: current_user, target_users: users)
+    chat_channel = DiscourseChat::DirectMessageChannelCreator.create!(target_users: users)
     render_serialized(chat_channel, ChatChannelSerializer, root: "chat_channel")
   end
 
--- a/app/controllers/direct_messages_controller.rb
+++ b/app/controllers/direct_messages_controller.rb
@@ -6,7 +6,7 @@ class DiscourseChat::DirectMessagesController < DiscourseChat::ChatBaseControlle
   def create
     guardian.ensure_can_chat!(current_user)
     users = users_from_usernames(current_user, params)
-    chat_channel = DiscourseChat::DirectMessageChannelCreator.create!(users)
+    chat_channel = DiscourseChat::DirectMessageChannelCreator.create!(acting_user: current_user, target_users: users)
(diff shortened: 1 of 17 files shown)


martin-brennan:
Failing test is unrelated

app/controllers/direct_messages_controller.rb
```diff
@@ -31,9 +31,16 @@ def index
   def users_from_usernames(current_user, params)
     params.require(:usernames)
 
+    usernames = if params[:usernames].is_a?(String)
+      params[:usernames].split(",")
```
nattsw, line 35: 😢

app/controllers/direct_messages_controller.rb
```diff
@@ -31,9 +31,16 @@ def index
...
+    else
+      params[:usernames]
+    end
```
nattsw, line 38: 😢

lib/direct_message_channel_creator.rb
```diff
@@ -1,26 +1,25 @@
 # frozen_string_literal: true
 
 module DiscourseChat::DirectMessageChannelCreator
-  attr_reader :chat_channel, :users
-
-  def self.create!(users)
-    direct_messages_channel = DirectMessageChannel.for_user_ids(users.map(&:id).uniq)
+  def self.create!(acting_user:, target_users:)
```
nattsw, line 4: Hmm I don't really understand why we need acting_user in this PR? Don't see it being used anywhere.

martin-brennan: In general don't think it's an issue to add a new thing in a PR that you aren't explicitly using, as long as you are about to use it very soon (in my case I would use it tomorrow)...but I see your point about the `git blame` which I use all the time, so I will defer this change until I work on ignore/mute etc.

app/controllers/direct_messages_controller.rb
martin-brennan: Yeah not great but both actions are using this so not a lot we can do really.

martin-brennan:
@nattsw removed `acting_user`, kept `target_users` though since it is used and will make the next PR easier.

nattsw (approved):

PR #991, opened 2022-07-05
DEV: chat-user-displayname

jjaffeux:
Correctly encapsulates all the logic of a chat user display name inside a component. We might want to use it at more places in the future.


jjaffeux:
Test failure unrelated and being worked out separately.

test/javascripts/components/chat-user-displayname-test.js
```diff
@@ -0,0 +1,76 @@
...
+discourseModule(
+  "Discourse Chat | Component | chat-user-displayname | prioritize username in UX",
+  function (hooks) {
+    setupRenderingTest(hooks);
+
+    componentTest("username and no name", {
+      template: hbs`{{chat-user-displayname user=user}}`,
+
+      async beforeEach() {
+        set(this.siteSettings, "prioritize_username_in_ux", true);
```
CvX, line 21: `this.siteSettings.set()` doesn't work there?

CvX (approved):

test/javascripts/components/chat-user-displayname-test.js
jjaffeux: I don't know, haven't tried for a long time, I got used to follow this pattern for whatever reason, will check later :+1:

assets/javascripts/discourse/components/chat-user-displayname.js
```diff
@@ -0,0 +1,33 @@
+import Component from "@ember/component";
+import { computed } from "@ember/object";
+import { formatUsername } from "discourse/lib/utilities";
+
+export default class ChatUserDisplayname extends Component {
```
nattsw, line 5: Looks like it should be `ChatUserDisplayName` since display name is two words 🧐

assets/javascripts/discourse/components/chat-user-displayname.js
```diff
@@ -0,0 +1,33 @@
...
+  tagName = "";
+  user = null;
+
+  @computed
+  get shouldPrioritizeNameInUx() {
+    return !this.siteSettings.prioritize_username_in_ux;
```
nattsw, line 11: Oh no.. I was going to nit about the word `ux` then I realised this site setting isn’t chat specific.

nattsw (approved):
This is very nice

test/javascripts/components/chat-user-displayname-test.js
jjaffeux: We talked of this with cvx and apparently during the tests siteSettings becomes a POJO making this not possible, we should fix it in core.

PR #995, opened 2022-07-06
FIX: Use as_json instead of to_json for chat DM message bus

martin-brennan:
Follow up to f1fb1d0b4843b6265a6f46d8b24c31930a69c015,
in this commit I accidentally changed as_json (which produces
a Hash) to to_json (which produces a JSON string) in the
ChatPublisher for the new DM channel. This makes the MessageBus
data unparseable, so in JS in the chat service it ends up creating
a public channel in the channel list with undefined data, leaving
a blank channel in the list with a blue unread indicator.

Before fix MessageBus data:

After fix MessageBus data:

UI error:

### DIFF
--- a/app/services/chat_publisher.rb
+++ b/app/services/chat_publisher.rb
@@ -108,7 +108,7 @@ module ChatPublisher
         chat_channel,
         scope: Guardian.new(user), # We need a guardian here for direct messages
         root: :chat_channel
-      ).to_json
+      ).as_json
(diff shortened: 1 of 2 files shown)


nattsw (approved):

nattsw:
Thanks Martin. Your descriptions are the best.

martin-brennan:
@nattsw thank you! :bowing_man:

PR #999, opened 2022-07-07
FIX: Create channel modal error with type param required

martin-brennan:
Follow up to ca528966dfc4be9a920a8477813892e19a38d0a2, channel
creation was prevented because we still required the `type`
parameter in the controller but it was removed from create-channel.js . That
param is redundant anyway since the only type of channel we can
make now is a Category channel.

This commit fixes the issue and removes the type requirement.

### DIFF
--- a/app/controllers/chat_channels_controller.rb
+++ b/app/controllers/chat_channels_controller.rb
@@ -62,15 +62,12 @@ class DiscourseChat::ChatChannelsController < DiscourseChat::ChatBaseController
   end
 
   def create
-    params.require([:type, :id, :name])
+    params.require([:id, :name])
     guardian.ensure_can_create_chat_channel!
-    raise Discourse::InvalidParameters unless params[:type].downcase == 'category'
     raise Discourse::InvalidParameters.new(:name) if params[:name].length > SiteSetting.max_topic_title_length
 
-    chatable_type = 'Category'
-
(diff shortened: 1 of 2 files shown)


martin-brennan:
JS tests are not related even though they are for the create channel modal, there is a different issue with that that I will report internally

tgxworld (approved):

app/controllers/chat_channels_controller.rb
```diff
@@ -62,23 +62,20 @@ def unfollow
...
     guardian.ensure_can_create_chat_channel!
-    raise Discourse::InvalidParameters unless params[:type].downcase == 'category'
     raise Discourse::InvalidParameters.new(:name) if params[:name].length > SiteSetting.max_topic_title_length
 
-    chatable_type = 'Category'
-
     exists = ChatChannel.exists?(
-      chatable_type: chatable_type,
+      chatable_type: 'Category',
       chatable_id: params[:id],
```
tgxworld, line 71: Not related to the change but I find this param name confusing.. I would think `param[:id]` is the id for the chat channel and not the foreign key of `chatable`.

martin-brennan: Yeah it is a bit odd perhaps we could change that in another PR

PR #1000, opened 2022-07-07
FIX: Do not fetch messages for draft channel with no ID

martin-brennan:
Follow up to 9c800e0146182ce6a4a3353138d0ba7c8ccbe2a3

When starting a new DM channel with one or more users where
no channel exists we return a 404 from the server. However
we were still trying to fetch messages for that channel in
the chat-live-pane, leading to a GET like this:

http://localhost:4200/chat/undefined/messages.json?page_size=50

This commit fixes the issue by not doing any additional fetching
in chat-live-pane if the chat channel ID is not present.

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -137,7 +137,8 @@ export default Component.extend({
 
     if (
       this.chatChannel &&
-      this.registeredChatChannelId !== this.chatChannel?.id
+      this.chatChannel.id &&
+      this.registeredChatChannelId !== this.chatChannel.id
     ) {
       this._cleanRegisteredChatChannelId();
       this.messageLookup = {};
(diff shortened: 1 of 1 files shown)


assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -137,7 +137,8 @@ export default Component.extend({
 
     if (
       this.chatChannel &&
-      this.registeredChatChannelId !== this.chatChannel?.id
```
martin-brennan, line 140: There was no need for safe operation since nothing else would be evaluated if `this.chatChannel` was null or undefined

martin-brennan:
Note there are no acceptance tests for creating a direct message channel that I could find, but I tested this manually locally.

tgxworld:
> Note there are no acceptance tests for creating a direct message channel that I could find, but I tested this manually locally.

:grin: Is there one we can add?

tgxworld (approved):

tgxworld:
> grin Is there one we can add?

Confirmed internally that chat is in flux at the moment and this area is likely to change so the addition of tests are skipped here.

PR #1008, opened 2022-07-07
UX: displays delete btn using red icon

jjaffeux:


jjaffeux:
Screenshot 2022-07-07 at 16 03 58

PR #1011, opened 2022-07-08
FIX: makes drawer not expanded on full page

jjaffeux:
Before this fix, the drawer would still be expanded when switching to full page, albeit hidden, making the live pane inside the drawer still existing and new messages would be appended to this instead of the live pane of the full page.

### DIFF
--- a/assets/javascripts/discourse/components/topic-chat-float.js
+++ b/assets/javascripts/discourse/components/topic-chat-float.js
@@ -259,8 +259,6 @@ export default Component.extend({
       return false;
     }
 
-    // Set activeChannel to null to avoid a moment where the chat composer is rendered twice.
-    // Since the mobile-file-upload button has an ID, a JS error will break things otherwise.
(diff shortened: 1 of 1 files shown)


assets/javascripts/discourse/components/topic-chat-float.js
```diff
@@ -256,10 +261,6 @@ export default Component.extend({
 
     // Set activeChannel to null to avoid a moment where the chat composer is rendered twice.
     // Since the mobile-file-upload button has an ID, a JS error will break things otherwise.
```
gschlager, line 263: That comment seems to be out of place after these changes. Should it be moved or updated?

jjaffeux: I will just remove it I think, it doesn’t make much sense anymore, thanks.

gschlager (approved):

PR #1019, opened 2022-07-13
UX: Fix width of full page chat when paired with experimental sidebar

tgxworld:
### Before
Screenshot from 2022-07-13 14-09-58

### After
Screenshot from 2022-07-13 14-09-10


hnb-ku (approved):

tgxworld:
Thank you @hnb-ku :+1:

PR #1020, opened 2022-07-13
UI: is focused background was not visible in dm creator

jjaffeux:


jjaffeux:
Before fix:
Screenshot 2022-07-13 at 09 00 12

After fix:
Screenshot 2022-07-13 at 08 59 41

hnb-ku (approved):

PR #1023, opened 2022-07-14
FEATURE: Respect ignoring / muting / blocking DM user preferences

martin-brennan:
This commit covers several areas of functionality. In core,
a user can do the following:

* Ignore users - This makes it so the user cannot be cannot
PM'd, notified via mention, or invited into a topic by the target
user and also hides the target user's posts from topics
* Mute users - This makes it so the user cannot be cannot
PM'd, notified via mention, or invited into a topic by the target
user
* Disable PMs - This makes it so the user cannot be PM'd by
anyone
* Enable PMs from allowed users - This makes it so the user cannot
be PM'd by anyone except those the user allows

This commit applies all these rules to chat:

* When a user is ignoring or muting a target user, the target user
cannot create a new direct message channel with them, and the
user cannot create direct message channels with people they are
ignoring or muting either
* When a user is ignoring or muting a target user, the target user's
mentions of the user will be suppressed, including when the target
user mentions a group the user is part of, or `@all`, or `@here`
* When a user is preventing PMs from target user, the target user
cannot create a new direct message channel with them
* When a user is ignoring a target user, the target user's messages
in any channel will be hidden and collapsed, with an option to un-collapse
the message, similar to core's "view hidden posts" functionality
*  When a user is ignoring a target user, the target user's new sent messages
will not cause the unread dot to appear in channel names, and will always be
hidden
* When a user is ignoring a target user and unfollows an existing direct
message channel with them, they are not automatically followed again
when a message is sent to that channel by the ignored user


app/services/chat_publisher.rb
```diff
@@ -7,7 +7,11 @@ def self.publish_new!(chat_channel, chat_message, staged_id)
     content[:stagedId] = staged_id
     permissions = permissions(chat_channel)
     MessageBus.publish("/chat/#{chat_channel.id}", content.as_json, permissions)
-    MessageBus.publish("/chat/#{chat_channel.id}/new-messages", { message_id: chat_message.id, user_id: chat_message.user_id }, permissions)
+    MessageBus.publish(
+      "/chat/#{chat_channel.id}/new-messages",
+      { message_id: chat_message.id, user_id: chat_message.user_id, username: chat_message.username },
```
martin-brennan, line 12: Had to add the username since the `CurrentUserSerializer` only includes the `ignored_user` usernames and I need to filter this on the client-side

spec/requests/direct_messages_controller_spec.rb
```diff
@@ -116,5 +116,52 @@ def create_dm_channel(user_ids)
...
+            post "/chat/direct_messages/create.json", params: { usernames: [usernames] }
+          }.not_to change { DirectMessageChannel.count }
+          expect(response.status).to eq(422)
+          expect(response.parsed_body["errors"]).to eq([I18n.t("chat.errors.not_accepting_dms", username: user1.username)])
+        end
+      end
+
+      describe "user ignoring the actor" do
+        before do
+          Fabricate(:ignored_user, user: user1, ignored_user: user, expiring_at: 1.day.from_now)
```
jjaffeux, line 136: does it need to be in a before block? (same question for others after

jjaffeux:
What happens concerning replying indicator? Will I see it in a channel with someone I muted? What happens in a channel with 3 people where one is muted by only another one?

spec/requests/direct_messages_controller_spec.rb
martin-brennan: Yes since they are all testing different ways the "creating dms with communication error" shared example could fail

martin-brennan:
> What happens concerning replying indicator? Will I see it in a channel with someone I muted? 

@jjaffeux replying indicator still appears, not sure of the behaviour in topics if you have ignored someone there, will check.

> What happens in a channel with 3 people where one is muted by only another one?

If I am Alice and I have muted Bob and we are in a 3 person DM with Joe, then Joe will see Alice and Bob's replies and get green indicators for them, but Alice will only see Joe's replies and get a green indicator for them

app/jobs/regular/chat_notify_mentioned.rb
```diff
@@ -2,14 +2,14 @@
 
 module Jobs
   class ChatNotifyMentioned < ::Jobs::Base
+    delegate :direct_message_channel?, to: :@chat_channel
```
jjaffeux, line 5: Im not a super fan of this :p Is there a reason you did it explicitly for this one?

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -458,16 +458,30 @@ export default Component.extend({
...
     messageData.messageLookupId = this._generateMessageLookupId(messageData);
-    if (this.targetMessageId && this.targetMessageId === messageData.id) {
-      messageData.expanded = true;
-    }
     const prepared = ChatMessage.create(messageData);
     this.messageLookup[messageData.messageLookupId] = prepared;
     return prepared;
   },
 
+  _handleMessageHidingAndExpansion(messageData) {
```
jjaffeux, line 468: Part of me thinks we shouldn't even send a message to someone if we don't want to show it. I understand it might be more complex and cause perf issues, but thought I would say it :p

It would also avoid having to deal with this client side.

app/jobs/regular/chat_notify_mentioned.rb
martin-brennan: I thought it was weird to store `@is_direct_message_channel` in a variable, don't feel super strong tho I can just revert

app/serializers/chat_message_serializer.rb
```diff
@@ -103,4 +104,13 @@ def user_flag_status
...
     user_flag_status.present?
   end
+
+  def hidden
+    return false if scope.anonymous?
+    ignored_user_ids.include?(object.user_id)
+  end
+
+  def ignored_user_ids
+    @ignored_user_ids ||= scope.user.ignored_user_records.pluck(:ignored_user_id)
```
jjaffeux, line 114: is this called for each rendered message?

assets/javascripts/discourse/components/chat-live-pane.js
martin-brennan: Well it's like the `deleted_at` stuff, we want to show the placeholder that can be expanded/collapsed. I was also concerned that if we didn't send the messages through via new-messages in MessageBus that it would cause tracking state sync issues.

app/jobs/regular/chat_notify_mentioned.rb
jjaffeux: It's not possible to just do `@chat_channel.direct_message_channel?` ?

lib/chat_notifier.rb
```diff
@@ -105,8 +130,7 @@ def chat_users
   end
 
   def rest_of_the_channel
-    chat_users
```
jjaffeux, line 108: this is changing nothing but formatting right?

app/serializers/chat_message_serializer.rb
martin-brennan: Yes, unless the user is anonymous, however it's memoized so shouldn't cause an N1 (and the user may already have some of this preloaded in core, will have to double check)

app/jobs/regular/chat_notify_mentioned.rb
martin-brennan: Yah, will change it to that.

app/serializers/chat_message_serializer.rb
jjaffeux: Yes please let's double check this, thanks.

lib/chat_notifier.rb
martin-brennan: Yeah, hard to see in the github diff

app/serializers/chat_message_serializer.rb
martin-brennan: We actually don't preload anything for the current user in the default provider:

https://github.com/discourse/discourse/blob/a24bcceb19cb7272f1725d3fe5e2dd6be2bc0128/lib/auth/default_current_user_provider.rb#L156-L156

The current user serializer also loads this data:

https://github.com/discourse/discourse/blob/a24bcceb19cb7272f1725d3fe5e2dd6be2bc0128/app/serializers/current_user_serializer.rb#L237-L239

So maybe it does make sense to only filter on client side? Otherwise we end up with both this and current user serializer querying this data :thinking:

jjaffeux: or maybe core should be changed to preload this ?

assets/javascripts/discourse/components/chat-live-pane.js
martin-brennan: Just remember this ignored functionality copies what we already have in topics -- if you have ignored a user inside the topic all their posts are collapsed behind a "VIEW HIDDEN" button. I do see what you mean though, I did toss up whether I should somehow blackhole all those new messages in the backend and not send them to the client via MessageBus if you have ignored someone. Maybe we should discuss it a bit more internally?

martin-brennan: I suppose you could argue since Mute and Ignore are supposed to apply to creating DMs, we should not be showing any of their messages at all inside the DMs, but inside public channels Ignore works the same as topics with VIEW HIDDEN?

app/serializers/chat_message_serializer.rb
martin-brennan: Yeah not sure of the implications of preloading more things on the User in core via default user provider, but maybe it makes sense to load these and maybe user options? Just hope that it doesn't make a performance hit, will try it out

lib/chat_notifier.rb
jjaffeux: We should have ruby formatting by the end of the day on chat 🎉

martin-brennan: Yay!!!

app/serializers/chat_message_serializer.rb
martin-brennan: Actually I am not sure it will make a difference to preload in core seeing as how this plucks IDs and core plucks usernames. I think I will just remove this and rely on the frontend filtering in `_prepareSingleMessage`

jjaffeux:
I found one bug:

- User A ignores User B
- User A IS NOT TRACKING a conversation with User B (ie: User B is not appearing in the sidebar)
- User B sends me a message, the User B appears in the sidebar with a green dot

Screenshot 2022-07-21 at 09 44 44

1) We shouldn't show the greendot for sure
2) Might be debatable, but I would say it shouldn't appear in the sidebar

mcwumbly:
Based on the above, I think there are three gaps we should close.

1. If the message channel already exists, and they have untracked it, then I agree, it should not become tracked.
2. If it is still tracked, then green dots should not appear.
3. Furthermore I think we should prevent User A from adding User B to a message if they already have them muted (we already do the reverse, but preventing this makes sense to me too).

app/controllers/chat_controller.rb
```diff
@@ -117,8 +117,23 @@ def create_message
...
+
+      # If any of the channel users is ignoring, muting, or preventing DMs from
+      # the current user then we shold not auto-follow the channel once again or
+      # publish the new channel.
+      user_ids_allowing_communication = UserCommScreener.new(
+        acting_user: current_user,
+        target_user_ids: @chat_channel.user_chat_channel_memberships.map(&:user_id)
+      ).allowing_actor_communication
+
+      if user_ids_allowing_communication.any?
```
martin-brennan, line 129: Not sure if there are other places in the codebase that are mucking around with membership, will have to check, and we should probably move those into a central place if there are.

app/jobs/regular/chat_notify_watching.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 module Jobs
+  # TODO (martin) A spec for this whole class.
```
martin-brennan, line 4: Note it is tested somewhat through `ChatMessageCreator`. It does appear that previously testing via integration tests was preferred, but I think any class like this should have tests of its own.

lib/direct_message_channel_creator.rb
```diff
@@ -40,4 +45,64 @@ def self.update_memberships(unique_target_users, chat_channel_id)
...
+      .each do |user_id|
+        raise NotAllowed.new(
+                I18n.t(
+                  "chat.errors.not_accepting_dms",
+                  username: target_users.find { |user| user.id == user_id }.username,
+                ),
+              )
+      end
+
+    # TODO (martin) Add methods from here and below to UserCommScreener in core to
```
martin-brennan, line 65: This can be done in a separate polish PR, don't want to hold this up because core needs another API on the class I added.

app/jobs/regular/chat_notify_watching.rb
jjaffeux: Yes I fully agree, there's been multiple debates on this recently, but my position is:
- fully unit tested class
- acceptance test for important categories of events, eg in a controller: we test what happens when an error is raised in the underlying lib, not each kind of error message

Well this is my position at least :p

spec/lib/chat_notifier_spec.rb
```diff
@@ -178,6 +218,34 @@ def build_cooked_msg(message_body, user, chat_channel: channel)
         expect(to_notify[:global_mentions]).to be_empty
         expect(to_notify[:direct_mentions]).to contain_exactly(user_2.id)
       end
+
+      describe "users ignoring or muting the user creating the message" do
+        it "does not include these ignoring and muting users with the publish_new_mention publish" do
```
jjaffeux, line 223: `it "does not publish new mention to these users" do` ?

jjaffeux (approved):
Tested it a lot tonight, it seems quite good to me :+1: I kinda expect we will have multiple rounds of refinements as this is quite complex.

Thanks for the awesome work @martin-brennan :heart:

app/controllers/chat_controller.rb
```diff
@@ -117,8 +117,23 @@ def create_message
...
     if @chat_channel.direct_message_channel?
-      @chat_channel.user_chat_channel_memberships.update_all(following: true)
-      ChatPublisher.publish_new_channel(@chat_channel, @chat_channel.chatable.users)
+
+      # If any of the channel users is ignoring, muting, or preventing DMs from
+      # the current user then we shold not auto-follow the channel once again or
+      # publish the new channel.
+      user_ids_allowing_communication = UserCommScreener.new(
+        acting_user: current_user,
+        target_user_ids: @chat_channel.user_chat_channel_memberships.map(&:user_id)
```
jjaffeux, line 126: Is there any reason to prefer map over pluck here?

test/javascripts/components/chat-message-test.js
```diff
@@ -1,16 +1,64 @@
...
 import ChatChannel from "discourse/plugins/discourse-chat/discourse/models/chat-channel";
-import EmberObject from "@ember/object";
 import I18n from "I18n";
-import { module } from "qunit";
+import { module, test } from "qunit";
 
 module("Discourse Chat | Component | chat-message", function (hooks) {
   setupRenderingTest(hooks);
 
+  function generateMessageProps(messageData = {}) {
```
jjaffeux, line 14: I swear this is living its last weeks

martin-brennan: Thank God! To quote Samuel L Jackson -- "Yes it deserves to die, and I hope it burns in hell!"

app/jobs/regular/chat_notify_watching.rb
martin-brennan: Same :heart:

app/controllers/chat_controller.rb
martin-brennan: Oh yeah good point will fix

martin-brennan:
Thanks @jjaffeux ! Definitely more refinements to come on this :+1:

PR #1027, opened 2022-07-15
FIX: Don't corrupt users' last read IDs.

romanrizzi:
When we set up a chat channel, we subscribe to a message bus channel that notifies us about updates, one of them being new messages from other users. To show them in real-time, we add them immediately, but since we don't check if
all the latest ones are already loaded, we may create a gap where some messages will be missing until you refresh the tab.

We are changing how we handle new message updates and checking if it's possible to scroll into the future, and if it is, it means we haven't loaded all the latest messages yet, so we'll display a "jump to latest" button.

https://user-images.githubusercontent.com/5025816/179126445-ed56b275-0260-42db-9b8a-458a1d1ed07f.mov

Thanks to this bug, we also noticed that our last read update logic is problematic because it aggressively updates to the latest unread and **loaded** message every two seconds without checking if the message is visible. Because of
this, the new message update that introduces the gap also updates your last read message to the latest channel's message.

Another scenario where this is problematic is when you visit a channel for one second, then switch to a different one. Here, the runner could mark up to 50 messages (max items loaded per page) as read. It also means the next time
you visit, we will start fetching from the unread message number 51, so you'll have to scroll up to find where you really left.

This PR implements a binary search to find an unread message available in the viewport and schedules a last read message update only when scrolling into the future, reducing the number of attempts to update. The downside is that we
end up with a few more requests per user, but the read progression feels more natural.

- [X] chat-scoped -- core unaffected

### View mode

- [X] docked (windowed/drawer)

### Browsers

- [X] safari
- [X] chrome
- [X] firefox

### Device

- [X] desktop – wide screen
- [X] mobile – `?mobile_view=1`

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -174,10 +174,6 @@ export default Component.extend({
           if (!this.chatChannel.isDraft) {
             this.set("previewing", !Boolean(trackedChannel));
             this.loadDraftForChannel(this.chatChannel.id);
-
-            if (!isTesting()) {
-              this._updateLastReadMessage(0);
-            }
           }
         });
     }
@@ -400,6 +396,9 @@ export default Component.extend({
       }
 
       this._observeMessages(this.messages);
+      if (!isTesting()) {
+        this._updateLastReadMessage();
+      }
 
       if (this.targetMessageId) {
         this.scrollToMessage(this.targetMessageId, {
@@ -778,12 +777,16 @@ export default Component.extend({
         return;
       }
     }
-    this.messages.pushObject(
-      this._prepareSingleMessage(
-        data.chat_message,
-        this.messages[this.messages.length - 1]
-      )
+
+    const preparedMessage = this._prepareSingleMessage(
+      data.chat_message,
+      this.messages[this.messages.length - 1]
     );
+
+    this.messages.pushObject(preparedMessage);
+    schedule("afterRender", () => {
+      this._observeMessages([preparedMessage]);
+    });
 
     if (this.messages.length >= MAX_RECENT_MSGS) {
       this.removeMessage(this.messages.shiftObject());
@@ -946,12 +949,9 @@ export default Component.extend({
         }
 
         let latestUnreadMsgId = this.lastSendReadMessageId;
-        if (
-          this.messages?.length &&
-          this.messages[this.messages.length - 1].id > latestUnreadMsgId
-        ) {
+        if (this.messages[this.messages.length - 1]?.id > latestUnreadMsgId) {
           const visibleMessages = document.querySelectorAll(
-            ".chat-message-container.visible"
+            ".chat-message-container[data-visible=true]"
           );
           if (visibleMessages?.length > 0) {
             latestUnreadMsgId = parseInt(
@@ -1463,11 +1463,7 @@ export default Component.extend({
 
     const callback = (entries) => {
       entries.forEach((entry) => {
-        if (entry.isIntersecting) {
-          entry.target.classList.add("visible");
-        } else {
-          entry.target.classList.remove("visible");
-        }
+        entry.target.dataset.visible = entry.isIntersecting;
       });
     };
 
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -63,6 +63,7 @@ export default Component.extend({
   _nextStagedMessageId: 0, // Iterate on every new message
   _lastSelectedMessage: null,
   targetMessageId: null,
+  hasNewMessages: null,
 
   chat: service(),
   router: service(),
@@ -123,7 +124,7 @@ export default Component.extend({
       this,
       "highlightOrFetchMessage"
     );
-    this._stopLastReadRunner();
+    this._cancelPendingReadUpdate();
 
     // don't need to removeEventListener from scroller as the DOM element goes away
     cancel(this.stickyScrollTimer);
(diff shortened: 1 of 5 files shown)


assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -905,34 +925,69 @@ export default Component.extend({
...
+      return this._findNextUnreadMessage(
+        candidateIdx + 1,
+        end,
+        lastReadMessageId
+      );
     }
   },
 
-  _stopLastReadRunner() {
+  _messageVisible(message) {
```
jjaffeux, line 977: How about we try something using IntersectionObserver?

romanrizzi: What do you have in mind? I don't immediately see how the observer would fit here. We don't know which of the unread messages is visible.

romanrizzi: I implemented the observer, but added a class instead of a data attribute. I forgot about that part of your suggestion. Is there any difference between class vs data attr?

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1400,4 +1426,34 @@ export default Component.extend({
...
+    const options = {
+      root: document.querySelector(".chat-messages-container"),
+      rootMargin: "0px",
+      threshold: 1.0,
+    };
+
+    const callback = (entries) => {
+      entries.forEach((entry) => {
+        if (entry.isIntersecting) {
+          entry.target.classList.add("visible");
```
pmusaraj, line 1440: It's better to use a data attribute here instead of a class because classes a) are more commonly used for styling (and in this case we don't want to add styling) and b) avoids name clashes in our stylesheets. The "visible" class is indeed used in a bunch of places in core and plugins, so avoiding it here would be best.

pmusaraj (commented):
I pulled the branch locally and noticed a small issue: when adding/receiving new messages, the new message containers aren't observed. Here's an illustration: 

image

I added a pink background to items with the "visible" class, and then sent messages to the channel. The last two are such sent messages, and the observer is not aware of them. Tried scrolling up/down and that still didn't fix it. (A refresh did, but that means that the last read ID won't be updated until a refresh.) 

Note that I saw the same issue on the message sender's screen.

romanrizzi:
Thanks for the feedback, @pmusaraj! I addressed both comments.

assets/stylesheets/common/common.scss
```diff
@@ -371,20 +371,35 @@ $float-height: 530px;
     bottom: 1em;
     border-radius: 100%;
     left: 50%;
-    opacity: 20%;
+    opacity: 75%;
```
pmusaraj, line 374: This feels a little strong. Maybe split the difference and go for 50% opacity?

pmusaraj (approved):
Looking good to me. Let's give it a try!

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1427,4 +1456,34 @@ export default Component.extend({
...
+      threshold: 1.0,
+    };
+
+    const callback = (entries) => {
+      entries.forEach((entry) => {
+        if (entry.isIntersecting) {
+          entry.target.dataset.visible = true;
+        } else {
+          entry.target.dataset.visible = false;
+        }
```
jjaffeux, line 1473:
```suggestion
entry.target.dataset.visible = entry.isIntersecting;
```

jjaffeux:
Ok awesome thanks (and Penar too :heart:), sorry didn't get much time for it this week. LGTM, lets try this. I didnt want to annoy you with this in this PR but I would want us to use a modifier or even a tag less component instead of having this directly in live pane code. Let's do this in a future PR.

assets/javascripts/discourse/templates/components/chat-live-pane.hbs
```diff
@@ -76,9 +76,12 @@
     <a
       href
       title={{i18n "chat.scroll_to_bottom"}}
-      class="btn btn-flat chat-scroll-to-bottom"
+      class="btn btn-flat chat-scroll-to-bottom {{if hasNewMessages "unread-messages"}}"
```
jjaffeux, line 79:
```suggestion
      class={{concat-class "btn" "btn-flat" "chat-scroll-to-bottom" (if hasNewMessages "unread-messages")}}
```

jjaffeux: Try it locally, I might have done it wrong here :p

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -923,15 +948,27 @@ export default Component.extend({
...
         }
 
-        let messageId;
-        if (this.messages?.length) {
-          messageId = this.messages[this.messages.length - 1]?.id;
+        let latestUnreadMsgId = this.lastSendReadMessageId;
+        if (
+          this.messages?.length &&
+          this.messages[this.messages.length - 1].id > latestUnreadMsgId
+        ) {
```
jjaffeux, line 955: Nitpicking here, but I wonder if we can't just go with:

```
this.messages[this.messages.length - 1]?.id > latestUnreadMsgId
```

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -387,6 +395,11 @@ export default Component.extend({
         return;
       }
 
+      if (!isTesting()) {
+        this._observeMessages(this.messages);
```
jjaffeux, line 399: What is the reason we take it out of testing?

jjaffeux (approved):

jjaffeux:
BTW I fixed the test failures and deprecations in another PR. You have linting to fix though.

assets/javascripts/discourse/components/chat-live-pane.js
romanrizzi: My bad. Only the last read update should be disabled.

PR #1031, opened 2022-07-15
UX: fully reimplements browse page

jjaffeux:
- new UI with better mobile support
- implements infinite scrolling
- allows to display closed and archived channels
- allows filtering of channels
- displays muted state from the list


jjaffeux:
Desktop light mode:

Screenshot 2022-07-15 at 20 26 16

Mobile dark mode:
Screenshot 2022-07-15 at 20 26 55

No results view:
Screenshot 2022-07-15 at 20 48 39

Archived channel:
Screenshot 2022-07-15 at 20 48 53

jjaffeux:
@chapoi If you want to do few tweaks on it early next week be my guest, otherwise that will be after the merge. One early feedback from @mcwumbly is that we might possibly want two lines of description instead of one. I also feel like it would be interesting it you think that makes sense.

ZogStriP (approved):

jjaffeux:
Oh yes I saw it the other day and forgot, will fix

PR #1037, opened 2022-07-18
FEATURE: WIP chat-toggle-membership-btn

martin-brennan:


assets/javascripts/discourse/components/chat-toggle-membership-btn.js
```diff
@@ -0,0 +1,91 @@
...
+import { action, computed } from "@ember/object";
+
+export default class ChatToggleMembershipBtn extends Component {
+  @service chat;
+  tagName = "";
+
+  channel = null;
+  afterToggle = null;
+  isLoading = false;
+  labelType = null;
```
martin-brennan, line 15: I am not sure if it is good to have so many options, but we have a lot of different join/leave channel button styles.

Chat browse page:

image
image

Leave channel button for DMs

image

Preview join channel button:

image

About channel:

image

assets/javascripts/discourse/templates/components/chat-channel-row.hbs
```diff
@@ -11,9 +11,14 @@
...
-    {{chat-channel-leave-btn
-      channel=channel
-      onLeaveChannel=(action "onLeaveChannel")
-    }}
+    {{#unless site.mobileView}}
+      <ChatToggleMembershipBtn
+        @channel={{channel}}
+        @afterToggle={{action "afterMembershipToggle"}}
+        @leaveIcon="times"
+        @leaveClasses="btn-flat chat-channel-leave-btn"
```
martin-brennan, line 19: `chat-channel-leave-btn` is very loaded and styled in a lot of CSS files which is why I left it in, not sure if it would be better to move all those styles to use `chat-toggle-membership-btn__leave-channel` ?

assets/stylesheets/common/chat-toggle-membership-btn.scss
```diff
@@ -0,0 +1,9 @@
+.chat-toggle-membership-btn {
```
martin-brennan, line 1: I am not sure if this is even needed because of `joinClasses` and `leaveClasses`

config/locales/client.en.yml
```diff
@@ -40,6 +40,8 @@ en:
         delete_channel: "Delete Channel"
         join_channel: "Join Channel"
         leave_channel: "Leave Channel"
+        join: "Join"
```
martin-brennan, line 43: We also have these translations many times in this same file

martin-brennan:
The `chat-channel-leave-btn` component would also be completely deleted by this PR

assets/javascripts/discourse/components/chat-toggle-membership-btn.js
```diff
@@ -0,0 +1,91 @@
+import Component from "@ember/component";
+import I18n from "I18n";
+import { inject as service } from "@ember/service";
+import { popupAjaxError } from "discourse/lib/ajax-error";
+import ChatApi from "discourse/plugins/discourse-chat/discourse/lib/chat-api";
+import { action, computed } from "@ember/object";
+
+export default class ChatToggleMembershipBtn extends Component {
```
jjaffeux, line 8: we should try glimmer component if possible here

assets/javascripts/discourse/components/chat-toggle-membership-btn.js
```diff
@@ -0,0 +1,91 @@
...
+  @service chat;
+  tagName = "";
+
+  channel = null;
+  afterToggle = null;
```
jjaffeux, line 13: I would go with onToggle I think

assets/javascripts/discourse/components/chat-toggle-membership-btn.js
```diff
@@ -0,0 +1,91 @@
+import Component from "@ember/component";
+import I18n from "I18n";
+import { inject as service } from "@ember/service";
+import { popupAjaxError } from "discourse/lib/ajax-error";
+import ChatApi from "discourse/plugins/discourse-chat/discourse/lib/chat-api";
+import { action, computed } from "@ember/object";
+
+export default class ChatToggleMembershipBtn extends Component {
+  @service chat;
+  tagName = "";
```
jjaffeux, line 10: won't be needed anymore with glimmer

assets/javascripts/discourse/components/chat-toggle-membership-btn.js
```diff
@@ -0,0 +1,91 @@
...
+  isLoading = false;
+  labelType = null;
+  joinTitle = null;
+  joinIcon = null;
+  joinClasses = "";
+  leaveTitle = null;
+  leaveIcon = null;
+  leaveClasses = "";
+
+  @computed("channel.following")
```
jjaffeux, line 23: I wonder if we don't go a different way:

- have reasonable defaults for labels/icons
- allow you to override them when calling the component

```
<ToggleMembershipButton options=(hash leaveIcon="bell" joinLabel="foo") />
```

assets/javascripts/discourse/components/chat-toggle-membership-btn.js
```diff
@@ -0,0 +1,91 @@
...
+    } else {
+      return I18n.t("chat.channel_settings.join_channel");
+    }
+  }
+
+  @action
+  onJoinChannel() {
+    this.set("isLoading", true);
+
+    return ChatApi.followChatChannel(this.channel.id)
```
jjaffeux, line 48: Maybe we implement some `channel.updateMembership(membership)` function in channel model, and call it in chat-api client that way this update membership part is seamless.

assets/javascripts/discourse/components/chat-toggle-membership-btn.js
```diff
@@ -0,0 +1,91 @@
...
+
+        this.channel.setProperties({
+          expanded: false,
+          following: false,
+          memberships_count: membership.user_count,
+        });
+
+        this.chat.stopTrackingChannel(this.channel);
+
+        if (this.afterToggle) {
```
jjaffeux, line 85: nitpick, but generally I write this like that now: `this.afterToggle?.()`

assets/javascripts/discourse/templates/components/chat-channel-about-view.hbs
```diff
@@ -71,21 +71,12 @@
...
-  {{else}}
-    {{d-button
-      action=(action "onJoinChannel")
-      label="chat.channel_settings.join_channel"
-      class="chat-form__btn join-channel-btn btn-flat"
-      icon="sign-in-alt"
-      disabled=isLoading
-    }}
-  {{/if}}
+  <ChatToggleMembershipBtn
```
jjaffeux, line 74: So for the public API I would try to follow this: mandatory -> top level, optional -> some options hashs

```
<ToggleChannelMembershipButton
  @channel={{...}}
  @onToggle={{...}}
  @options={{hash
    joinIcon="bell"
    leaveClass="foo"
  }}
/>
```

Note:
- I changed the name, in the future I will drop the chat from many names, and I think it makes more sense to use channel here anyways
- s/Btn/Button, for parity with d-button
- onToggle as mentioned sooner
- class and not classes, if you look at html we say class even if you can pass multiple values, this is the class value

assets/javascripts/discourse/templates/components/chat-channel-row.hbs
jjaffeux: I think we could do

```
toggle-channel-membership-btn -leave
toggle-channel-membership-btn -join
```

That would be right @chapoi no?

assets/javascripts/discourse/components/chat-toggle-membership-btn.js
martin-brennan: So something like this inside the `channel` model? Then called like `this.channel.updateMembership(false/true).then(() => this.onToggle.?())` ?

```javascript
updateMembership(follow) {
  if (follow) {
    return ChatApi.followChatChannel(this.channel.id)
      .then((membership) => {
        this.channel.setProperties({
          following: true,
          muted: membership.muted,
          desktop_notification_level: membership.desktop_notification_level,
          mobile_notification_level: membership.mobile_notification_level,
          memberships_count: membership.user_count,
        });

        this.chat.startTrackingChannel(this.channel);
      });
  } else {
    return ChatApi.unfollowChatChannel(this.channel.id)
      .then((membership) => {
        this.channel.setProperties({
          expanded: false,
          following: false,
          memberships_count: membership.user_count,
        });

        this.chat.stopTrackingChannel(this.channel);
      })
      .catch(popupAjaxError);
  }
}
```

assets/javascripts/discourse/components/chat-toggle-membership-btn.js
martin-brennan: Yeah still wasn't sure whether I should, I will try it out :+1:

assets/javascripts/discourse/components/chat-toggle-membership-btn.js
martin-brennan: Yeah that's what I was wondering whether an options hash would be better, will go with this!

assets/javascripts/discourse/templates/components/chat-channel-about-view.hbs
martin-brennan: All great points, thanks

assets/javascripts/discourse/templates/components/chat-channel-row.hbs
martin-brennan: I thought it would be `__leave` since it's an element under the block? Vs `--leave` if it was a modifier, like `--danger`?

martin-brennan: I mainly wasn't sure whether I should change the CSS files to no longer use `chat-channel-leave-btn` and to target this new class

jjaffeux: I know @chapoi has been using this in browse page for -closed and -archived so I would think it can be used here too?

assets/javascripts/discourse/components/chat-toggle-membership-btn.js
martin-brennan: Ohhh I see, yes sounds good :+1:

assets/javascripts/discourse/components/chat-toggle-membership-btn.js
martin-brennan: As discussed can't do this because then that would require the ChatChannel model to become a native class...and then lots of other things need to change

assets/javascripts/discourse/models/chat-channel.js
```diff
@@ -64,6 +64,17 @@ const ChatChannel = RestModel.extend({
     return !READONLY_STATUSES.includes(this.status);
   },
 
+  updateMembership(following, membership) {
+    this.setProperties({
+      following,
+      expanded: following,
```
jjaffeux, line 70: what is this property doing already?

martin-brennan: I have no idea it's kind of a weird one, I can't find it anywhere. Maybe we just trash it and see, the only thing I can see with expanded are chat-message and also the topic-chat-float which obviously does not rely on this

jjaffeux: Yes I feel this is some odd thing which shouldn't be here, I don't what an expanded membership/channel is, let's trash this.

assets/javascripts/discourse/templates/components/chat-channel-row.hbs
chapoi: Sorry, I don't get mails from these mentions, should change that
I use modifiers modified (get it) with just 1 dash. So indeed -leave

jjaffeux (approved):

PR #1044, opened 2022-07-18
UI: tweaks to chat browse stylings

jjaffeux:


CvX:
I swear that's not the same screenshot as before 😄 

image

The issue is the same though 😉 chat sidebar, Chromium

jjaffeux:
Yep I share the implementation now :)

PR #1045, opened 2022-07-18
FIX: Hook to event instead of model callback to make sure we auto-join users

romanrizzi:


jjaffeux:
Is it really more sure? Can't a user be made active by other means?

romanrizzi:
> Is it really more sure? Can't a user be made active by other means?

This is the only point where activation happens outside of creation AFAIK. All the other places assume email is valid when creating.

jjaffeux:
> > Is it really more sure? Can't a user be made active by other means?
> 
> This is the only point where activation happens outside of creation AFAIK. All the other places assume email is valid when creating.
> 
> The problem with the previous callback is that it wasn't running when using `save!` instead of `update!`

Ok, honestly I'm not fan of having to trigger an event ourselves, if implementation changes we might not cache it in the test. OK yo merge this, but if you can think of a better test that would be great 🙌

romanrizzi:
> Ok, honestly I'm not fan of having to trigger an event ourselves, if implementation changes we might not cache it in the test. OK to merge this, but if you can think of a better test that would be great 🙌

Yes, that's a good suggestion. Thanks! I added a test that calls the account activation controller 👌

PR #1047, opened 2022-07-18
FIX: ensures deleting a channel redirects to chat home

jjaffeux:
Note this commit also refactors fabricators to make them more powerful and extendable.


test/javascripts/helpers/fabricators.js
```diff
@@ -1,57 +1,56 @@
...
+  cooked: "This is a test message",
+});
 
-export default function fabricate(model, options = {}) {
-  let base;
+const directMessageChannelFabricator = Fabricator({
+  id: 1,
+  chatable_type: CHATABLE_TYPES.directMessageChannel,
+  status: "open",
+  chatable: directChannelChatableFabricator(),
```
martin-brennan, line 37: This should have `__model: "ChatChannel",` as well right?

test/javascripts/helpers/fabricators.js
```diff
@@ -1,57 +1,56 @@
...
+});
 
-  const final = Object.assign(base, options);
-  switch (model) {
-    case "chat-channel":
-      return ChatChannel.create(final);
-  }
-}
+export default {
+  chatChannel: (options) => ChatChannel.create(chatChannelFabricator(options)),
```
martin-brennan, line 51: You don't need `ChatChannel.create` here and the next line because of `__model` right? The Fabricator should return the correct instance

test/javascripts/acceptance/close-chat-channel-modal-test.js
```diff
@@ -0,0 +1,42 @@
+import { acceptance } from "discourse/tests/helpers/qunit-helpers";
```
martin-brennan, line 1: This should be `delete-chat-channel-modal-test` I think, also will we add the other tests for it back in at some point from that old delete test that was removed recently?

martin-brennan (approved):
A few minor things...really loving where the Fabricators are heading :+1:

test/javascripts/helpers/fabricators.js
jjaffeux: Yes I added the __model stuff after and forgot about this :p

PR #1049, opened 2022-07-19
FIX: For non-open channels do not show Join on preview card

martin-brennan:
When we show the chat channel preview card if you
are not a member of the channel, we should not be
showing the Join button if the channel is anything
but open (e.g. closed/archived/etc)

### DIFF
--- a/assets/javascripts/discourse/components/chat-channel-preview-card.js
+++ b/assets/javascripts/discourse/components/chat-channel-preview-card.js
@@ -3,6 +3,7 @@ import { popupAjaxError } from "discourse/lib/ajax-error";
 import { isEmpty } from "@ember/utils";
 import ChatApi from "discourse/plugins/discourse-chat/discourse/lib/chat-api";
 import { action, computed } from "@ember/object";
+import { alias } from "@ember/object/computed";
 import { inject as service } from "@ember/service";
 
 export default class ChatChannelPreviewCard extends Component {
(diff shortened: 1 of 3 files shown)


test/javascripts/components/chat-channel-preview-card-test.js
```diff
@@ -86,5 +86,35 @@ discourseModule(
         "it shows a link to browse all channels"
       );
     });
+
+    test("closed channel", async function (assert) {
```
tgxworld, line 90: I don't think it is the concern of this component test to know about the different state of the different channels. I think we only need a single test here where `channel.isOpen` is `false`.

martin-brennan: `isOpen` is derived from the channel status and cannot be set directly:

https://github.com/discourse/discourse-chat/blob/31789e24c1eeda0b992b5417f07802b52bf42957/assets/javascripts/discourse/models/chat-channel.js#L79-L82

But I can just remove everything except the `closed channel` test here

martin-brennan:
@tgxworld can you re-review please

tgxworld (approved):

assets/javascripts/discourse/components/chat-channel-preview-card.js
```diff
@@ -12,6 +13,8 @@ export default class ChatChannelPreviewCard extends Component {
 
   channel = null;
 
+  @alias("channel.isOpen") showJoinButton;
```
jjaffeux, line 16: no alias please, it's bi directionnal setter we dont want this

jjaffeux (changes requested):

jjaffeux (approved):

PR #1056, opened 2022-07-19
DEV: Fix some typos

CvX:
"participanting" is my fav 😃


gschlager (approved):
"notherwise" is a good one as well :grin:

PR #1058, opened 2022-07-20
FEATURE: Statistics for chat messages, users, and channels

martin-brennan:
This commit introduces statistics for chat that will show
on the /about endpoint and can also be used for other reporting.

### DIFF
--- a/lib/chat_statistics.rb
+++ b/lib/chat_statistics.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+class DiscourseChat::Statistics
+  def self.about_messages
+    {
+      last_day: ChatMessage.where('created_at > ?', 1.days.ago).count,
+      "7_days" => ChatMessage.where('created_at > ?', 7.days.ago).count,
+      "30_days" => ChatMessage.where('created_at > ?', 30.days.ago).count,
+      previous_30_days: ChatMessage.where('created_at BETWEEN ? AND ?', 60.days.ago, 30.days.ago).count,
+      count: ChatMessage.count
+    }
+  end
+
+  def self.about_channels
+    {
+      last_day: ChatChannel.where(status: :open).where('created_at > ?', 1.days.ago).count,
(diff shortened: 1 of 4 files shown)


spec/lib/chat_statistics_spec.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+describe DiscourseChat::Statistics do
+  fab!(:frozen_time) { DateTime.parse("2022-07-08 09:30:00") }
+
+  def minus_time(time)
```
martin-brennan, line 6: This weirdness has to be done since `freeze_time` does not work with `fab!` , the `before do` blocks run after the `fab!` calls

jjaffeux (approved):
LGTM

lib/chat_statistics.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+class DiscourseChat::Statistics
+  def self.about_messages
+    {
+      last_day: ChatMessage.where('created_at > ?', 1.days.ago).count,
+      "7_days" => ChatMessage.where('created_at > ?', 7.days.ago).count,
+      "30_days" => ChatMessage.where('created_at > ?', 30.days.ago).count,
+      previous_30_days: ChatMessage.where('created_at BETWEEN ? AND ?', 60.days.ago, 30.days.ago).count,
```
tgxworld, line 9: @martin-brennan I don't think we have a index that covers these queries here. The queries are likely going to be expensive because the `chat_messages` table is likely to have alot of rows.

martin-brennan: Oh yeah good idea, we have that on `Post` and `Topic` in core which are used for `About` there. We don't have it on `UserAction` (for likes) or `User` though, core stats here:

https://github.com/discourse/discourse/blob/98bacbd2c6b9fe57167cd32af5eb4839b4a5d1f6/app/models/about.rb#L82-L104

PR #1065, opened 2022-07-20
FIX: Add a scheduled job to auto-join users when their state changes.

romanrizzi:
Additionally, trigger an auto-join for a user the first time we set `last_seen_at` instead of on create to avoid adding users who never visited the site.

### DIFF
--- a/app/jobs/scheduled/auto_join_users.rb
+++ b/app/jobs/scheduled/auto_join_users.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module Jobs
+  class AutoJoinUsers < ::Jobs::Scheduled
+    every 1.hour
+
+    def execute(_args)
+      ChatChannel
+        .where(auto_join_users: true)
+        .each { |channel| UserChatChannelMembership.enforce_automatic_channel_memberships(channel) }
+    end
+  end
+end
--- a/spec/jobs/scheduled/auto_join_users_spec.rb
+++ b/spec/jobs/scheduled/auto_join_users_spec.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe Jobs::AutoJoinUsers do
+  it "works" do
+    Jobs.run_immediately!
+    channel = Fabricate(:chat_channel, auto_join_users: true)
+    user = Fabricate(:user, last_seen_at: 1.minute.ago, active: true)
+
+    membership = UserChatChannelMembership.find_by(user: user, chat_channel: channel)
+    expect(membership).to be_nil
+
+    subject.execute({})
+
+    membership = UserChatChannelMembership.find_by(user: user, chat_channel: channel)
+    expect(membership.following).to eq(true)
+  end
+end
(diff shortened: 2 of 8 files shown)


spec/jobs/scheduled/auto_join_users_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+describe Jobs::AutoJoinUsers do
+  it 'works' do
+    Jobs.run_immediately!
+    channel = Fabricate(:chat_channel, auto_join_users: true)
+    user = Fabricate(:user, last_seen_at: 1.minute.ago, active: true)
+
+    subject.execute({})
+
+    membership = UserChatChannelMembership.find_by(user: user, chat_channel: channel)
```
jjaffeux, line 13: would probably be interesting to track than the job is actually changing this value, ATM with this test following could be true before execute and nothing would fail

spec/plugin_spec.rb
```diff
@@ -254,18 +254,35 @@ def assert_user_following_state(user, channel, following:)
...
+      end
+
+      it 'queueus a job on save! too' do
+        user.last_seen_at = 1.minute.ago
         user.save!
 
         assert_user_following_state(user, channel, following: true)
       end
 
+      it "does nothing if it's not last_seen_at first update" do
```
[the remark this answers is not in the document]
romanrizzi: We only want to run the job the first time `last_seen_at` is set. We previously hooked to `user_created`, but that could leave us with auto-joined users who never visited the site.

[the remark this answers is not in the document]
jjaffeux: `it "doesn’t process user until user has been seen" do` ?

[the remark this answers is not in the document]
romanrizzi: It's actually the opposite. Will come up with a less confusing name, though.

jjaffeux:
You have few linting to fix due to syntax tree, but should be easy, let me know if you need help

app/jobs/scheduled/auto_join_users.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Jobs
+  class AutoJoinUsers < ::Jobs::Scheduled
+    every 6.hours
```
jjaffeux, line 5: I wonder if we don't want this more often? Like 1 hour?

romanrizzi: Sounds good to me!

romanrizzi:
@jjaffeux I pushed an additional change to change how we enforce the `max_chat_auto_joined_users` limit. This job will more than once now.

spec/requests/users_controller_spec.rb
```diff
@@ -2,18 +2,19 @@
...
       Jobs.run_immediately!
       UsersController.any_instance.stubs(:honeypot_or_challenge_fails?).returns(false)
       SiteSetting.send_welcome_message = false
-      SiteSetting.chat_enabled = true
     end
 
     it "triggers the auto-join process" do
+      user = Fabricate(:user, last_seen_at: 1.minute.ago, active: false)
+      email_token = Fabricate(:email_token, user: user)
+      SiteSetting.chat_enabled = true
```
jjaffeux, line 16: why is this moved here?

jjaffeux (approved):

spec/requests/users_controller_spec.rb
romanrizzi: Thought it was necessary, but it's not. Will revert.

PR #1071, opened 2022-07-20
FIX: ensures user can see channel even with membership

jjaffeux:
Only treats channels with category as chatable, further todo should work out direct message channels.

### DIFF
--- a/app/queries/chat_channel_memberships_query.rb
+++ b/app/queries/chat_channel_memberships_query.rb
@@ -4,9 +4,17 @@ class ChatChannelMembershipsQuery
   def self.call(channel, limit: 50, offset: 0, username: nil)
     query =
       UserChatChannelMembership
-        .includes(:user)
+        .joins(:user)
         .where(user: User.activated.not_suspended.not_staged)
         .where(chat_channel: channel, following: true)
+
+    if channel.category_channel? && channel.allowed_group_ids
+      query =
+        query.joins(user: :group_users).where(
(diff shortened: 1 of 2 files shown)


app/queries/chat_channel_memberships_query.rb
```diff
@@ -4,10 +4,18 @@ class ChatChannelMembershipsQuery
...
-        .includes(:user)
+        .joins(:user)
         .where(user: User.activated.not_suspended.not_staged)
         .where(chat_channel: channel, following: true)
 
+    if channel.category_channel? && channel.allowed_group_ids
+      query =
+        query.joins(user: :group_users).where(
+          "group_users.group_id IN (?)",
+          channel.allowed_group_ids,
```
martin-brennan, line 15: The only possible issue I can think with this is that, do we want admins to be able to show up here regardless of whether they are in the group for the category? If so we should change to `group_users.group_id IN (?) OR users.admin`. I am basing this by what I remember doing in https://github.com/discourse/discourse-chat/blob/0dec9030b8d738d720fbd585544fb59c1c663157/lib/chat_channel_fetcher.rb#L16-L16

martin-brennan (approved):
Only one question, looks great

app/queries/chat_channel_memberships_query.rb
jjaffeux: My first reaction wouod be no, but maybe Im wrong... I will start like this and add it later if need is raised, thanks

PR #1079, opened 2022-07-21
FEATURE: show status on the direct message users list

AndrewPrigorshnev:
Note that before merging this, https://github.com/discourse/discourse/pull/17497 and https://github.com/discourse/discourse/pull/17598 should be merged in Core.

This adds "live" status to the chat popup:
Screenshot 2022-07-21 at 23 46 50

We're going to have a better design for displaying status, but the component with that design isn't merged in Core yet.

**_Edit:_**
_Initially, this PR also added "live" user status to the sidebar:_
Screenshot 2022-07-21 at 23 46 35

_It stopped working after changes introduce by https://github.com/discourse/discourse-chat/pull/977. Status will be added to the sidebar in a separate PR._


assets/javascripts/discourse/models/chat-channel.js
```diff
@@ -115,6 +116,23 @@ const ChatChannel = RestModel.extend({
...
 });
 
+ChatChannel.reopenClass({
+  create(args) {
+    args = args || {};
+    this._intiUserModels(args);
+    return this._super(args);
+  },
+
+  _intiUserModels(args) {
```
jjaffeux, line 126: This would probably need a unit test

SamSaffron:
I am a bit confused, where is `user.user_status` being preloaded in the controller? We need to make sure we include the relation to avoid the N+1.

AndrewPrigorshnev:
> We need to make sure we include the relation to avoid the N+1.

I'm sorry, I forgot to handle it, I've pushed a fix. 

Also, for the future, I believe we need to further optimize this place. When messaging a new user, we add that user to the direct messages list on the sidebar, and it's pretty slow because we reload _all_ users at that moment instead of just adding the new user to the list. This will also make code simpler, but from what I see this improvement would take a serious refactoring.

assets/javascripts/discourse/models/chat-channel.js
AndrewPrigorshnev: Not sure about this, frankly. In general, we should be testing _behavior_, not _implementation_. This just wraps user-like data into user models, which is clearly an implementation detail.

jjaffeux:
> > I am a bit confused, where is `user.user_status` being preloaded in the controller? We need to make sure we include the relation to avoid the N+1.
> 
> Yep I had the same thought yesterday. I know Andrei created a specific Serializer to make it easier to optimize perf here, but then I would also have been expected to see some code about user_status where load direct messages. AFAIK we have nothing about this in Chat code ATM.

Im fine with either here, and honestly tired of arguing on these subjects.

assets/javascripts/discourse/models/chat-channel.js
jjaffeux: Im fine with either here, and honestly tired of arguing on these subjects :)

jjaffeux:
Ok I just tried it it seems to work fine (I have the listener error but this is due to me not having core commit). N+1 seems correctly handled:

Screenshot 2022-07-22 at 20 12 56

Feel free to merge when ready :) Thanks.

jjaffeux (approved):

jjaffeux:
I took the liberty to fix various linting issues, be sure to read our topic about ruby syntax tree

AndrewPrigorshnev:
> (I have the listener error but this is due to me not having core commit)

Yes, it awaits merging of core PRs, shouldn't work with current version of Core. And I've checked with mini profiler as well :).

AndrewPrigorshnev:
After recent changes introduce by https://github.com/discourse/discourse-chat/pull/977 status on the sidebar in this PR stopped working. But it still adds status here:

180305108-59b2b728-5b02-4fe5-9528-c18732185cb0

So I'm merging it, and I'll add status to the sidebar in a separate PR. I'm also correcting the description of the PR

PR #1080, opened 2022-07-22
FIX: Hook to user_seen event to auto-join users on their first visit.

romanrizzi:


spec/plugin_spec.rb
```diff
@@ -254,35 +254,25 @@ def assert_user_following_state(user, channel, following:)
 
     describe "when a user is created" do
       fab!(:category) { Fabricate(:category) }
-      fab!(:user) { Fabricate(:user) }
+      let(:user) { Fabricate(:user, last_seen_at: nil, first_seen_at: nil) }
```
jjaffeux, line 257: do we need let now?

romanrizzi: Yes. Re-using the user with `fab!` makes the first test flaky.

spec/plugin_spec.rb
```diff
@@ -254,35 +254,25 @@ def assert_user_following_state(user, channel, following:)
...
         assert_user_following_state(user, channel, following: false)
       end
 
       it "does nothing if auto-join is disabled" do
         channel.update!(auto_join_users: false)
 
-        user.update!(last_seen_at: 1.minute.ago)
+        user.update_last_seen!
 
         assert_user_following_state(user, channel, following: false)
```
jjaffeux, line 277: Maybe we need a test with `user.update!(first_seen_at: Time.now)`? Or do we have it somewhere already?

romanrizzi: I believe the spec in line 265 covers that scenario

jjaffeux (approved):

PR #1081, opened 2022-07-22
FIX: Unescape emoji in channel title when displaying in core sidebar

tgxworld:


jjaffeux (approved):

tgxworld:
Thank you for reviewing @jjaffeux

PR #1082, opened 2022-07-22
FEATURE: show status next to avatar on chat messages

AndrewPrigorshnev:
_Note that before merging this, https://github.com/discourse/discourse/pull/17497 and https://github.com/discourse/discourse/pull/17598 should be merged in Core._

This adds "live" status to the user avatar on chat messages. Note that we'll improve design a bit later:

Screenshot 2022-07-26 at 00 03 05

### DIFF
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -448,16 +448,21 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
   private
 
   def preloaded_chat_message_query
-    ChatMessage
+    query = ChatMessage
       .includes(in_reply_to: [:user, chat_webhook_event: [:incoming_chat_webhook]])
       .includes(:revisions)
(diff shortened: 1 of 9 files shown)


app/controllers/chat_controller.rb
```diff
@@ -452,6 +452,7 @@ def preloaded_chat_message_query
       .includes(in_reply_to: [:user, chat_webhook_event: [:incoming_chat_webhook]])
       .includes(:revisions)
       .includes(:user)
+      .includes(user: :user_status )
```
ZogStriP, line 455: Should we do this only when `SiteSetting.enable_user_status` is enabled?

ZogStriP (approved):

app/controllers/chat_controller.rb
AndrewPrigorshnev: We should! Fixed.

PR #1083, opened 2022-07-24
UX: improves state of the composer on mobile

jjaffeux:


jjaffeux:
I will double check this on Monday, but this should be mostly done.

assets/javascripts/discourse/routes/chat.js
```diff
@@ -25,6 +25,7 @@ export default DiscourseRoute.extend({
   activate() {
     schedule("afterRender", () => {
       document.body.classList.add("has-full-page-chat");
+      document.documentElement.classList.add("full-page-chat");
```
jjaffeux, line 28: eventually I want to move to have only this

assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -680,6 +680,24 @@ export default Component.extend(TextareaTextManipulation, {
     this.onValueChange?.(this.value, this._uploads, this.replyToMsg);
   },
 
+  @action
+  onTextareaFocusin(target) {
```
ZogStriP, line 684: Very minor but I would uppercase the I in "In" since it's two words "Focus In"

```suggestion
  onTextareaFocusIn(target) {
```

assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -680,6 +680,24 @@ export default Component.extend(TextareaTextManipulation, {
...
+    if (!this.capabilities.isIOS) {
+      return;
+    }
+
+    // hack to prevent the whole viewport
+    // to move on focus input
+    target = document.querySelector(".chat-composer-input");
+    target.style.transform = "translateY(-99999px)";
+    target.focus();
+    window.requestAnimationFrame(() => {
```
ZogStriP, line 694: Double RAF? 😱 

Can you add a comment to explain why this is required/how it works?

jjaffeux: there's already the word hack at the top of the block, what else do you want ? :D

assets/javascripts/discourse/components/chat-vh.js
```diff
@@ -0,0 +1,46 @@
...
+    );
+  }
+
+  @bind
+  setVH() {
+    if (this.isDestroying || this.isDestroyed) {
+      return;
+    }
+
+    const composerVH =
```
ZogStriP, line 37: I'm curious: why the `* 0.01`?

jjaffeux: We want to use it as the unit for 1% after

assets/javascripts/discourse/components/chat-composer.js
ZogStriP: I guess if there's an explanation of how the hacks works, it gets easier to know when to delete it.

jjaffeux: AFAIK this is because of this: https://stackoverflow.com/a/44146560

I'm not really sure I want to carry this comment, if you remove this, you should know what you do.

ZogStriP (approved):

PR #1084, opened 2022-07-25
FEATURE: holiday flair in sidebar

lis2:
Show flair when a user participating in direct message is on holiday

Screen Shot 2022-07-25 at 4 19 45 pm

### DIFF
--- a/assets/javascripts/discourse/initializers/chat-sidebar.js
+++ b/assets/javascripts/discourse/initializers/chat-sidebar.js
@@ -10,6 +10,7 @@ import { DRAFT_CHANNEL_VIEW } from "discourse/plugins/discourse-chat/discourse/s
 import { avatarUrl } from "discourse/lib/utilities";
 import { dasherize } from "@ember/string";
 import { emojiUnescape } from "discourse/lib/text";
+import { decorateUsername } from "discourse/helpers/decorate-username-selector";
 
 export default {
   name: "chat-sidebar",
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/initializers/chat-sidebar.js
```diff
@@ -249,12 +250,21 @@ export default {
...
 
+            get oneOnOneMessage() {
+              return this.channel.chatable.users.length === 1;
+            }
+
             get text() {
-              return this.channel.title.replaceAll("@", "");
+              const username = this.channel.title.replaceAll("@", "");
+              if (this.oneOnOneMessage) {
+                return htmlSafe(`${username} ${decorateUsername(username)}`);
```
martin-brennan, line 260: Should `username` be here twice? Or does `decorateUsername(username)` not include the `username` in its output?

martin-brennan (approved):
Minor question but otherwise LGTM

assets/javascripts/discourse/initializers/chat-sidebar.js
lis2: Yes, decorateUsername only contains suffix flairs. 

Example usage of decorate-username-selector is below. We display username and then suffix
https://github.com/discourse/discourse/blob/main/app/assets/javascripts/discourse/app/templates/user-selector-autocomplete.hbr#L7;L9

test/javascripts/acceptance/core-sidebar-test.js
```diff
@@ -56,6 +59,21 @@ acceptance("Discourse Chat - Core Sidebar", function (needs) {
...
+        if (username === "hawk") {
+          return `<span class="on-holiday">${emojiUnescape(
+            ":desert_island:"
+          )}</span>`;
+        }
+      });
+    });
+  });
+  needs.hooks.afterEach(function () {
+    resetUsernameDecorators();
```
tgxworld, line 74: I don't think we need to call this as this has been configured after each test; https://github.com/discourse/discourse/blob/main/app/assets/javascripts/discourse/tests/helpers/qunit-helpers.js#L168

tgxworld (approved):

PR #1086, opened 2022-07-25
FIX: Staff can participate in private categories' chat presence channels.

romanrizzi:
When a staff member joins a channel, and it's not a member of one of the groups with access to it, it should still be able to join the presence
channel.

### DIFF
--- a/app/models/chat_channel.rb
+++ b/app/models/chat_channel.rb
@@ -105,11 +105,21 @@ class ChatChannel < ActiveRecord::Base
   end
 
   def allowed_group_ids
-    category_channel? ? chatable.secure_group_ids : nil
+    return if !category_channel?
+
+    staff_groups = Group::AUTO_GROUPS.slice(:staff, :moderators, :admins).values
+    chatable.secure_group_ids.to_a.concat(staff_groups)
   end
 
   def public_channel_title
     chatable.name
+  end
+
+  def read_restricted?
+    return true if direct_message_channel?
+    return chatable.read_restricted? if category_channel?
+
+    true
   end
 
   def title(user)
--- a/spec/models/chat_channel_spec.rb
+++ b/spec/models/chat_channel_spec.rb
@@ -14,16 +14,43 @@ describe ChatChannel do
   end
 
   describe "#allowed_user_ids" do
-    it "is correct for each channel type" do
+    it "returns participants when the channel is a DM" do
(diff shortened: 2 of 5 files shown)


pmusaraj (approved):
This seems fine, but so I understand properly, `allowed_user_ids` in a private channel will only contain staff ids (presumably, non-staff participants are stored elsewhere)?

romanrizzi:
> This seems fine, but so I understand properly, `allowed_user_ids` in a private channel will only contain staff ids (presumably, non-staff participants are stored elsewhere)?

Participants have access through the groups associated to the category, but staff can join a channel even if it's not a member of one of those groups.

spec/models/chat_channel_spec.rb
```diff
@@ -16,7 +16,14 @@
   describe "#allowed_user_ids" do
     it "is correct for each channel type" do
```
jjaffeux, line 17: I feel like we need more tests with better names here. Basically what we to test:
- direct message has only the allowed user ids of the users in the channel
- a public channel has not allowed user ids
- a private category channel has only staff members as allowed user ids

app/models/chat_channel.rb
```diff
@@ -107,7 +107,11 @@ def chatable_has_custom_fields?
   end
 
   def allowed_user_ids
-    direct_message_channel? ? chatable.user_ids : nil
+    return chatable.user_ids if direct_message_channel?
+
+    staff_memberships =
+      user_chat_channel_memberships.includes(:user).select { |member| member.user.staff? }
+    staff_memberships.empty? ? nil : staff_memberships.map(&:user_id)
```
jjaffeux, line 114: Code golf, I think you can do this for second line:

```
staff_memberships.map(&:user_id).presence
```

Also Im not too concerned about perf here but wouldn't pluck be slightly better?

jjaffeux:
Im wondering if should auto add admin/staff/moderator AUTO groups ids to this list?

romanrizzi:
> Im wondering if we shouldnt auto add admin/staff/moderator AUTO groups ids to this list?

I don't know why I didn't think about this before 🤦

jjaffeux:
> > Im wondering if we shouldnt auto add admin/staff/moderator AUTO groups ids to this list?
> 
> 
> 
> I don't know why I didn't think about this before 🤦 

there are a lot of things to think about :p

pmusaraj (approved):

PR #1087, opened 2022-07-26
FIX: Don’t add users when chatable isn’t present

Flink:
Currently the `AutoManageChannelMemberships` job will add users to a channel following a set of rules. It won’t do anything if the channel doesn’t exist or if the chatable isn’t a category.
But if the chatable doesn’t exist anymore then the job will raise an exception.

This PR addresses this issue by simply checking if the chatable exists before trying to do anything.

### DIFF
--- a/app/jobs/regular/auto_manage_channel_memberships.rb
+++ b/app/jobs/regular/auto_manage_channel_memberships.rb
@@ -10,7 +10,7 @@ module Jobs
           chatable_type: "Category",
         )
 
-      return if !channel
+      return if !channel&.chatable
 
       processed =
         UserChatChannelMembership.where(
--- a/spec/jobs/regular/auto_manage_channel_memberships_spec.rb
+++ b/spec/jobs/regular/auto_manage_channel_memberships_spec.rb
@@ -103,6 +103,17 @@ describe Jobs::AutoManageChannelMemberships do
         end
       end
     end
+
+    context "when chatable doesn’t exist anymore" do
+      before do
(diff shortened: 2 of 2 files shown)


jjaffeux:
Good catch thanks @Flink, can you have a look @romanrizzi please?

Flink:
@jjaffeux Thanks :)
While this patch will avoid the job to run for nothing and crash, we may need to do a followup to take action when the chatable is destroyed. Here I guess for whatever reason the category is destroyed (I caught this because of errors we have in our logs) but it’s not reflected on the `ChatChannel` model linked to it. Not sure what we want to do though, destroy the channel upon category deletion?

jjaffeux:
Hmmm not sure it's complicated because of security. I would say we have no choice but to destroy channels too.

What do we do from the topics when we destroy a category in core? Do we auto assign uncategorized? What if the caterogy was read restricted? I should check this later but feel free to dig this if you want to.

What do you think @mcwumbly ?

Flink:
> I should check this later but feel free to dig this if you want to.

I’ll have a look :+1:

romanrizzi (approved):
Change looks good to me! As you mentioned, we should follow up and decide what to do with the channel when the category gets deleted.

Flink:
> What do we do from the topics when we destroy a category in core? Do we auto assign uncategorized? What if the caterogy was read restricted?

Just had a look and you can’t destroy a category if it has topics in it or subcategories. So it must be empty to be able to delete it.

PR #1091, opened 2022-07-27
UX: inline composer on full-page chat layout

awesomerobot:
Scoped to `desktop` > `has-full-page-chat`

Tested on `Firefox`, `Safari`, & `Chrome`

This inlines the topic composer when you open it on a full-screen chat, and makes chat (and the sidebar if shown) adjust their height to fit. This grants users full access to chat features while using the composer (previously chat input and controls were hidden _behind_ the fix-positioned composer). 

Screen Shot 2022-07-26 at 11 00 35 PM

Screen Shot 2022-07-26 at 11 12 33 PM

### DIFF
--- a/assets/stylesheets/desktop/desktop.scss
+++ b/assets/stylesheets/desktop/desktop.scss
@@ -64,6 +64,7 @@
 }
 
 // Full Page Styling in Core
+
 .has-full-page-chat:not(.discourse-sidebar) {
   --max-chat-width: 1200px;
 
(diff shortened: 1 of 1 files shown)


pmusaraj (approved):
This is elegant and seems to work well. I only have two concerns: 

- this introduces another layout variant for chat, and it may be tricky to test against it and support it in the long run
- have you tested on iPad (Safari and DiscourseHub)? We can also merge and test live, but I believe the desktop styles also apply on iPad 

cc @jjaffeux for visibility

PR #1093, opened 2022-07-27
FIX: use new enable_sidebar setting

lis2:
Related to https://github.com/discourse/discourse/pull/17662

### DIFF
--- a/test/javascripts/acceptance/core-sidebar-test.js
+++ b/test/javascripts/acceptance/core-sidebar-test.js
@@ -14,11 +14,12 @@ import { emojiUnescape } from "discourse/lib/text";
 import User from "discourse/models/user";
 
 acceptance("Discourse Chat - Core Sidebar", function (needs) {
-  needs.user({ experimental_sidebar_enabled: true, has_chat_enabled: true });
+  needs.user({ has_chat_enabled: true });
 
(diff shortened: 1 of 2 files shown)


tgxworld (approved):

tgxworld:
@lis2 Let's quickly merge this in but let us also follow up with a test here.

test/javascripts/acceptance/core-sidebar-test.js
```diff
@@ -14,11 +14,12 @@ import { emojiUnescape } from "discourse/lib/text";
 import User from "discourse/models/user";
 
 acceptance("Discourse Chat - Core Sidebar", function (needs) {
-  needs.user({ experimental_sidebar_enabled: true, has_chat_enabled: true });
```
tgxworld, line 17: Hmm is there a reason why the test didn't fail when the setting was no longer valid?

lis2: It was failing, I just fixed it quickly to have green CI. I have to add test that old sidebar is not visible (that will be next PR)

tgxworld: ahh ok :+1:

PR #1097, opened 2022-07-27
FIX: Adjust reply indicator spacing on iOS

pmusaraj:
Fixes extra space showing below the reply indicator on iOS. The culprit is `safe-area-inset-bottom`, iOS mistakenly still adds a value there even when the keyboard is visible. Thankfully it doesn't look like we need to account for safe areas here, the input is still clear of the device's bottom on iOS. 

### View mode

### Browsers

- [x] safari
- [x] chrome
- [ ] firefox

### Device

- [x] desktop – wide screen
- [x] mobile – DiscourseHub

### DIFF
--- a/assets/stylesheets/common/chat-replying-indicator.scss
+++ b/assets/stylesheets/common/chat-replying-indicator.scss
@@ -4,9 +4,7 @@
 
 .chat-replying-indicator {
   color: var(--primary-medium);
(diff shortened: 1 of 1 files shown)


jjaffeux (approved):

jjaffeux:
Awesome, thanks for looking this, very helpful ❤️

PR #1102, opened 2022-07-29
DEV: Add created_at index for ChatMessage

martin-brennan:
Follow up to ddab84061ccfa2003fa52e73b1c364a7014c47e4,
there could be a lot of messages in the chat message
table, so we want to make sure these stat queries are
using an index.

### DIFF
--- a/db/migrate/20220729032237_add_index_to_chat_message_created_at.rb
+++ b/db/migrate/20220729032237_add_index_to_chat_message_created_at.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddIndexToChatMessageCreatedAt < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :chat_messages, :created_at, algorithm: :concurrently
+  end
+end
--- a/db/migrate/20220729032237_add_index_to_chat_message_created_at.rb
+++ b/db/migrate/20220729032237_add_index_to_chat_message_created_at.rb
@@ -4,6 +4,17 @@ class AddIndexToChatMessageCreatedAt < ActiveRecord::Migration[7.0]
   disable_ddl_transaction!
 
   def change
-    add_index :chat_messages, :created_at, algorithm: :concurrently
+    execute <<~SQL
+    CREATE INDEX CONCURRENTLY IF NOT EXISTS
+    idx_chat_messages_by_created_at_not_deleted
+    ON chat_messages (created_at)
+    WHERE deleted_at IS NULL
+    SQL
+  end
+
+  def down
+    execute <<~SQL
+    DROP INDEX IF EXISTS idx_chat_messages_by_created_at_not_deleted
+    SQL
   end
 end
--- a/app/models/chat_message.rb
+++ b/app/models/chat_message.rb
@@ -220,6 +220,6 @@ end
 #
 # Indexes
 #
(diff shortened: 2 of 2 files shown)


martin-brennan:
Doesn't _seem_ like this is being used in tests, so maybe I need deleted_at too:

```
Seq Scan on chat_messages  (cost=0.00..17.00 rows=1 width=116) (actual time=0.006..0.007 rows=1 loops=1)\n  Filter: ((deleted_at IS NULL) AND (created_at > '2022-07-07 09:30:00'::timestamp without time zone))\n  Rows Removed by Filter: 10\nPlanning Time: 0.038 ms\nExecution Time: 0.016 ms
```

tgxworld:
> (deleted_at IS NULL)

For the stats, I think we should be counting deleted messages too because chat messages are deleted at some periodic interval but from the stats perspective, those are still considered chat messages which have been created.

> Doesn't seem like this is being used in tests, so maybe I need deleted_at too:

You'll need to test on an instance with more rows. Since there is only 1 row in the table, a seq scan is more efficient.

martin-brennan:
> For the stats, I think we should be counting deleted messages too because chat messages are deleted at some periodic interval but from the stats perspective, those are still considered chat messages which have been created.

No we aren't doing this intentionally, we have discussed this internally, all other About stats do not count deleted.

Also chat messages deleted by the retention policies are _permanently_ deleted, not trashed.

> You'll need to test on an instance with more rows. Since there is only 1 row in the table, a seq scan is more efficient.

Ah OK will try that and see.

tgxworld:
> Also chat messages deleted by the retention policies are permanently deleted, not trashed.

O wouldn't this be a problem if I have a retention policy of 10 days or even 1 day. The last 7/30 days stats are just going to reflect the current day.

db/migrate/20220729032237_add_index_to_chat_message_created_at.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddIndexToChatMessageCreatedAt < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :chat_messages, :created_at, algorithm: :concurrently
```
tgxworld, line 7: The current discussion aside, since the `chat_messages` table is likely to be big, I'll recommend adding a `WHERE deleted_at IS NULL` filter to this index to keep it small and efficient.

martin-brennan: Yep have already changed it to that locally, just trying to trigger it now with a larger record set.

martin-brennan:
> O wouldn't this be a problem if I have a retention policy of 10 days or even 1 day. The last 7/30 days stats are not going to be accurate.

Yes we've discussed this internally already and the tradeoff is acceptable for now, see t/69224/28

martin-brennan:
@tgxworld changed the index to use deleted_at, and it is being used correctly on a larger data set:

```
 Index Scan using idx_chat_messages_by_created_at_not_deleted on chat_messages  (cost=0.28..225.70 rows=96 width=186) (actual time=0.056..1.144 rows=1134 loops=1)
   Index Cond: (created_at > '2022-06-29 04:04:03.369917'::timestamp without time zone)
 Planning Time: 1.563 ms
 Execution Time: 1.357 ms
(4 rows)
```

tgxworld (approved):

martin-brennan:
Thanks @tgxworld !

PR #1108, opened 2022-07-30
PERF: prevents N+1 on public channels due to topic url

jjaffeux:


jjaffeux:
On the same state of channels in a local instance:

Before:
Screenshot 2022-07-30 at 17 29 20
Screenshot 2022-07-30 at 17 29 49

After:
Screenshot 2022-07-30 at 17 28 30
Screenshot 2022-07-30 at 17 26 07

CvX (approved):

PR #1109, opened 2022-07-31
FIX: restores presence and displays it on sidebar

jjaffeux:


CvX (approved):

jjaffeux:
I have a very big PR coming which is homogenizing routing for drawer and full page but I still more time, this will stop the bleeding for now.

PR #1117, opened 2022-08-02
FIX: Do not allow chat uploads if secure_media is enabled

martin-brennan:
To avoid unexpected behaviour, since we do not currently
support securing chat uploads via the `secure_media` setting
we are disabling chat uploads if `secure_media` is also enabled
via an initial migration. We also disable chat uploads if `secure_media`
is changed from disabled to enabled, and we prevent changing
`chat_allow_uploads` to true via a site setting validator if `secure_media`
is enabled.

This condition can be overridden with `DISCOURSE_ALLOW_UNSECURE_CHAT_UPLOADS`
at operator risk. This condition will be removed once secure media
rework to allow for chat uploads is completed later in the year.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -17,6 +17,7 @@ en:
     default_emoji_reactions: "Default emoji reactions for chat messages. Add up to 5 emojis for quick reaction."
     errors:
       chat_default_channel: "The default chat channel must be a public channel."
+      chat_upload_not_allowed_secure_media: "Chat uploads are not allowed if secure media is enabled."
   system_messages:
     chat_channel_archive_complete:
       title: "Chat Channel Archive Complete"
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -17,7 +17,7 @@ en:
     default_emoji_reactions: "Default emoji reactions for chat messages. Add up to 5 emojis for quick reaction."
     errors:
       chat_default_channel: "The default chat channel must be a public channel."
-      chat_upload_not_allowed_secure_media: "Chat uploads are not allowed if secure media is enabled."
+      chat_upload_not_allowed_secure_media: "Chat uploads are not allowed when secure media site setting is enabled."
   system_messages:
     chat_channel_archive_complete:
       title: "Chat Channel Archive Complete"
--- a/db/migrate/20220802014549_disable_chat_uploads_if_secure_media_enabled.rb
+++ b/db/migrate/20220802014549_disable_chat_uploads_if_secure_media_enabled.rb
@@ -15,7 +15,7 @@ class DisableChatUploadsIfSecureMediaEnabled < ActiveRecord::Migration[7.0]
     secure_media_enabled =
       DB.query_single("SELECT value FROM site_settings WHERE name = 'secure_media'").first == "t"
 
-    if secure_media_enabled && chat_uploads_enabled && !ENV["DISCOURSE_ALLOW_UNSECURE_CHAT_UPLOADS"]
+    if secure_media_enabled && chat_uploads_enabled && !GlobalSetting.allow_unsecure_chat_uploads
       if chat_allow_uploads_value.nil?
         DB.exec(
           "
--- a/db/migrate/20220802014549_disable_chat_uploads_if_secure_media_enabled.rb
+++ b/db/migrate/20220802014549_disable_chat_uploads_if_secure_media_enabled.rb
@@ -9,16 +9,20 @@ class DisableChatUploadsIfSecureMediaEnabled < ActiveRecord::Migration[7.0]
   # The env var DISCOURSE_ALLOW_UNSECURE_CHAT_UPLOADS can be set to keep
   # it enabled, but this is strongly advised against.
   def up
-    chat_allow_uploads_value = DB.query_single("SELECT value FROM site_settings WHERE name = 'chat_allow_uploads'").first
+    chat_allow_uploads_value =
+      DB.query_single("SELECT value FROM site_settings WHERE name = 'chat_allow_uploads'").first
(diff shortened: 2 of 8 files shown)


spec/validators/chat_allow_uploads_validator_spec.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require "rails_helper"
```
tgxworld, line 3:
```suggestion
```

This is no longer required.

spec/validators/chat_allow_uploads_validator_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+describe ChatAllowUploadsValidator do
```
tgxworld, line 5:
```suggestion
RSpec.describe ChatAllowUploadsValidator do
```

Just a minor note but in RSpec 4, we need to use `RSpec.describe` as `describe` will be deprecated.

lib/validators/chat_allow_uploads_validator.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class ChatAllowUploadsValidator
+  def initialize(opts = {})
+    @opts = opts
+  end
+
+  def valid_value?(value)
+    return true if value == "f"
```
tgxworld, line 9:
```suggestion
```

If I'm not wrong, you can remove this line and we will still get the same behaviour.

lib/validators/chat_allow_uploads_validator.rb
```diff
@@ -0,0 +1,21 @@
...
+    if value == "t" && SiteSetting.secure_media && !ENV["DISCOURSE_ALLOW_UNSECURE_CHAT_UPLOADS"]
+      return false
+    end
+    true
+  end
+
+  def error_message
+    if SiteSetting.secure_media && !ENV["DISCOURSE_ALLOW_UNSECURE_CHAT_UPLOADS"]
```
tgxworld, line 17: I think we should extract the conditional logic here in to a method instead of duplicating it in two different methods.

config/locales/server.en.yml
```diff
@@ -17,6 +17,7 @@ en:
     default_emoji_reactions: "Default emoji reactions for chat messages. Add up to 5 emojis for quick reaction."
     errors:
       chat_default_channel: "The default chat channel must be a public channel."
+      chat_upload_not_allowed_secure_media: "Chat uploads are not allowed if secure media is enabled."
```
tgxworld, line 20:
```suggestion
      chat_upload_not_allowed_secure_media: "Chat uploads are not allowed if secure media site setting is enabled."
```

db/migrate/20220802014549_disable_chat_uploads_if_secure_media_enabled.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class DisableChatUploadsIfSecureMediaEnabled < ActiveRecord::Migration[7.0]
+  ##
+  # At this point in time, secure media is not compatible with discourse-chat,
+  # so if it is enabled then chat uploads must be disabled to avoid undesirable
+  # behaviour.
+  #
+  # The env var DISCOURSE_ALLOW_UNSECURE_CHAT_UPLOADS can be set to keep
+  # it enabled, but this is strongly advised against.
```
tgxworld, line 10: I don't think the ENV should be the concern of a migration. Given that most people will not even have the chance to configure the ENV before the migration is run, I think we can just remove the ENV and manually update in places where it is necessary?

spec/validators/chat_allow_uploads_validator_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    it "does not allow chat uploads to be enabled" do
+      validator = described_class.new
+      expect(validator.valid_value?("t")).to eq(false)
+      expect(validator.error_message).to eq(
+        I18n.t("site_settings.errors.chat_upload_not_allowed_secure_media"),
+      )
+    end
+
+    it "does allow chat uploads to be enabled if DISCOURSE_ALLOW_UNSECURE_CHAT_UPLOADS is set" do
+      ENV["DISCOURSE_ALLOW_UNSECURE_CHAT_UPLOADS"] = "t"
```
tgxworld, line 28: Should we specify the value that the ENV should set? Another spec sets it to `true` while this spec sets it to `t`.

martin-brennan: Irrelevant now I have changed to GlobalSetting

db/migrate/20220802014549_disable_chat_uploads_if_secure_media_enabled.rb
martin-brennan: Irrelevant now I have changed to GlobalSetting, also migration is fine IMO we want to automate things like this not do manual busywork. This is not really intended for average users, neither is secure_media in general really

spec/validators/chat_allow_uploads_validator_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    it "does not allow chat uploads to be enabled" do
+      validator = described_class.new
+      expect(validator.valid_value?("t")).to eq(false)
+      expect(validator.error_message).to eq(
+        I18n.t("site_settings.errors.chat_upload_not_allowed_secure_media"),
+      )
+    end
+
+    it "does allow chat uploads to be enabled if allow_unsecure_chat_uploads global setting is set" do
```
tgxworld, line 23:
```suggestion
    it "allows chat uploads to be enabled if allow_unsecure_chat_uploads global setting is enabled" do
```

tgxworld (approved):

PR #1118, opened 2022-08-02
DEV: Fix RUBYOPT for syntax-tree linting

martin-brennan:
-W0 suppresses all warnings, _including_ the syntax-tree
ones, whereas we just want to suppress these experimental
feature warnings to stop noise:

> warning: Pattern matching is experimental, and the behavior may change in future versions of Ruby!

Followup to 2c7bdf674322f8dac7d6566b879802171eac27da


martin-brennan:
Test failure unrelated flaky

tgxworld (approved):

PR #1121, opened 2022-08-02
FIX: Don't stage sent message if there are more to load.

romanrizzi:
When a user sends a message, we upsert it at the end and force a scroll to the future, assuming we have all the latest messages loaded. However, this is not
always the case, which causes the scroll to load old messages.

This PR handles this scenario by loading the latest, then scrolling to the newly created message.


ZogStriP (commented):
Any chances this could be tested, looks like a bad UX we definitely don't want to see regress.

ZogStriP (approved):

PR #1122, opened 2022-08-02
REFACTOR: Track message visibility with an  Ember modifier

romanrizzi:


jjaffeux:
@romanrizzi we talked of this with @davidtaylorhq and we propose we try the following:

- Create a `ChatLivePaneInterserctionObserver` service. It would have document for root, hold the instance of the InstersectionObserve and the callback logic, and expose observe/unobserve
- The modifier would be on each message we inspect, we would inject the service in the modifier (https://github.com/ember-modifier/ember-modifier#example-with-service-injection) and call observe/unobserve on the service as needed

Does it make sense to you?

davidtaylorhq:
I can't find any documentation on testing modifiers, but this looks like a good example: https://github.com/emberjs/ember-render-modifiers/blob/master/tests/integration/modifiers/did-insert-test.js

romanrizzi:
Cool! I'll switch to a service instead.

davidtaylorhq:
@romanrizzi @jjaffeux you may be interested in https://github.com/discourse/discourse/pull/17777

I've introduced https://www.npmjs.com/package/ember-on-resize-modifier, which also introduces a companion `resizeObserver` service. Could be a useful reference for this Chat work. Source can be found here: https://github.com/PrecisionNutrition/ember-resize-kitchen-sink/tree/main/packages

test/javascripts/modifiers/track-message-visibility-test.js
```diff
@@ -0,0 +1,36 @@
...
+      await render(template);
+      await settled();
+
+      assert.ok(
+        exists("div[data-visible=true]"),
+        "message is marked as visible"
+      );
+    });
+
+    test("Marks message as visible when it doesn't intersect with the viewport", async function (assert) {
```
romanrizzi, line 24: I can't understand why this one fails. 😢 

 It's super flaky when I run it locally. I tried setting the `display: none;`after rendering it to force the observe to re-run, but it's not working either. Do you have any ideas, @davidtaylorhq?

davidtaylorhq: `await settled()` can only wait for things that are pending in the Ember-managed RunLoop. It doesn't have any way of waiting for any pending browser-managed events like the IntersectionObserver callbacks.

I think the easiest solution here is probably the [`waitFor()`](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#waitfor) helper. So, instead of `await settled()`, we would do

```
import { waitFor } from "@ember/test-helpers";

...

await waitFor("div[data-visible=false]");
```

Making that change in both tests seems to solve the flakiness for me on local 🤞

assets/javascripts/discourse/services/chat-message-visibility-observer.js
```diff
@@ -0,0 +1,35 @@
...
+      rootMargin: "-10px",
+    };
+
+    const callback = (entries) => {
+      entries.forEach((entry) => {
+        entry.target.dataset.visible = entry.isIntersecting;
+      });
+    };
+
+    this.observer = new IntersectionObserver(callback, options);
```
davidtaylorhq, line 21: I wonder if it would be cleaner to make `callback` a method on the service, rather than an anonymous function?

assets/javascripts/discourse/templates/components/chat-message.hbs
```diff
@@ -20,6 +20,7 @@
   {{on "touchend" this.handleTouchEnd passive=true}}
   {{on "mouseenter" (fn this.onHoverMessage message (hash desktopOnly=true))}}
   {{on "mouseleave" (fn this.onHoverMessage null (hash desktopOnly=true))}}
+  {{track-message-visibility}}
```
davidtaylorhq, line 23: Modifiers can be namespaced using directories (just the same as components), so we could make this `{{chat/track-message-visibility}}`

test/javascripts/components/chat-message-test.js
```diff
@@ -116,4 +116,16 @@ module("Discourse Chat | Component | chat-message", function (hooks) {
       "has the correct hidden css class and expand button within"
     );
   });
+
+  test("Message marked as visible", async function (assert) {
+    this.setProperties(generateMessageProps());
+
+    await render(template);
+    await settled();
```
davidtaylorhq, line 124: As mentioned in the other comment, let's make this `waitFor` to reduce the chance of flakiness

davidtaylorhq (commented):
This looks great to me - just a few small comments.

In future, I wonder if we could improve this even further. Right now, marking something read is a two-step process:

1. IntersectionObserver fires, and sets data-visible

2. chat-live-pane sets up an `scroll` listener. 1 second after scroll, check for `data-visible` messages in the DOM, then do an `ajax` to update the server

Kinda feels like we should cut out the `data-visible`/scroll/timer part of this and go straight from `IntersectionObserver` -> (`channel` model)? -> `ajax`?

Not a blocker for this PR - just thinking out loud here 🤔

assets/javascripts/discourse/services/chat-message-visibility-observer.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  observe(element) {
+    this.observer.observe(element);
+  }
+
+  unobserve(element) {
+    this.observer.unobserve(element);
+  }
+}
```
davidtaylorhq, line 35: We should implement a `willDestroy()` method here which cleans up the `IntersectionObserver` instance. (services are created/destroyed at the beginning/end of every test)

romanrizzi:
`waitFor` was the missing piece. Thanks, @davidtaylorhq!

> Kinda feels like we should cut out the data-visible/scroll/timer part of this and go straight from IntersectionObserver -> (channel model)? -> ajax?

I agree we could simplify this even further. One thing that I like about the current approach is that it only fires on specific events, and we're not constantly queueing and canceling updates, so I wonder if we could implement something along these lines on the chat service.

davidtaylorhq (approved):

PR #1126, opened 2022-08-03
DEV: fix tests

jjaffeux:
Automatically correct the title of a channel has exposed multiple incorrect tests. This should bring us to a sane state.


davidtaylorhq (approved):
🚀

PR #1127, opened 2022-08-03
FIX: Mark message as visible as soon as the first line appears in the viewport

romanrizzi:


assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1457,8 +1457,8 @@ export default Component.extend({
   _setupVisibleMessagesObserver() {
     const options = {
       root: document.querySelector(".chat-messages-container"),
-      rootMargin: "0px",
-      threshold: 0.6,
+      rootMargin: "-30px",
+      threshold: 0,
```
jjaffeux, line 1461: I think this is the default

jjaffeux (approved):

jjaffeux:
:tada: thanks

PR #1129, opened 2022-08-03
FEATURE: Keyboard shortcut to toggle open/close of chat drawer

keegangeorge:
This PR adds the functionality to:
- open the chat drawer using the key <kbd>-</kbd>
- close the chat drawer while focused on the input using the key <kbd>esc</kbd>

It also adds the details to the shortcut menu:
Screen Shot 2022-08-03 at 3 48 35 PM

### DIFF
--- a/test/javascripts/acceptance/chat-keyboard-shortcuts-test.js
+++ b/test/javascripts/acceptance/chat-keyboard-shortcuts-test.js
@@ -259,18 +259,20 @@ acceptance("Discourse Chat - Keyboard shortcuts", function (needs) {
 
     await triggerKeyEvent(document.body, "keydown", "-");
     await settled();
-    assert.ok(visible(".topic-chat-float-container"), "chat float is open");
+    assert.ok(exists(".topic-chat-drawer-content"), "chat float is open");
   });
 
   test("Escape to close chat float", async function (assert) {
     await visit("/latest");
     this.chatService.set("sidebarActive", false);
     this.chatService.set("chatWindowFullPage", false);
+
     await click(".header-dropdown-toggle.open-chat");
     await settled();
-    assert.ok(
-      exists(".topic-chat-float-container.hidden"),
-      "chat drawer hidden"
-    );
+
+    const composerInput = query(".chat-composer-input");
+    await focus(composerInput);
+    await triggerKeyEvent(document.body, "keydown", "Escape");
+    assert.ok(!exists(".topic-chat-drawer-content"), "chat float is closed");
   });
 });
--- a/test/javascripts/acceptance/chat-keyboard-shortcuts-test.js
+++ b/test/javascripts/acceptance/chat-keyboard-shortcuts-test.js
@@ -251,4 +251,26 @@ acceptance("Discourse Chat - Keyboard shortcuts", function (needs) {
       "modal dismissed after submitting link"
     );
   });
+
+  test("toggle chat in float", async function (assert) {
+    await visit("/latest");
+    this.chatService.set("sidebarActive", false);
+    this.chatService.set("chatWindowFullPage", false);
+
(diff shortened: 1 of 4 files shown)


martin-brennan (approved):
Very nice and thank you for the detailed PR description, exactly what we want to see :+1:

test/javascripts/acceptance/chat-keyboard-shortcuts-test.js
```diff
@@ -251,4 +251,26 @@ acceptance("Discourse Chat - Keyboard shortcuts", function (needs) {
...
+
+    await triggerKeyEvent(document.body, "keydown", "-");
+    await settled();
+    assert.ok(visible(".topic-chat-float-container"), "chat float is open");
+  });
+
+  test("Escape to close chat float", async function (assert) {
+    await visit("/latest");
+    this.chatService.set("sidebarActive", false);
+    this.chatService.set("chatWindowFullPage", false);
```
martin-brennan, line 268: Oh you missed the `triggerKeyEvent` for escape here, probably why the test failed

keegangeorge: Thanks @martin-brennan! 

I've added that back, but it's still failing the test unfortunately. I've tried using different selectors too. However, it doesn't seem to be working. Do you know what I might be missing?

CvX: It might not be a test issue - does it work in the app? Global shortcuts don't work when input fields are focused. 😃

martin-brennan: Yeah it worked locally when I checked it out

martin-brennan: > Global shortcuts don't work when input fields are focused. 

They do, since it works for the chat composer's other shortcuts :) And it worked for the escape one locally too

CvX: Ah, I got it backwards! This means `-` binding shouldn't be global then, otherwise typing this character in any input field will open the chat drawer, right?

martin-brennan: Oh yes you are right! That is a pickle :thinking: @keegangeorge you have to make sure there is no `document.activeElement` I guess when `-` is pressed so the chat drawer doesn't open if you are already focused on another input

test/javascripts/acceptance/chat-keyboard-shortcuts-test.js
```diff
@@ -251,4 +251,28 @@ acceptance("Discourse Chat - Keyboard shortcuts", function (needs) {
...
+
+    await click(".header-dropdown-toggle.open-chat");
+    await settled();
+
+    const composerInput = query(".chat-composer-input");
+    await focus(composerInput);
+    await triggerKeyEvent(document.body, "keydown", "Escape");
```
martin-brennan, line 275: I think instead of triggering the key event on `document.body`, it has to do `triggerKeyEvent(composerInput)` since that needs to be focused and receiving the key event to have this shortcut work

keegangeorge: Thanks! This fixed it 😄 Tests are passing now

test/javascripts/acceptance/chat-keyboard-shortcuts-test.js
keegangeorge: @CvX Thanks for catching this and @martin-brennan thanks for the advice on how to handle it. I used the `element.target` since the element was already being passed. Let me know what you think.

jjaffeux:
LGTM thanks ❤️ Let's merge this.

PR #1131, opened 2022-08-04
FIX: Moving memberships out of channel decorator

martin-brennan:
This commit makes a few key changes around how membership-related
data for chat channels is serialized, and cleans up the `ChatChannelFetcher`.

I have added a `current_user_membership` to the `ChatChannelSerializer`
which is used to store the current user's `UserChatChannelMembership`
data for that channel rather than storing that data on top level model attributes
for the channel. This has necessitated a lot of changes to the frontend and
tests and fixtures.

Also enabled by this change is the removal of the `previewing` property
in all the chat components -- instead we can now rely on the
`channel.current_user_membership.following` boolean which gets rid
of the bug where if there were more than N channels in the sidebar we
would not consider some of them followed even if the user was a member.

Finally I got rid of some deprecated usages of `current-user:main`, changing
to `service:current-user`, and same for the `session` service.


app/serializers/chat_channel_settings_serializer.rb
```diff
@@ -1,5 +0,0 @@
-# frozen_string_literal: true
```
martin-brennan, line 1: This file was not even used anywhere

assets/javascripts/discourse/models/chat-channel.js
```diff
@@ -124,12 +125,18 @@ const ChatChannel = RestModel.extend({
 
     return this.memberships_count;
   },
+
+  @computed("current_user_membership.following")
+  get isFollowing() {
```
martin-brennan, line 130: Not sure if it is worth also making shortcuts like `isMuted` for these types of things, `isFollowing` was just the most used

app/queries/chat_channel_memberships_query.rb
```diff
@@ -5,6 +5,7 @@ def self.call(channel, limit: 50, offset: 0, username: nil)
     query =
       UserChatChannelMembership
         .joins(:user)
+        .includes(:user)
```
martin-brennan, line 8: Fixes an N1 on the members list

app/serializers/chat_channel_serializer.rb
```diff
@@ -88,6 +83,12 @@ def include_auto_join_users?
     scope.can_edit_chat_channel?
   end
 
+  def current_user_membership
+    return if !@current_user_membership
+    @current_user_membership.chat_channel = object
```
martin-brennan, line 88: This is kind of weird but we have the `chat_channel` already so it avoids a separate query

lib/chat_channel_fetcher.rb
```diff
@@ -64,6 +71,7 @@ def self.secured_public_channels(guardian, memberships, options = { following: t
           "LEFT JOIN categories ON categories.id = chat_channels.chatable_id AND chat_channels.chatable_type = 'Category'",
         )
         .where(chatable_type: ChatChannel.public_channel_chatable_types)
+        .where("chat_channels.id IN (#{generate_allowed_channel_ids_sql(guardian)})")
```
martin-brennan, line 74: This is done rather than having the `next if !guardian.can_see_chat_channel?(channel)` inside the decorator loops

lib/chat_channel_fetcher.rb
```diff
@@ -171,22 +130,36 @@ def self.secured_direct_message_channels(user_id, memberships, guardian)
         UserField.all.pluck(:id).map { |fid| "#{User::USER_FIELD_PREFIX}#{fid}" }
     User.preload_custom_fields(channels.map { |c| c.chatable.users }.flatten, preload_fields)
 
-    unread_counts_per_channel = unread_counts(channels, user_id)
+    decorate_memberships_with_tracking_data(guardian, channels, memberships)
+  end
 
-    channels.filter_map do |channel|
-      next if !guardian.can_see_chat_channel?(channel)
+  def self.decorate_memberships_with_tracking_data(guardian, channels, memberships)
```
martin-brennan, line 136: Okay I am still not really happy with how this part is, trying to think of a different way of doing this. Now I've moved `unread_count` and `unread_mentions` down to the membership level as model attributes I have to decorate them still. Maybe we have a sub-section of the channel serializer for current user tracking state as well to put this? Not sure...also really not a fan that we have to parse the notification JSON data out

lib/chat_channel_fetcher.rb
```diff
@@ -171,22 +130,36 @@ def self.secured_direct_message_channels(user_id, memberships, guardian)
...
 
-      channel
+      if membership
+        membership.unread_mentions =
+          mention_notification_data.count do |data|
+            data["chat_channel_id"] == channel.id &&
+              data["chat_message_id"] > (membership.last_read_message_id || 0)
+          end
+
+        # direct message channels cannot be muted, so we always need the unread count
```
martin-brennan, line 156: This will of course change with a later TODO

assets/javascripts/discourse/models/chat-channel.js
jjaffeux: Yep I'm fine just adding them later as the need arises

app/models/chat_channel.rb
```diff
@@ -31,9 +31,9 @@ def add(user)
         UserChatChannelMembership.find_or_initialize_by(user_id: user.id, chat_channel: self)
 
       if !membership.following
+        update!(user_count: (user_count || 0) + 1)
```
jjaffeux, line 34: This line and the change in the file is to ensure we have updated user count before generating the membership object which will be used in the serializer, that way we get the correct count from the channel, as the membership is pulling it from here

assets/javascripts/discourse/lib/chat-api.js
```diff
@@ -50,7 +50,7 @@ export default class ChatApi {
         method: "POST",
       }
     ).then((membership) => {
-      channel.updateMembership(false, membership);
+      channel.updateMembership(membership);
```
jjaffeux, line 53: I don't think we need this first param anymore

assets/javascripts/discourse/models/chat-channel.js
```diff
@@ -66,14 +66,15 @@ const ChatChannel = RestModel.extend({
...
     this.current_user_membership.setProperties({
-      following,
+      following: membership.following,
       muted: membership.muted,
       desktop_notification_level: membership.desktop_notification_level,
       mobile_notification_level: membership.mobile_notification_level,
-      memberships_count: membership.user_count,
+      user_count: membership.user_count,
     });
+    this.set("memberships_count", membership.user_count);
```
jjaffeux, line 77: This is a weird case we have to handle, as ultimately channel is owner of this value but here we only get an updated membership and we have to reflect the new count to the parent channel

martin-brennan: No longer needed now, removed ;)

jjaffeux: 🕺

jjaffeux (approved):

PR #1133, opened 2022-08-04
FIX: various optimisations on mobile for live pane

jjaffeux:
- uses scrollIntoView instead of custom code
- prevents scrollable list to lag behind rendering by removing relative where possible and using `-webkit-overflow-scrolling: touch;` and ` -webkit-transform: translateZ(0px);` on relative element
- displays loading bar on initial loading
- applies safari white screen fix only on intentional scroll and not when automatically filling pane
- delays onHoverMessage to improve perf when scrolling on desktop
- removes unexpected border radius


jjaffeux:
I can't really test most of this as it's a lot of weird hacks for iOS, and live pane is quite untestable until it goes under heavy refactoring.

ZogStriP (approved):

PR #1136, opened 2022-08-05
FIX: removes ideal first channel logic from unfollowChatChannel

jjaffeux:
This logic shouldn't be here, but in openChannel, this is causing various bugs and doing more work than necessary ATM, eg: when unfollowing a channel currently it will first go to index, pick ideal first channel and then it will open the channel you just unfollowed. Removing this logic will avoid all of this.


AndrewPrigorshnev (approved):

jjaffeux:
This is already tested in the App, I just removed behavior which was not tested

PR #1138, opened 2022-08-05
FEATURE: Show user statuses in search results when sending a direct message

AndrewPrigorshnev:
The status will appear after merging https://github.com/discourse/discourse/pull/17716 and https://github.com/discourse/discourse/pull/17809 in Core.


jjaffeux:
Can we add a test in direct-message-creator-test please?

AndrewPrigorshnev:
> Can we add a test in direct-message-creator-test please?

Sure thing. I've added a test, but it will start passing only after we merge https://github.com/discourse/discourse/pull/17716 in Core.

AndrewPrigorshnev:
I also realized that I forgot to add styles to status on the `user-info` component in Core. I'll mark this PR with the do-not-merge label for a short time, will fix styles in Core, and then will merge this.

jjaffeux (approved):

PR #1143, opened 2022-08-07
DEV: Remove unnecessary `await settled()`

CvX:


jjaffeux:
What do we do with this?

gschlager (approved):

PR #1145, opened 2022-08-07
DEV: Remove unused fixture code

CvX:


test/javascripts/chat-fixtures.js
```diff
@@ -196,7 +195,7 @@ export const chatChannels = {
 function addSettingsAttrs(channel) {
   channel.current_user_membership = channel.current_user_membership || {};
   channel.current_user_membership.unread_count = 0;
-  channel.current_user_membership.unrad_mentions = 0;
+  channel.current_user_membership.unread_mentions = 0;
```
tgxworld, line 198: Does this typo not affect any features?

jjaffeux: Most of unread stuff is very badly tested, so iI wouldn't be surprised, this is just setting some general stuff and probably most of it is not used. We can keep it though as it's a valid expectation to have this here, will remove it once we move to fabricators.

jjaffeux (approved):

test/javascripts/chat-fixtures.js
CvX: Hm, actually the whole thing seems to be unused 👀

jjaffeux:
I already have this removed in upcoming fabricator pr, fine for this one but please dont spend too much time hunting for this

jjaffeux (approved):

PR #1148, opened 2022-08-08
Revert "DEV: Fix `site:main` deprecation"

ZogStriP:
Reverts discourse/discourse-chat#1144

This was generating the following exception: "Uncaught TypeError: Cannot read properties of undefined (reading 'lookup')" when trying to create a new "personal chat".

### DIFF
--- a/assets/javascripts/discourse/initializers/chat-sidebar.js
+++ b/assets/javascripts/discourse/initializers/chat-sidebar.js
@@ -372,7 +372,7 @@ export default {
             }
 
             get actions() {
-              const site = getOwner(this).lookup("site:main");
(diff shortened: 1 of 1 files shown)


oblakeerickson (approved):

assets/javascripts/discourse/initializers/chat-sidebar.js
```diff
@@ -373,13 +372,14 @@ export default {
             }
 
             get actions() {
+              const site = getOwner(this).lookup("site:main");
```
davidtaylorhq, line 375: `site:main` is deprecated
```suggestion
              const site = getOwner(this).lookup("service:site");
```

PR #1149, opened 2022-08-08
FIX: implements DeletedChatUser as a placeholder to deleted users

jjaffeux:
This solution allows to implement all the logic server side and let the client side fully functional without any modification.

The only known limit of this solution is the hardcoded username "deleted". If needed we could implement a site setting to let users modify it.


jjaffeux:
Screenshot 2022-08-08 at 16 39 07
Screenshot 2022-08-08 at 16 39 09
Screenshot 2022-08-08 at 16 40 30

app/models/deleted_chat_user.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class DeletedChatUser < User
+  def username
+    "deleted"
```
ZogStriP, line 5: Should this be `I18n`able?

app/models/direct_message_channel.rb
```diff
@@ -9,7 +9,10 @@ def user_can_access?(user)
   end
 
   def chat_channel_title_for_user(chat_channel, acting_user)
-    users = direct_message_users.map(&:user) - [acting_user]
+    users =
+      (direct_message_users.map(&:user) - [acting_user]).map do |user|
+        user ? user : DeletedChatUser.new
```
ZogStriP, line 14: I _think_ this should work as well?

```suggestion
        user || DeletedChatUser.new
```

app/serializers/direct_message_channel_serializer.rb
```diff
@@ -4,7 +4,7 @@ class DirectMessageChannelSerializer < ApplicationSerializer
   has_many :users, serializer: UserWithCustomFieldsAndStatusSerializer, embed: :objects
 
   def users
-    users = object.direct_message_users.map(&:user)
+    users = object.direct_message_users.map(&:user).map { |u| u ? u : DeletedChatUser.new }
```
ZogStriP, line 7:
```suggestion
    users = object.direct_message_users.map(&:user).map { |u| u || DeletedChatUser.new }
```

app/models/direct_message_channel.rb
jjaffeux: It should yes 😅 too much js

spec/models/deleted_chat_user_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe DeletedChatUser do
+  describe "#username" do
+    it "returns a default username" do
+      expect(subject.username).to eq("deleted")
```
ZogStriP, line 8: `I18n` as well 😉

spec/models/direct_message_channel_spec.rb
```diff
@@ -53,14 +53,16 @@
...
+      it "returns a placeholder username" do
+        direct_message_channel = Fabricate(:direct_message_channel, users: [user1, user2])
+        user2.destroy!
+        direct_message_channel.reload
 
-      expect(
-        direct_message_channel.chat_channel_title_for_user(chat_channel, user1),
-      ).to eq chat_channel.id
+        expect(direct_message_channel.chat_channel_title_for_user(chat_channel, user1)).to eq(
+          "@deleted",
```
ZogStriP, line 63: `I18n`?

jjaffeux:
Showing as deleted by default:

Screenshot 2022-08-08 at 17 59 30

ZogStriP (approved):

PR #790, opened 2022-04-21
DEV: Move chat selection buttons out of chat-live-pane

martin-brennan:
At the moment there is just a Quote button when selecting messages,
but we are adding another button here and potentially more
in future, it makes more sense to have the chat selection
management buttons to be in their own component, not clogging
up chat-live-pane.

PR #1150, opened 2022-08-08
FIX: Restore/fix quote-in-topic and copy-quote features

CvX:
1. Restored the message (regressed in #1146)
2. Fixed the animation (regressed in #790)
3. Fixed element placement
4. Restored and updated tests
5. Restored "quote in topic" on desktop (regressed in #974)
6. Exposed "copy quote" as a separate button

---

### View mode

### Browsers

- [ ] safari
- [x] chrome
- [x] firefox

### Device

- [x] desktop – wide screen


jjaffeux (approved):
Cant try locally right now but lgtm, thanks ❤️

PR #1159, opened 2022-08-09
FIX: Don't try to scroll to the last read message when fetching from latest.

romanrizzi:
Either clicking the jump arrow or sending a message when there are unloaded ones will fetch the latest page and scroll to the latest one. Calling `_markLastReadMessage()` in these scenarios will break the scrolling since it also tries to scroll to the last read message immediately after, which we can't guarantee it's on the loaded page.

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -1038,7 +1038,7 @@ export default Component.extend({
       });
 
     if (this.details.can_load_more_future) {
-      msgCreationPromise.then(this._fetchAndScrollToLatest);
+      msgCreationPromise.then(() => this._fetchAndScrollToLatest());
     } else {
       const stagedMessage = this._prepareSingleMessage(
         // We need to add the user and created at for presentation of staged message
@@ -1350,10 +1350,7 @@ export default Component.extend({
   restickScrolling(event) {
     event.preventDefault();
 
-    return this._fetchAndScrollToLatest().then(() => {
-      this.set("stickyScroll", true);
-      this._stickScrollToBottom();
-    });
+    return this._fetchAndScrollToLatest();
   },
 
   focusComposer() {
@@ -1491,7 +1488,8 @@ export default Component.extend({
     return this.fetchMessages(this.chatChannel, {
       fetchFromLastMessage: true,
     }).then(() => {
-      this.scrollToMessage(this.messages[this.messages.length - 1].id);
+      this.set("stickyScroll", true);
+      this._stickScrollToBottom();
     });
   },
 });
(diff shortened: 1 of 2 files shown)


jjaffeux:
I wonder if we can write a component test for this, building lots of messages with fabricator and trying to set a state where we the bottom button would show

jjaffeux:
If you cant make it for now roman, just add a skipped test explaining what we would want to test here. I will do this when tests are in better shape.

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -189,9 +189,13 @@ export default Component.extend({
   },
 
   @bind
-  onScrollHandler() {
+  onScrollHandler(e) {
     cancel(this.stickyScrollTimer);
-    this.stickyScrollTimer = discourseDebounce(this, this.onScroll, 100);
+    this.stickyScrollTimer = discourseDebounce(
+      this,
+      () => this.onScroll(e),
```
jjaffeux, line 196: Does it work? I don't think using an anonymous function and debounce works as it always create a new context. Try using a high value instead of 100 and see if it's debounced.

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -617,7 +622,7 @@ export default Component.extend({
     }
   },
 
-  onScroll() {
+  onScroll(e) {
```
jjaffeux, line 625: That's a nitpick, but I prefer event much more than e, makes it way easier to grep for it

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1493,4 +1490,13 @@ export default Component.extend({
       });
     });
   },
+
+  _fetchAndScrollToLatest() {
+    return this.fetchMessages(this.chatChannel, {
+      fetchFromLastMessage: true,
+    }).then(() => {
+      this.set("stickyScroll", true);
```
jjaffeux, line 1498: this can be destroyed at this point given fetchMessages is async

assets/javascripts/discourse/components/chat-live-pane.js
romanrizzi: Yes, it works! Unless I missed something, the fat arrow should bind `this` to the context given to debounce.

romanrizzi: Tried with a high value, and `discourseDebounce` works as expected this way 👍

jjaffeux: OK I see good :+1: thanks

jjaffeux (approved):

jjaffeux:
Just tried it locally, looks good :+1: thanks for the test too.

PR #1167, opened 2022-08-12
UX: makes oneboxes full width to prevent different width

jjaffeux:
Before:

Screenshot 2022-08-12 at 13 17 55

After:

Screenshot 2022-08-12 at 13 18 12


OsamaSayegh (approved):

CvX:
Yiss 🤩

PR #1170, opened 2022-08-12
FIX: prevents esc shortcut to remove content when in full page

jjaffeux:
Prior to this fix navigating to a channel in full page and pressing <kbd>escape</kbd> would erase content of the live pane.

### DIFF
--- a/assets/javascripts/discourse/initializers/chat-keyboard-shortcuts.js
+++ b/assets/javascripts/discourse/initializers/chat-keyboard-shortcuts.js
@@ -48,9 +48,9 @@ export default {
       return true;
     };
     const isDrawerExpanded = () => {
-      return (
-        document.querySelector(".topic-chat-container.expanded")?.length === 1
-      );
+      return document.querySelector(".topic-chat-float-container:not(.hidden)")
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/initializers/chat-keyboard-shortcuts.js
```diff
@@ -47,6 +47,8 @@ export default {
       }
       return true;
     };
+    const isDrawerExpanded = () =>
+      document.querySelector(".topic-chat-container.expanded").length === 1;
```
keegangeorge, line 51: Can we add the brackets around this so it looks more clear?

Also, I'm getting an error here:
```
plugins/discourse-chat/assets/javascripts/javascripts/discourse/initializers/chat-keyboard-shortcuts:60 Uncaught TypeError: Cannot read properties of null (reading 'length')
```

keegangeorge (changes requested):
The `esc` to close chat in drawer mode is not working for me... Not too sure why

keegangeorge (approved):
LGTM!

PR #1174, opened 2022-08-13
UX: show status on the direct message chats list using the new component with rich tooltip

AndrewPrigorshnev:
Before:

Screenshot 2022-08-14 at 01 16 06

After:

Screenshot 2022-08-14 at 01 16 57

### DIFF
--- a/assets/javascripts/discourse/templates/components/chat-channel-row.hbs
+++ b/assets/javascripts/discourse/templates/components/chat-channel-row.hbs
@@ -11,10 +11,7 @@
   {{chat-channel-title channel=channel unreadIndicator=true}}
 
   {{#if this.showUserStatus}}
-    {{emoji
-      channel.chatable.users.firstObject.status.emoji
-      title=channel.chatable.users.firstObject.status.description
-    }}
+    <UserStatusMessage @status={{channel.chatable.users.firstObject.status}} />
   {{/if}}
 
   {{#if (and options.leaveButton channel.isFollowing (not site.mobileView))}}
--- a/test/javascripts/components/chat-channel-row-test.js
+++ b/test/javascripts/components/chat-channel-row-test.js
@@ -124,7 +124,7 @@ module("Discourse Chat | Component | chat-channel-row", function (hooks) {
     },
 
     async test(assert) {
-      assert.ok(exists(".emoji[title='Off to dentist']"));
+      assert.ok(exists(".user-status-message"));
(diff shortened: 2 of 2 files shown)


ZogStriP (approved):

assets/javascripts/discourse/templates/components/chat-channel-row.hbs
```diff
@@ -11,10 +11,7 @@
   {{chat-channel-title channel=channel unreadIndicator=true}}
 
   {{#if this.showUserStatus}}
-    {{emoji
-      channel.chatable.users.firstObject.status.emoji
-      title=channel.chatable.users.firstObject.status.description
-    }}
+    <UserStatusMessage @status={{channel.chatable.users.firstObject.status}} />
```
jjaffeux, line 14: I wonder if we can't encapsulate the "showUserStatus" logic inside this directly?

jjaffeux (approved):

assets/javascripts/discourse/templates/components/chat-channel-row.hbs
AndrewPrigorshnev: > I wonder if we can't encapsulate the "showUserStatus" logic inside this directly?

Do you mean to encapsulate the logic inside the `UserStatusMessage` component? If yes, that's impossible, because that logic is specific to the chat:

https://github.com/discourse/discourse-chat/blob/f82d6ce2a64522847e67259528da0975b279f09e/assets/javascripts/discourse/components/chat-channel-row.js#L65-L76

while we use the component in many places, not only in chat.

jjaffeux: ok I see

AndrewPrigorshnev: > It does the raise the question of what do we do if we want a custom container only when we show this...

I'm thinking about this and feeling like maybe it's better to completely avoid using such wrapping divs? 🤔 I believe I used such wrapping with user status in a couple of places, and probably those places should be fixed.

I was adding those divs in order to make possible styling I needed. Instead, I probably had to add a div-container inside the `UserStatusMessage` component. That would enable me to use styles I wanted, at the same time I would just use `<UserStatusMessage />` without a wrapper in templates.

But I'm assuming that styling is the only reason to have such a wrapping div, or are there other reasons?

PR #1176, opened 2022-08-16
DEV: Fix flaky validator spec

CvX:
Don't hard-code ids in spec. Those will eventually fail.

Also improved the whole spec and `valid_value?` method.


CvX:
cc: @martin-brennan, @nattsw

gschlager (approved):

martin-brennan (approved):

PR #1179, opened 2022-08-18
FEATURE: Allow changing DM channel notification settings

martin-brennan:
This commit allows users to change the Mute settings
for DM channels as they can for regular category channels,
as well as changing the desktop and mobile notification
levels for the channel if it is not muted. It also changes
the unread_count logic for DM channels -- since the channels
can now be muted, the unread count does not always show regardless
for DM channels.

### DIFF
--- a/app/jobs/regular/chat_notify_watching.rb
+++ b/app/jobs/regular/chat_notify_watching.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 module Jobs
-  # TODO (martin) A spec for this whole class.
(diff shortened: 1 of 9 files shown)


martin-brennan:
Note to reviewers: View this with whitespace turned off, the channel settings view is actually a 1 line change.

ZogStriP (approved):

jjaffeux (approved):

martin-brennan:
> My only note is that we shouldn't have all these MessageBus, PostAlerter mentions in the acceptance tests.

Ah yep fair enough, removed the mentions.

PR #1182, opened 2022-08-18
FIX: Display error when switching channels returns a 429

romanrizzi:


test/javascripts/acceptance/chat-live-pane-test.js
```diff
@@ -170,3 +170,63 @@ acceptance(
...
+        return helper.response({ success: "OK" });
+      });
+    });
+
+    test("Handles 429 errors by displaying an alert", async function (assert) {
+      await visit("/chat/channel/1/cat");
+
+      assert.ok(exists(`.bootbox`), "We displayed a 429 error");
+
+      await click(".bootbox span.d-button-label");
```
pmusaraj, line 229: Do we need this last click action? We're not asserting anything after it...

pmusaraj (approved):

test/javascripts/acceptance/chat-live-pane-test.js
```diff
@@ -170,3 +170,63 @@ acceptance(
     });
   }
 );
+
+acceptance(
+  "Discourse Chat - Chat live pane - handling 429 errors",
+  function (needs) {
+    needs.user({
+      username: "eviltrout",
```
jjaffeux, line 178: probably don't need username and id here

test/javascripts/acceptance/chat-live-pane-test.js
```diff
@@ -170,3 +170,63 @@ acceptance(
...
+      can_chat: true,
+      has_chat_enabled: true,
+    });
+    needs.settings({
+      chat_enabled: true,
+    });
+
+    needs.pretender((server, helper) => {
+      server.get("/chat/:chatChannelId/messages.json", () => {
+        return helper.response(429, {
```
jjaffeux, line 189: do we need that much info in the test? we only check 429 afaict

test/javascripts/acceptance/chat-live-pane-test.js
romanrizzi: We should clean things up after the test finishes. The modal doesn't close itself.

PR #1185, opened 2022-08-20
FEATURE: show user status on the direct message channels on the sidebar

AndrewPrigorshnev:
- Show user status on sidebar
- Make status "live"
- Acceptance tests

### View mode

### Browsers

- [ ] safari
- [ ] chrome
- [ ] firefox

### Device

- [ ] desktop – wide screen


assets/javascripts/discourse/initializers/chat-sidebar.js
```diff
@@ -251,7 +263,11 @@ export default {
             get text() {
               const username = this.channel.title.replaceAll("@", "");
               if (this.oneOnOneMessage) {
-                return htmlSafe(`${username} ${decorateUsername(username)}`);
+                const status = this.channel.chatable.users[0].get("status");
```
jjaffeux, line 266: probably doesn't need get here

jjaffeux (approved):

ZogStriP (approved):

assets/javascripts/discourse/initializers/chat-sidebar.js
AndrewPrigorshnev: This is on purpose in this case. We show this status on a glimmer component in sidebar and to make it live we need either make `user.status` property tracked or call `get` here. We went with `get` because we don't want to add tracked properties to models for now (see [this discussion](https://github.com/discourse/discourse/pull/18010#issuecomment-1222039811)).

jjaffeux: damn... this is certainly unexpected, ok 🤷‍♂️

PR #1194, opened 2022-08-22
FIX: relies on core composer-height which should now be correct

jjaffeux:
composer-height used to be unreliable, hence all the custom code in discourse-chat. This is now a thing of the past and we can simply rely on this.


assets/stylesheets/common/common.scss
```diff
@@ -893,15 +896,6 @@ html.has-full-page-chat {
     }
   }
 
-  // make space for the composer so it doesn't overlay chat content
```
jjaffeux, line 896: this was actually dead code

ZogStriP (approved):

PR #1195, opened 2022-08-22
FEATURE: hides public channels section if unusable

jjaffeux:
Unusable is defined by the following conditions:
- user is not staff
- user doesn’t have any possible public channel

Note this commit is also implementing preload of current user chat channels, preventing a /chat/channels ajax call on each first page load. This was on the roadmap and was needed in this commit to allow for clean initial state.

### DIFF
--- a/assets/javascripts/discourse/components/channels-list.js
+++ b/assets/javascripts/discourse/components/channels-list.js
@@ -42,11 +42,14 @@ export default class ChannelsList extends Component {
     "currentUser.{staff,has_joinable_public_channels}"
   )
   get displayPublicChannels() {
-    return (
-      !this.publicChannelsEmpty ||
-      this.currentUser?.staff ||
-      this.currentUser?.has_joinable_public_channels
-    );
+    if (this.publicChannelsEmpty) {
+      return (
+        this.currentUser?.staff ||
+        this.currentUser?.has_joinable_public_channels
+      );
+    }
+
+    return true;
   }
 
   @computed("inSidebar")
--- a/assets/javascripts/discourse/components/channels-list.js
+++ b/assets/javascripts/discourse/components/channels-list.js
@@ -37,6 +37,18 @@ export default class ChannelsList extends Component {
     }`;
   }
 
+  @computed(
+    "publicChannelsEmpty",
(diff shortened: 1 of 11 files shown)


martin-brennan (approved):
Looks great :+1:

ZogStriP (approved):
LGTM 👍

assets/javascripts/discourse/components/channels-list.js
```diff
@@ -37,6 +37,21 @@ export default class ChannelsList extends Component {
...
 
+  @computed(
+    "publicChannelsEmpty",
+    "currentUser.{staff,has_joinable_public_channels}"
+  )
+  get displayPublicChannels() {
+    if (this.publicChannelsEmpty) {
+      return (
+        this.currentUser?.staff ||
+        this.currentUser?.has_joinable_public_channels
```
ZogStriP, line 48: I wonder if we should make the server return true for "has joinable public channels" for staff members so we don't have to check both on the client?

jjaffeux: Yes I hesitated on this, my current thinking is that we want to keep `has_joinable_public_channels` "correct". You might be a staff and have followed all the existing channels and we want to return false in this case otherwise this would be quite unexpected.

assets/javascripts/discourse/services/chat.js
```diff
@@ -294,39 +301,43 @@ export default class Chat extends Service {
...
+  _processChannels(channels) {
+    this.setProperties({
+      publicChannels: A(
+        this.sortPublicChannels(
+          (channels.public_channels || []).map((channel) =>
+            this.processChannel(channel)
+          )
+        )
+      ),
+      // We don't need to sort direct message channels, as the channel list
```
ZogStriP, line 321: Is this comment relevant? Looks like we're sorting them by calling "sortDirectMessageChannels".

ZogStriP (approved):

assets/javascripts/discourse/services/chat.js
jjaffeux: Yes never quite understood this comment, I should watch this more closely some day

PR #1197, opened 2022-08-24
FIX: deletes trashed messages when hitting retention limit

jjaffeux:
Prior to this fix, we would keep trashed messages in the DB forever.


martin-brennan (approved):
LGTM, let's burn the trash :fire:

PR #1206, opened 2022-08-29
DEV: rewrites flakey/skipped acceptance tests into components tests

jjaffeux:


test/javascripts/acceptance/chat-test.js
```diff
@@ -146,63 +146,6 @@ acceptance("Discourse Chat - without unread", function (needs) {
     assert.ok(query(".topic-chat-container").classList.contains("channel-9"));
   });
 
-  test("Clicking mention notification inside other full page channel switches the channel", async function (assert) {
```
romanrizzi, line 149: I don't think we can test this in a component test 🤔. Are we rewriting this one?

jjaffeux: At some point I just trust a link is going to work, so I don't test the interaction, but I know i generated a valid link and there's not reason for this to not work, so my confidence is quite high here.

This test was flakey and given the confidence I have as explained above, I think our time is better used on something else than trying to understand the flakeyness (although I dod have few ideas of why it was flakey).

romanrizzi (approved):

PR #1207, opened 2022-08-29
REFACTOR: Trigger read updates through the visibility observer.

romanrizzi:
This change lets the visibility observer trigger a read update directly as soon as the message is visibile, simplifying the chat-live-pane, and making
the behavior easier to test.

The chat service handles the debouncing of the update action, and the chat-channel model, the ajax request and tracking of the last read message sent.

As a side note, I don't think removing the document#querySelectorAll call it's possible because we need to recalculate the current last visible message
when the function fires since a scroll could have already happened.


romanrizzi:
@davidtaylorhq - This follows up on your suggestion in #1122. I think you'll be interested in taking a look.

assets/javascripts/discourse/services/chat-message-visibility-observer.js
```diff
@@ -13,6 +15,11 @@ export default class ChatMessageVisibilityObserver extends Service {
   _observerCallback(entries) {
     entries.forEach((entry) => {
       entry.target.dataset.visible = entry.isIntersecting;
+
+      if (entry.isIntersecting && !isTesting()) {
+        const chatService = getOwner(this).lookup("service:chat");
```
davidtaylorhq, line 20: I wonder if it would be cleaner to inject this into the ChatMessageVisibilityObserver?

```javascript
import { inject as service } from "@ember/service";

export default class ChatMessageVisibilityObserver extends Service {
  @service("chat") chatService;
  ...
```

assets/javascripts/discourse/services/chat.js
```diff
@@ -878,6 +881,45 @@ export default class Chat extends Service {
     );
   }
 
+  updateLastReadMessage() {
+    discourseDebounce(this, this._queuedReadMessageUpdate, READ_INTERVAL);
+  }
+
+  _queuedReadMessageUpdate() {
+    const visibleMessages = document.querySelectorAll(
```
davidtaylorhq, line 889: > As a side note, I don't think removing the document#querySelectorAll call it's possible because we need to recalculate the current last visible message when the function fires since a scroll could have already happened.

I was thinking that the visibility-observer service would take care of everything, and we wouldn't need the `data-visible` attribute at all. Sketching it out a little:

```javascript
chatService.updateLastReadMessage(channel, messageId)
```
and then the service would do something like

```javascript
updateLastReadMessage(channel, messageId) {
  if(this.queuedUpdate && (this.queuedUpdate.channel === channel) && (this.queuedUpdate?.messageId > messageId)){
    return; // Already queued an update to a more recent message
  }
  this.queuedUpdate = { channel, messageId };
  discourseDebounce(this, this._queuedReadMessageUpdate, READ_INTERVAL);
}

_queuedReadMessageUpdate(){
  const [channel, latestUnreadMsgId] = this.queuedUpdate;
  const hasUnreadMessages = latestUnreadMsgId > channel.lastSendReadMessageId;
  // ... (existing implemementation)
}
```

Would something like that work? Storing the most-recent state in `queuedUpdate` means that it will almost have the most up-to-date data, even after the debounce delay.

romanrizzi: My first approach was somewhere along these lines, but unfortunately, it won't work. There are various scenarios where we force a scroll after fetching messages, which makes the highest message ID briefly visible to the observer, but it's not what the user sees. If the highest message ID gets passed to the service, all other calls won't update the ID.

assets/javascripts/discourse/services/chat-message-visibility-observer.js
romanrizzi: This didn't work for me, but now I realize I also need to decorate the callback function with a `@bind`.

assets/javascripts/discourse/services/chat.js
davidtaylorhq: > which makes the highest message ID briefly visible to the observer, but it's not what the user sees

That feels like a bug, right? Surely we don't want the visibility observer to fire for messages which are not truly... visible? But also I realise it might be very tricky to unpick/fix that, so we might have to just go ahead with this as-is for now.

romanrizzi: Yes, but it wasn't evident until this change. The observer starts firing as soon as a message appears on the pane, so one solution that comes to mind is to tell the observer to wait until the `chat-live-pane` finishes preparing them, scrolling, etc. We could track this with a boolean in the service and expose methods to start/pause.

romanrizzi: Unfortunately, this is much more work than I originally anticipated. Due to how we handle scrolling using the ember's run loop, sometimes scheduling and afterRender callback, and sometimes using next + afterRender, it's not obvious how we should fix this.

If we're OK with it, I'd like to move forward with this as it is now. Hopefully, we'll be able to simplify the scrolling logic in the `chat-live-pane` component at some point and get rid of it.

PR #1209, opened 2022-08-30
FEATURE: Use direct_message_enabled_groups for DM access

martin-brennan:
This commit introduces the direct_message_enabled_groups setting,
which is used to control which groups can send direct messages in
chat. Staff can always send direct messages.

If a user is not part of any direct_message_enabled_groups,
then they cannot create new DM channels, cannot send messages
in existing DM channels, and if they have no existing DM channels
they will not see the DM channel list in the sidebar.


assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -129,6 +131,28 @@ export default {
...
+            this.isInAnyGroups(
+              this.siteSettings.direct_message_enabled_groups
+                .split("|")
+                .map((groupId) => parseInt(groupId, 10))
+            )
+          );
+        },
+      });
+
+      // TODO (martin) Better way to do this??
```
martin-brennan, line 151: @jjaffeux I wanted to ask for your advice on this, unless I either did this or, mysteriously, called `User.create()` after modifying the class, I was unable to call `allowDirectMessages` on the current user. Am I doing something wrong or is this fine to do?

assets/javascripts/discourse/services/chat.js
```diff
@@ -60,6 +60,22 @@ export default class Chat extends Service {
...
+  @computed("currentUser.staff", "currentUser.groups.[]")
+  get userCanDirectMessage() {
+    if (!this.currentUser) {
+      return false;
+    }
+
+    return (
+      this.currentUser.staff ||
+      this.currentUser.isInAnyGroups(
+        (this.siteSettings.direct_message_enabled_groups || "11") // trust level 1 auto group
```
martin-brennan, line 72: This is mainly a workaround for the limitations of JS tests loading settings, this setting should never be null IRL. This PR in core is working on the solution but I don't want to hold up this chat feature for it https://github.com/discourse/discourse/pull/18413

config/locales/server.en.yml
```diff
@@ -15,9 +15,11 @@ en:
     chat_allow_uploads: "Allow uploads in public chat channels and direct message channels."
     chat_archive_destination_topic_status: "The status that the destination topic should be once a channel archive is completed. This only applies when the destination topic is a new topic, not an existing one."
     default_emoji_reactions: "Default emoji reactions for chat messages. Add up to 5 emojis for quick reaction."
+    direct_message_enabled_groups: "Allow users within these groups to create direct messages and reply to direct messages. Trust level groups include all trust levels above that number, for example choosing trust_level_1 also allows trust_level_2, 3, 4 users to send DMs. Note that staff can always send direct messages no matter what."
```
tgxworld, line 18:
```suggestion
    direct_message_enabled_groups: "Allow users within these groups to create direct messages and reply to direct messages. Trust level groups include all trust levels above that number, for example choosing trust_level_1 also allows trust_level_2, 3, 4 users to send direct messages. Note that staff can always send direct messages no matter what."
```

config/locales/server.en.yml
```diff
@@ -15,9 +15,11 @@ en:
...
     chat_message_flag_allowed_groups: "Users in these groups are allowed to flag chat messages."
     errors:
       chat_default_channel: "The default chat channel must be a public channel."
+      direct_message_enabled_groups_invalid: "You must specify at least one group for this setting. If you do not want anyone except staff to send DMs, choose the staff group."
```
tgxworld, line 22:
```suggestion
      direct_message_enabled_groups_invalid: "You must specify at least one group for this setting. If you do not want anyone except staff to send direct messages, choose the staff group."
```

For consistency with the other setting's description.

spec/lib/direct_message_channel_creator_spec.rb
```diff
@@ -106,6 +108,33 @@
...
+          existing_channel = subject.create!(acting_user: user_1, target_users: [user_1, user_1])
+        }.to raise_error(Discourse::InvalidAccess)
+        expect(ChatChannel.count).to eq(channel_count)
+        expect(UserChatChannelMembership.count).to eq(membership_count)
+      end
+
+      context "when user is staff" do
+        before { user_1.update!(admin: true) }
+
+        it "doesn't create an error and doesn't create a new chat channel" do
```
tgxworld, line 128: I'm not sure I understand why an admin is not allowed to create a new chat channel here? Since the setting is configured to TL4, it means anyone TL4 and above can create a direct message right?

spec/lib/guardian_extensions_spec.rb
```diff
@@ -267,4 +267,63 @@
...
+
+      it "returns false if the channel is archived" do
+        channel.update!(status: :archived)
+        expect(staff_guardian.can_create_channel_message?(channel)).to eq(false)
+      end
+
+      context "for direct message channels" do
+        it "returns true if enable_personal_messages is false since staff can always send DMs" do
+          channel.update!(status: :open)
+          SiteSetting.enable_personal_messages = false
```
tgxworld, line 291: Is this site setting useful in the context of this test? I looked at the method which we're testing here and the site setting doesn't seem to be used in the method at all.

lib/guardian_extensions.rb
```diff
@@ -46,8 +50,12 @@ def can_move_chat_messages?(channel)
   end
 
   def can_create_channel_message?(chat_channel)
-    return chat_channel.open? || chat_channel.closed? if is_staff?
-    chat_channel.open?
+    valid_statuses = is_staff? ? %w[open closed] : ["open"]
```
tgxworld, line 53: Since staff consists of both admins and moderators, I just want to confirm that is what we want here where moderators can create channels and direct message channels as well.

spec/lib/guardian_extensions_spec.rb
```diff
@@ -267,4 +267,63 @@
...
+
+      it "returns false if the channel is archived" do
+        channel.update!(status: :archived)
+        expect(guardian.can_create_channel_message?(channel)).to eq(false)
+      end
+
+      context "for direct message channels" do
+        it "returns false if enable_personal_messages is false" do
+          channel.update!(status: :open)
+          SiteSetting.enable_personal_messages = false
```
tgxworld, line 316: Same concern as https://github.com/discourse/discourse-chat/pull/1209/files#r986525687

spec/lib/guardian_extensions_spec.rb
```diff
@@ -267,4 +267,63 @@
...
+          expect(guardian.can_create_channel_message?(dm_channel)).to eq(false)
+        end
+
+        it "returns false if the user is not the correct trust level for sending personal messages" do
+          user.update!(trust_level: 1)
+          channel.update!(status: :open)
+          SiteSetting.min_trust_to_send_messages = 4
```
tgxworld, line 323: Should we be using the new site setting here?

martin-brennan:
Good points @tgxworld thanks, I opened this ages ago so I think my eyes glazed over a little. Will reexamine today and address comments :+1:

spec/lib/direct_message_channel_creator_spec.rb
martin-brennan: This is returning the existing channel based on the context above, will slightly change this description so it is clearer.

lib/guardian_extensions.rb
martin-brennan: Yes that is correct

tgxworld (approved):
Looks good :+1:

PR #1211, opened 2022-08-30
FIX: Prevent deleting a category that has channels

Flink:
Currently nothing prevents from deleting a category which has chat enabled on it. This already led to a bug (see commit 07f9cb4).

This PR addresses this issue by not allowing the deletion of a category if a channel is linked to it. This is the same approach we already have for not deleting a category when topics or sub-categories are present.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -170,3 +170,7 @@ en:
       select_title: "Set chat summary emails frequency to:"
       never: Never
       when_away: Only when away
+
+  category:
+    cannot_delete:
+      has_chat_channels: "Can't delete this category because it has chat channels."
--- a/lib/extensions/category_extension.rb
+++ b/lib/extensions/category_extension.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module DiscourseChat::CategoryExtension
+  extend ActiveSupport::Concern
+
+  prepended do
+    has_one :chat_channel, as: :chatable
+  end
+
+  def cannot_delete_reason
+    return I18n.t('category.cannot_delete.has_chat_channels') if chat_channel
+    super
+  end
+end
--- a/plugin.rb
+++ b/plugin.rb
@@ -164,6 +164,7 @@ after_initialize do
   load File.expand_path("../lib/extensions/user_option_extension.rb", __FILE__)
   load File.expand_path("../lib/extensions/user_notifications_extension.rb", __FILE__)
   load File.expand_path("../lib/extensions/user_email_extension.rb", __FILE__)
+  load File.expand_path("../lib/extensions/category_extension.rb", __FILE__)
   load File.expand_path("../lib/slack_compatibility.rb", __FILE__)
   load File.expand_path("../lib/post_notification_handler.rb", __FILE__)
   load File.expand_path("../lib/secure_media_compatibility.rb", __FILE__)
@@ -216,16 +217,16 @@ after_initialize do
       limited_pretty_text_markdown_rules: ChatMessage::MARKDOWN_IT_RULES,
     }
 
-    Guardian.class_eval { include DiscourseChat::GuardianExtensions }
-    UserNotifications.class_eval { prepend DiscourseChat::UserNotificationsExtension }
-    UserOption.class_eval { prepend DiscourseChat::UserOptionExtension }
-    Category.class_eval { has_one :chat_channel, as: :chatable }
+    Guardian.prepend DiscourseChat::GuardianExtensions
+    UserNotifications.prepend DiscourseChat::UserNotificationsExtension
+    UserOption.prepend DiscourseChat::UserOptionExtension
+    Category.prepend DiscourseChat::CategoryExtension
     User.class_eval do
       has_many :user_chat_channel_memberships, dependent: :destroy
       has_many :chat_message_reactions, dependent: :destroy
       has_many :chat_mentions
     end
-    Jobs::UserEmail.class_eval { prepend DiscourseChat::UserEmailExtension }
+    Jobs::UserEmail.prepend DiscourseChat::UserEmailExtension
 
     Bookmark.register_bookmarkable(ChatMessageBookmarkable)
   end
--- a/spec/lib/guardian_extensions_spec.rb
+++ b/spec/lib/guardian_extensions_spec.rb
@@ -2,7 +2,7 @@
 
 require "rails_helper"
 
-describe DiscourseChat::GuardianExtensions do
+RSpec.describe DiscourseChat::GuardianExtensions do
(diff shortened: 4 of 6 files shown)


plugin.rb
```diff
@@ -216,16 +217,16 @@ def self.onebox_template
...
 
-    Guardian.class_eval { include DiscourseChat::GuardianExtensions }
-    UserNotifications.class_eval { prepend DiscourseChat::UserNotificationsExtension }
-    UserOption.class_eval { prepend DiscourseChat::UserOptionExtension }
-    Category.class_eval { has_one :chat_channel, as: :chatable }
+    Guardian.prepend DiscourseChat::GuardianExtensions
+    UserNotifications.prepend DiscourseChat::UserNotificationsExtension
+    UserOption.prepend DiscourseChat::UserOptionExtension
+    Category.prepend DiscourseChat::CategoryExtension
     User.class_eval do
```
Flink, line 224: I guess we could probably normalize the use of `prepend` with a module as I think it would be better from a code organization point of view.

spec/lib/guardian_extensions_spec.rb
```diff
@@ -243,5 +243,28 @@
         end
       end
     end
+
+    describe "#can_delete_category?" do
+      alias_matcher :be_able_to_delete_category, :be_can_delete_category
```
Flink, line 248: Trying something here to make specs clearer to read.

config/locales/server.en.yml
```diff
@@ -170,3 +170,7 @@ en:
       select_title: "Set chat summary emails frequency to:"
       never: Never
       when_away: Only when away
+
+  category:
+    cannot_delete:
+      has_chat_channel: "Can't delete this category because it has a chat channel."
```
mcwumbly, line 176:
```suggestion
      has_chat_channels: "Can't delete this category because it has chat channels."
```

lib/extensions/category_extension.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module DiscourseChat::CategoryExtension
+  extend ActiveSupport::Concern
+
+  prepended do
+    has_one :chat_channel, as: :chatable
+  end
+
+  def cannot_delete_reason
+    return I18n.t('category.cannot_delete.has_chat_channel') if chat_channel
```
mcwumbly, line 11:
```suggestion
    return I18n.t('category.cannot_delete.has_chat_channels') if chat_channel
```

spec/models/category_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  describe "#cannot_delete_reason" do
+    subject(:reason) { category.cannot_delete_reason }
+
+    context "when a chat channel is present" do
+      let(:channel) { Fabricate(:chat_channel) }
+      let(:category) { channel.chatable }
+
+      it "returns a message" do
+        expect(reason).to match(/can't delete this category because it has a chat channel/i)
```
mcwumbly, line 16:
```suggestion
        expect(reason).to match(/can't delete this category because.*chat/i)
```

Flink:
@mcwumbly Thanks, just applied your suggestions :ok_hand:

spec/lib/guardian_extensions_spec.rb
jjaffeux: not sure I follow, can you expand please?

lib/extensions/category_extension.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module DiscourseChat::CategoryExtension
+  extend ActiveSupport::Concern
+
+  prepended do
+    has_one :chat_channel, as: :chatable
+  end
+
+  def cannot_delete_reason
+    return I18n.t('category.cannot_delete.has_chat_channels') if chat_channel
```
romanrizzi, line 11: If there is more than one reason, shouldn't we combine them?

spec/models/category_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  describe "#cannot_delete_reason" do
+    subject(:reason) { category.cannot_delete_reason }
+
+    context "when a chat channel is present" do
+      let(:channel) { Fabricate(:chat_channel) }
+      let(:category) { channel.chatable }
+
+      it "returns a message" do
+        expect(reason).to match(/can't delete this category because.*chat/i)
```
jjaffeux, line 16: should we match on the I18n instead?

jjaffeux (approved):
LGTM thanks

plugin.rb
Flink: I did it where we only had a call `prepend` or where I had to change something. I think we should probably do it for `User` for example (but I didn’t as it’s not necessarily related to this PR) :slightly_smiling_face:

spec/lib/guardian_extensions_spec.rb
Flink: Yes sorry. Defining this alias allows to write `expect(staff_guardian).to be_able_to_delete_category(category)` on line 259 which reads better than `expect(staff_guardian).to be_can_delete_category(category)` or `expect(staff_guardian.can_delete_category?(category)).to eq(true)` IMO.

lib/extensions/category_extension.rb
Flink: The original method is this one:
```ruby
  def cannot_delete_reason
    return I18n.t('category.cannot_delete.uncategorized') if self.uncategorized?
    return I18n.t('category.cannot_delete.has_subcategories') if self.has_children?

    if self.topic_count != 0
      oldest_topic = self.topics.where.not(id: self.topic_id).order('created_at ASC').limit(1).first
      if oldest_topic
        I18n.t('category.cannot_delete.topic_exists', count: self.topic_count, topic_link: "<a href=\"#{oldest_topic.url}\">#{CGI.escapeHTML(oldest_topic.title)}</a>")
      else
        # This is a weird case, probably indicating a bug.
        I18n.t('category.cannot_delete.topic_exists_no_oldest', count: self.topic_count)
      end
    end
  end
```
So I guess we could combine messages but the original implementation doesn’t seem to care about this (a category can have subcategories _and_ topics). Not sure it’s worth the time?

romanrizzi: I don't think so, no. Thanks for clarifying!

plugin.rb
jjaffeux: Ok cool, free to do it in a follow up PR thanks

spec/models/category_spec.rb
jjaffeux: the first one

jjaffeux: mostly I dont think changing the translation should break this test

jjaffeux:
Feel free to merge tomorrow when you feel ready :+1:

PR #1212, opened 2022-08-30
FEATURE: adds support for unreliable network

jjaffeux:
- Allows to resend staged messages
- Displays an icon when network is unreliable
- Doesn't make the sed button unusable

### DIFF
--- a/assets/javascripts/discourse/components/chat-live-pane.js
+++ b/assets/javascripts/discourse/components/chat-live-pane.js
@@ -743,8 +743,10 @@ export default Component.extend({
       const stagedMessage = this.messageLookup[`staged-${data.stagedId}`];
       if (stagedMessage) {
         stagedMessage.setProperties({
+          error: null,
           staged: false,
           id: data.chat_message.id,
+          staged_id: null,
           excerpt: data.chat_message.excerpt,
         });
 
@@ -1032,12 +1034,13 @@ export default Component.extend({
     // Start ajax request but don't return here, we want to stage the message instantly when all messages are loaded.
     // Otherwise, we'll fetch latest and scroll to the one we just created.
     // Return a resolved promise below.
-    const msgCreationPromise = ajax(`/chat/${this.chatChannel.id}.json`, {
-      type: "POST",
-      data,
-    })
+    const msgCreationPromise = ChatApi.sendMessage(this.chatChannel.id, data)
       .catch((error) => {
         this._onSendError(data.staged_id, error);
+      })
+      .then((r) => {
+        console.log(r);
+        return r;
       })
       .finally(() => {
         if (this._selfDeleted) {
@@ -1105,9 +1108,37 @@ export default Component.extend({
   _onSendError(stagedId, error) {
     const stagedMessage = this.messageLookup[`staged-${stagedId}`];
     if (stagedMessage) {
-      stagedMessage.set("error", error.jqXHR.responseJSON.errors[0]);
-      this._resetAfterSend();
+      if (error.jqXHR?.responseJSON?.errors?.length) {
+        stagedMessage.set("error", error.jqXHR.responseJSON.errors[0]);
+      } else {
+        stagedMessage.set("error", "network_error");
+      }
     }
+
+    this._resetAfterSend();
+  },
+
+  @action
+  didResendStagedMessage(stagedMessage) {
+    this.set("sendingLoading", true);
+
+    const data = {
+      cooked: stagedMessage.cooked,
+      message: stagedMessage.message,
+      upload_ids: stagedMessage.upload_ids,
+      staged_id: stagedMessage.stagedId,
+    };
+
+    ChatApi.sendMessage(this.chatChannel.id, data)
+      .catch((error) => {
+        this._onSendError(data.staged_id, error);
+      })
+      .finally(() => {
+        if (this._selfDeleted) {
+          return;
+        }
+        this.set("sendingLoading", false);
+      });
   },
 
   @action
@@ -1328,6 +1359,10 @@ export default Component.extend({
 
   @action
   onHoverMessage(message, options = {}) {
+    if (message?.staged) {
+      return;
+    }
+
     discourseDebounce(
       this,
       this.debouncedOnHoverMessage,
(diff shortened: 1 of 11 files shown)


jjaffeux:
Screenshot 2022-08-30 at 15 42 09
Screenshot 2022-08-30 at 15 42 23

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1105,9 +1104,37 @@ export default Component.extend({
   _onSendError(stagedId, error) {
     const stagedMessage = this.messageLookup[`staged-${stagedId}`];
     if (stagedMessage) {
-      stagedMessage.set("error", error.jqXHR.responseJSON.errors[0]);
-      this._resetAfterSend();
+      if (error.jqXHR?.responseJSON?.errors?.length) {
```
jjaffeux, line 1107: I wanted to use `readyState` but this has proven to be super hard to fake in tests, and this is what we do in core.

assets/javascripts/discourse/templates/components/chat-message.hbs
```diff
@@ -151,7 +151,15 @@
 
           {{#if message.error}}
             <div class="chat-send-error">
-              {{message.error}}
+              {{#if (eq message.error "network_error")}}
```
jjaffeux, line 154: Im mixed on this check, but the alternative is adding some kind of error type, which would be more state to handle, not sure it's worth the complexity

assets/javascripts/discourse/services/chat.js
```diff
@@ -52,6 +52,7 @@ export default class Chat extends Service {
   directMessagesLimit = 20;
   _chatOpen = false;
   _fetchingChannels = null;
+  isNetworkUnreliable = false;
```
jjaffeux, line 55: I might make it into its own service at some point and toggle it's value from more ajax calls

pmusaraj: Yeah, this seems like a service we would want to have in core, eventually.

test/javascripts/acceptance/chat-composer-test.js
```diff
@@ -0,0 +1,87 @@
...
+    chatChannelPretender(server, helper);
+    server.get("/chat/:id/messages.json", () =>
+      helper.response({ chat_messages: [], meta: {} })
+    );
+    server.post("/chat/drafts", () => helper.response(500, {}));
+    server.post("/chat/:id.json", () => {
+      sendAttempt += 1;
+      return sendAttempt === 1
+        ? helper.response(500, {})
+        : helper.response({ sucess: true });
```
pmusaraj, line 26:
```suggestion
        : helper.response({ success: true });
```

test/javascripts/acceptance/chat-composer-test.js
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  test("Draft with unreliable network", async function (assert) {
+    this.chatService.set("chatWindowFullPage", false);
+    await visit("/chat/channel/11/-");
+    await fillIn(".chat-composer-input", "network-error-message");
+
+    assert.ok(
+      exists(".chat-composer__unreliable-network"),
+      "it displays a network error icon"
```
pmusaraj, line 84: Does the draft save action take place automatically on `fillIn`? Because, otherwise, it seems like this is missing an action here.

assets/javascripts/discourse/templates/components/chat-message.hbs
```diff
@@ -151,7 +151,15 @@
...
+                <DButton
+                  class="btn-primary btn-small retry-staged-message-btn"
```
pmusaraj, line 156: Nitpicky, but this button as primary seems large. Maybe the situation calls for it, though. (Could also try with a `btn-flat`.)

assets/javascripts/discourse/templates/components/chat-live-pane.hbs
```diff
@@ -59,6 +59,7 @@
         isHovered=(eq message.id hoveredMessageId)
         onHoverMessage=(action "onHoverMessage")
         onSwitchChannel=onSwitchChannel
+        didResendStagedMessage=this.didResendStagedMessage
```
pmusaraj, line 62: Shouldn't this be called `resendStagedMessage` given that it's not a hook, as in, it is not adding a callback to the didResendStagedMessage event.

assets/javascripts/discourse/services/chat.js
```diff
@@ -851,7 +852,15 @@ export default class Chat extends Service {
...
     }
 
-    ajax("/chat/drafts", { type: "POST", data });
+    ajax("/chat/drafts", { type: "POST", data, ignoreUnsent: false })
+      .then(() => {
+        this.set("isNetworkUnreliable", false);
+      })
+      .catch((error) => {
+        if (!error.jqXHR?.responseJSON?.errors?.length) {
+          this.set("isNetworkUnreliable", true);
```
pmusaraj, line 861: Do we make a recurring polling call to `/chat/drafts`? I am wondering whether we need a way to periodically (every 30 seconds or minute) ensure that the network is still unreliable, otherwise we could be showing the icon to a lurker for a long time (i.e. until they next try to send).

pmusaraj (approved):
Looks generally good, just left a few mostly minor comments.

assets/javascripts/discourse/templates/components/chat-message.hbs
jjaffeux: My feeling is that it's not supposed to show often, so probably ok, but we can iterate on it of that becomes an issue, no objection

test/javascripts/acceptance/chat-composer-test.js
jjaffeux: yes it does, as soon as composer value changes

assets/javascripts/discourse/templates/components/chat-live-pane.hbs
jjaffeux: I changed this name many times 😅 Im ok with `resendStagedMessage` will make the change

assets/javascripts/discourse/services/chat.js
jjaffeux: Yes that's why I  noted early that I might add this network stuff in more ajax calls

jjaffeux: Honestlly I'm not a big fan of the 30 seconds network poll, but what we could do, is reseting network state after 30s

that way we ensure that if no action are taken, we don't stay in this state forever.

pmusaraj: Oh yeah, that's good! Reset after 30/60s makes sense, no need to stay in "alert mode" forever.

PR #1218, opened 2022-09-01
DEV: Clean up a bit status related code

Flink:


app/models/chat_channel.rb
```diff
@@ -12,7 +13,7 @@ class ChatChannel < ActiveRecord::Base
 
   has_one :chat_channel_archive
 
-  enum status: { open: 0, read_only: 1, closed: 2, archived: 3 }, _scopes: false
+  enum :status, { open: 0, read_only: 1, closed: 2, archived: 3 }, scopes: false
```
Flink, line 16: Using the new Rails 7 notation.

app/models/chat_channel.rb
```diff
@@ -21,6 +22,30 @@ class ChatChannel < ActiveRecord::Base
...
+    def public_channel_chatable_types
+      ["Category"]
+    end
+
+    def chatable_types
+      public_channel_chatable_types << "DirectMessageChannel"
+    end
+  end
+
+  statuses.keys.each do |status|
```
Flink, line 43: Used some simple meta-programming to define `open!` and the likes. This will also ensure those methods are up to date with the actual statuses.

app/models/chat_channel.rb
```diff
@@ -54,22 +79,6 @@ def remove(user)
     end
   end
 
-  def open?
```
Flink, line 57: Removed all those methods as they’re already defined by the `enum` keyword from AR.

app/models/chat_channel.rb
```diff
@@ -139,68 +148,29 @@ def title_from_chatable
     end
   end
 
+  private
+
   def change_status(acting_user, target_status)
```
Flink, line 153: Made this method private since it’s only used by `open!` and the likes.

app/models/chat_channel.rb
```diff
@@ -139,68 +148,29 @@ def title_from_chatable
...
-    return if !ChatChannel.statuses.include?(target_status.to_s)
```
Flink, line 143: This conditional is not necessary as we’re now controlling the `target_status` value.

app/models/chat_channel.rb
```diff
@@ -139,68 +148,29 @@ def title_from_chatable
...
     return if !Guardian.new(acting_user).can_change_channel_status?(self, target_status)
-    old_status = self.status
```
Flink, line 145: No need for `old_status`/`new_status` variables as AR provides those (`status` for the current one and `status_previously_was` for the one that was set before saving the model).

app/models/user_chat_channel_membership.rb
```diff
@@ -1,21 +1,19 @@
 # frozen_string_literal: true
 
 class UserChatChannelMembership < ActiveRecord::Base
+  NOTIFICATION_LEVELS = { never: 0, mention: 1, always: 2 }
+
   belongs_to :user
   belongs_to :chat_channel
   belongs_to :last_read_message, class_name: "ChatMessage", optional: true
 
-  DEFAULT_NOTIFICATION_LEVEL = :mention
```
Flink, line 8: Not used anywhere.

app/models/user_chat_channel_membership.rb
```diff
@@ -1,21 +1,19 @@
...
-  NOTIFICATION_LEVELS = { never: 0, mention: 1, always: 2 }
-  VALIDATED_ATTRS = %i[following muted desktop_notification_level mobile_notification_level]
```
Flink, line 10: Not used anywhere.

app/models/user_chat_channel_membership.rb
```diff
@@ -1,21 +1,19 @@
...
-  enum desktop_notification_level: NOTIFICATION_LEVELS, _prefix: :desktop_notifications
-  enum mobile_notification_level: NOTIFICATION_LEVELS, _prefix: :mobile_notifications
+  enum :desktop_notification_level, NOTIFICATION_LEVELS, prefix: :desktop_notifications
+  enum :mobile_notification_level, NOTIFICATION_LEVELS, prefix: :mobile_notifications
+  enum :join_mode, { manual: 0, automatic: 1 }
```
Flink, line 12: Grouped the `enum` together.

spec/models/chat_channel_spec.rb
```diff
@@ -78,7 +78,7 @@
 
       expect(events).to include(
         event_name: :chat_channel_status_change,
-        params: [{ channel: private_category_channel, old_status: :open, new_status: :closed }],
+        params: [{ channel: private_category_channel, old_status: "open", new_status: "closed" }],
```
Flink, line 81: I assumed having strings instead of symbols for event params won’t matter much. If that’s not the case, just tell me :slightly_smiling_face:

jjaffeux (approved):
LGTM

PR #1219, opened 2022-09-01
UX: implements a chat skeleton loader

jjaffeux:


assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -95,6 +95,9 @@ export default Component.extend({
       passive: true,
     });
     window.addEventListener("resize", this.onResizeHandler);
+    window.addEventListener("mousewheel", this.onScrollHandler, {
```
jjaffeux, line 98: This is due to ensure we are never stuck at the top without firing any scroll event, due to throttling and network it is possible to end up in this state.

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -28,7 +28,7 @@ import {
 } from "discourse/lib/user-presence";
 
 const MAX_RECENT_MSGS = 100;
-const STICKY_SCROLL_LENIENCE = 4;
+const STICKY_SCROLL_LENIENCE = 50;
```
jjaffeux, line 31: Increased to ensure we see the loading skeleton, especially on iOS

romanrizzi (approved):

ZogStriP (approved):

PR #1223, opened 2022-09-01
FIX: moves chat-msg-actions-mobile into live-pane

jjaffeux:
This move allows to apply translated3D on parent chat-messages-container (needed to avoid a case of lag rendering on iOS) without suffering from the bug of translated3D + fixed positioning (which is basically creating a different positioning context for fixed)

### DIFF
--- a/assets/javascripts/discourse/templates/components/chat-live-pane.hbs
+++ b/assets/javascripts/discourse/templates/components/chat-live-pane.hbs
@@ -37,6 +37,8 @@
 
 {{chat-retention-reminder chatChannel=chatChannel}}
 
+<div class="chat-message-actions-mobile-anchor"></div>
+
 <div class="chat-messages-scroll chat-messages-container">
   <div class="chat-messages-container">
     {{#if (or this.loading this.loadingMorePast)}}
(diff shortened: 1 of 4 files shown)


assets/stylesheets/mobile/chat-message.scss
```diff
@@ -12,6 +12,7 @@
   @include unselectable;
 }
 
+.chat-messages-scroll .chat-messages-container,
```
jjaffeux, line 15: we sadly have two "chat-messages-container" element in code atm so I'm forced to declare the hierarchy

assets/javascripts/discourse/templates/components/chat-live-pane.hbs
```diff
@@ -37,6 +37,8 @@
 
 {{chat-retention-reminder chatChannel=chatChannel}}
 
+<div class="chat-message-actions-mobile-anchor"></div>
```
jjaffeux, line 40: Longer term I want to move this here for real, but this requires many different refactoring first

martin-brennan (approved):

PR #1241, opened 2022-09-15
UX: Tweak padding on mobile

CvX:
Before / After

Screenshot 2022-09-14 at 22 06 48 Screenshot 2022-09-14 at 22 05 59

(actually, the composer is now about 6px lower than shown in "After" but I didn't want to retake the screenshot 😛 you can see the final position below)

How our message composer compares now to other apps:

inputs


jjaffeux:
Yes, LGTM, I have some of these changes as part of new emoji picker work

jjaffeux (approved):

PR #1247, opened 2022-09-16
UX: Add lazy loading and dominant-color placeholder for uploads

davidtaylorhq:
Also takes the opportunity to make `chat-upload` a Glimmer Component

(requires https://github.com/discourse/discourse/pull/18248)


nattsw (approved):
Lovely

PR #1248, opened 2022-09-19
UX: Add help text for webhook URL and remove from index page

davidtaylorhq:
Screenshot 2022-09-19 at 13 00 15


assets/javascripts/discourse/templates/admin-plugins-chat.hbs
```diff
@@ -92,6 +92,9 @@
     <div class="control-group">
       <label class="control-label">{{i18n "chat.incoming_webhooks.url"}}</label>
       <label>{{selectedWebhook.url}}</label>
```
jjaffeux, line 94: I know it's not you, but double label seems weird...

jjaffeux (approved):

PR #1251, opened 2022-09-21
DEV: Refactor user_count calculation and chat membership management

martin-brennan:
This commit intends to refactor and unify updates to chat channel
membership into a single `DiscourseChat::ChatChannelMembershipManager` class,
as well as making the `user_count` updates of the `ChatChannel` table
more consistent, as previously they happened in many places and it was
not always happening.

The `user_count` update is moved into a job, along with adding a new
`user_count_stale` column to the `ChatChannel` table. Whenever someone
is added/removed from the membership table, we update `user_count_stale`
to true _if it is not already_ and then enqueue a job in 3 seconds to update
the user count, which then sets `user_count_stale` back to false. This way
the query to count all the members, which can be expensive, is done out
of band, and the updates are throttled so we don't thrash the count unnecessarily.

We then send a MessageBus message to the client about the channel
metadata change, and if the user is on the member list for the channel
we update the count and reload the list of members.

### DIFF
--- a/app/jobs/regular/auto_join_channel_batch.rb
+++ b/app/jobs/regular/auto_join_channel_batch.rb
@@ -4,7 +4,10 @@ module Jobs
   class AutoJoinChannelBatch < ::Jobs::Base
     def execute(args)
       return "starts_at or ends_at missing" if args[:starts_at].blank? || args[:ends_at].blank?
-      return "End is higher than start" if args[:ends_at] < args[:starts_at]
+      start_user_id = args[:starts_at].to_i
+      end_user_id = args[:ends_at].to_i
+
+      return "End is higher than start" if end_user_id < start_user_id
 
       channel =
         ChatChannel.find_by(
@@ -20,8 +23,8 @@ module Jobs
 
       query_args = {
         chat_channel_id: channel.id,
-        start: args[:starts_at],
-        end: args[:ends_at],
+        start: start_user_id,
+        end: end_user_id,
         suspended_until: Time.zone.now,
         last_seen_at: 3.months.ago,
         channel_category: channel.chatable_id,
@@ -33,7 +36,7 @@ module Jobs
       # Only do this if we are running auto-join for a single user, if we
       # are doing it for many then we should do it after all batches are
       # complete for the channel in Jobs::AutoManageChannelMemberships
-      if args[:starts_at] == args[:ends_at]
+      if start_user_id == end_user_id
         DiscourseChat::ChatChannelMembershipManager.recalculate_user_count!(channel.id)
       end
 
--- a/app/jobs/regular/auto_join_channel_batch.rb
+++ b/app/jobs/regular/auto_join_channel_batch.rb
@@ -30,11 +30,12 @@ module Jobs
 
       new_member_ids = DB.query_single(create_memberships_query(category), query_args)
 
-      DB.exec(<<~SQL, channel_id: channel.id, joined: new_member_ids.size)
-        UPDATE chat_channels
-        SET user_count = user_count + :joined
-        WHERE id = :channel_id
-      SQL
+      # Only do this if we are running auto-join for a single user, if we
+      # are doing it for many then we should do it after all batches are
+      # complete for the channel in Jobs::AutoManageChannelMemberships
+      if args[:starts_at] == args[:ends_at]
+        DiscourseChat::ChatChannelMembershipManager.recalculate_user_count!(channel.id)
+      end
 
       ChatPublisher.publish_new_channel(channel.reload, User.where(id: new_member_ids))
     end
--- a/app/jobs/regular/auto_join_channel_batch.rb
+++ b/app/jobs/regular/auto_join_channel_batch.rb
@@ -37,7 +37,7 @@ module Jobs
       # are doing it for many then we should do it after all batches are
       # complete for the channel in Jobs::AutoManageChannelMemberships
       if start_user_id == end_user_id
-        DiscourseChat::ChatChannelMembershipManager.recalculate_user_count(channel)
+        DiscourseChat::ChatChannelMembershipManager.new(channel).recalculate_user_count
       end
 
       ChatPublisher.publish_new_channel(channel.reload, User.where(id: new_member_ids))
--- a/app/jobs/regular/auto_join_channel_batch.rb
+++ b/app/jobs/regular/auto_join_channel_batch.rb
@@ -37,7 +37,7 @@ module Jobs
       # are doing it for many then we should do it after all batches are
       # complete for the channel in Jobs::AutoManageChannelMemberships
       if start_user_id == end_user_id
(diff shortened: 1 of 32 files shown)


app/jobs/scheduled/update_user_counts_for_chat_channels.rb
```diff
@@ -2,25 +2,25 @@
 
 module Jobs
   class UpdateUserCountsForChatChannels < ::Jobs::Scheduled
-    every 2.hours
+    every 1.hour
 
+    # FIXME: This could become huge as the amount of channels grows, we
```
martin-brennan, line 7: Didn't want to change too many things in this PR, this can be done another time

app/services/chat_publisher.rb
```diff
@@ -195,6 +195,14 @@ def self.publish_channel_status(chat_channel)
     )
   end
 
+  def self.publish_chat_channel_metadata(chat_channel)
```
martin-brennan, line 198: Perhaps in future we will have more to add here to make this more useful

martin-brennan:
Side note, the whole `user_count` turning into `memberships_count` via the serializer thing is weird, we should just stick to one or the other. Personally I think `memberships_count` is better.

app/services/chat_publisher.rb
martin-brennan: Also interested to hear about this approach, I didn't think doing a joined/left channel event would be necessary (too noisy) and multiple people can join/leave at once with auto join channel so I didn't want to send tons of these through either. Member list only really needs to be updated in one place and DM channels cannot change members currently either.

lib/chat_channel_membership_manager.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  def self.follow_channel(user:, channel_id: nil, channel: nil)
+    ensure_channel_or_id_provided!(channel_id, channel)
+    membership =
+      find_for_user(user: user, channel_id: channel_id, channel: channel, initialize: true)
+
+    ActiveRecord::Base.transaction do
+      if !membership.following
+        membership.following = true
```
jjaffeux, line 27: `membership.update!(following: true)` ?

lib/chat_channel_membership_manager.rb
```diff
@@ -0,0 +1,83 @@
...
+        membership.save!
```
jjaffeux, line 28:
```suggestion
        membership.update!(following: true)
```

lib/chat_channel_membership_manager.rb
```diff
@@ -0,0 +1,83 @@
...
+        recalculate_user_count!(channel_id || channel.id)
+      end
+    end
+
+    membership
+  end
+
+  def self.unfollow_channel(user:, channel_id: nil, channel: nil)
```
jjaffeux, line 36: where is the `channel_id`  signature used? If we could I would prefer a signature like this one: `unfollow(acting_user:, channel:)`

app/jobs/scheduled/update_user_counts_for_chat_channels.rb
```diff
@@ -2,25 +2,25 @@
...
       current_count = chat_channel.user_count || 0
-      new_count =
-        chat_channel
-          .user_chat_channel_memberships
-          .joins(:user)
-          .where(following: true)
-          .merge(User.activated.not_suspended.not_staged)
-          .count
-
+      new_count = ChatChannelMembershipsQuery.call(chat_channel, count_only: true)
```
jjaffeux, line 19: Wonder if signature shouldn't be: `ChatChannelMembershipsQuery.count(channel)` ?

db/migrate/20220901034107_add_user_count_stale_to_channel.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddUserCountStaleToChannel < ActiveRecord::Migration[7.0]
+  def change
+    add_column :chat_channels, :user_count_stale, :boolean, default: false, null: false
```
jjaffeux, line 5: Shouldn't we consider that by default it's stale?

jjaffeux (approved):

db/migrate/20220901034107_add_user_count_stale_to_channel.rb
martin-brennan: Hmmm I don't think so, I think it should only be stale if something changes to increase/decrease this count.

app/jobs/scheduled/update_user_counts_for_chat_channels.rb
martin-brennan: Yes that's better, will change

lib/chat_channel_membership_manager.rb
martin-brennan: Yeah not so sure about this channel_id thing I did, I think in all cases we've had the channel already, so will get rid of this.

lib/chat_channel_membership_manager.rb
```diff
@@ -0,0 +1,83 @@
...
+      UserChatChannelMembership.find_or_initialize_by(params)
+    else
+      UserChatChannelMembership.includes(:user, :chat_channel).find_by(params)
+    end
+  end
+
+  def self.follow_channel(user:, channel_id: nil, channel: nil)
+    ensure_channel_or_id_provided!(channel_id, channel)
+    membership =
+      find_for_user(user: user, channel_id: channel_id, channel: channel, initialize: true)
```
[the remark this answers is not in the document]
martin-brennan: Good point, will change.

app/jobs/regular/auto_join_channel_batch.rb
```diff
@@ -30,11 +33,12 @@ def execute(args)
 
       new_member_ids = DB.query_single(create_memberships_query(category), query_args)
 
-      DB.exec(<<~SQL, channel_id: channel.id, joined: new_member_ids.size)
-        UPDATE chat_channels
-        SET user_count = user_count + :joined
-        WHERE id = :channel_id
-      SQL
+      # Only do this if we are running auto-join for a single user, if we
```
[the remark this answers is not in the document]
martin-brennan: Yes because we have a single user version of this here that's called by multiple things:

https://github.com/discourse/discourse-chat/blob/93e0cd8dcdfbff965357b2fbdaa2a271fd1bf556/lib/chat_channel_membership_manager.rb#L65-L72

I think the extra edge case of `auto_manage_channel_memberships` possibly only doing one user is fine here, shouldn't cause a double count because of the `user_count_stale` throttling

lib/chat_channel_membership_manager.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+class DiscourseChat::ChatChannelMembershipManager
```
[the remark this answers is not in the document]
martin-brennan: Actually yeah I think this will be better, changing

PR #1255, opened 2022-09-21
Revert "DEV: Temporarily skip reactions tests (#1254)"

keegangeorge:
Core is now passing, and we can revert the skipping of tests.


CvX (approved):

keegangeorge:
Tests should be passing after https://github.com/discourse/discourse/pull/18331 is merged into core. Will run tests again after.

PR #1262, opened 2022-09-27
FEATURE: Improve chat messages flagging.

romanrizzi:
This PR re-uses core's flag modal so users can select a flag type and send PMs when selecting notify_user or notify_moderators. It also lets staff take actions, or queue chat messages for review.

One difference is that we don't have the concept of "hiding" as we do for posts, so we have to delete messages instead on those scenarios.

### DIFF
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -429,20 +429,36 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
   end
 
   def flag
-    params.require([:chat_message_id])
-    chat_message = ChatMessage.includes(:chat_channel).find_by(id: params[:chat_message_id])
+    permitted_params =
+      params.permit(
+        %i[chat_message_id flag_type_id message is_warning take_action queue_for_review],
+      )
 
-    raise Discourse::InvalidParameters unless chat_message
-    set_channel_and_chatable_with_access_check(chat_channel_id: chat_message.chat_channel_id)
-    guardian.ensure_can_flag_chat_message!(chat_message)
+    chat_message =
+      ChatMessage.includes(:chat_channel).find_by(id: permitted_params[:chat_message_id])
+    raise Discourse::InvalidParameters.new(:chat_message_id) if !chat_message
 
-    if chat_message.reviewable_score_for(current_user).exists?
-      return render json: success_json # Already flagged
+    flag_type_id = permitted_params[:flag_type_id].to_i
+
+    if !ReviewableScore.types.values.include?(flag_type_id)
+      raise Discourse::InvalidParameters.new(:flag_type_id)
     end
 
-    reviewable = chat_message.add_flag(current_user)
-    ChatPublisher.publish_flag!(chat_message, current_user, reviewable)
-    render json: success_json
+    set_channel_and_chatable_with_access_check(chat_channel_id: chat_message.chat_channel_id)
+
+    result =
+      DiscourseChat::ChatReviewQueue.new.flag_message(
+        chat_message,
+        guardian,
+        flag_type_id,
+        permitted_params,
+      )
+
+    if result[:success]
+      render json: success_json
+    else
+      render_json_error(result[:errors])
+    end
   end
 
   def set_draft
--- a/assets/javascripts/discourse/components/chat-message.js
+++ b/assets/javascripts/discourse/components/chat-message.js
@@ -19,6 +19,8 @@ import { inject as service } from "@ember/service";
 import { popupAjaxError } from "discourse/lib/ajax-error";
 import discourseLater from "discourse-common/lib/later";
 import isZoomed from "discourse/plugins/discourse-chat/discourse/lib/zoom-check";
+import showModal from "discourse/lib/show-modal";
+import ChatMessageFlag from "discourse/plugins/discourse-chat/discourse/lib/chat-message-flag";
 
 let _chatMessageDecorators = [];
 
@@ -661,6 +663,27 @@ export default Component.extend({
     });
   },
 
+  // TODO(roman): For backwards-compatibility.
+  //   Remove after the 3.0 release.
+  _legacyFlag() {
+    bootbox.confirm(
+      I18n.t("chat.confirm_flag", {
+        username: this.message.user?.username,
+      }),
+      (confirmed) => {
+        if (confirmed) {
+          ajax("/chat/flag", {
+            method: "PUT",
+            data: {
+              chat_message_id: this.message.id,
+              flag_type_id: 7, // notify_moderators
+            },
+          }).catch(popupAjaxError);
+        }
+      }
+    );
+  },
+
   @action
   reply() {
     this.setReplyTo(this.message.id);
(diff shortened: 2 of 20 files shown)


assets/javascripts/discourse/lib/chat-message-flag.js
```diff
@@ -0,0 +1,95 @@
...
+  title() {
+    return "flagging.title";
+  }
+
+  customSubmitLabel() {
+    return "flagging.notify_action";
+  }
+
+  submitLabel() {
+    return "flagging.action";
```
martin-brennan, line 15: This should be something more specific to chat, like "Flag Message" instead of "Flag Post"

martin-brennan: Saw the other messages are all customized already, nice!

config/settings.yml
```diff
@@ -80,3 +80,6 @@ plugins:
     default: 10000
     hidden: true
     client: true
+  min_trust_to_flag_chat_messages:
```
martin-brennan, line 83: I think we should follow the pattern we have been doing in core recently, and no longer use trust-based settings. Instead use a `group_list` setting called e.g. `chat_message_flag_allowed_groups` and default it to auto group ID 11 (TL1). The PR I linked in core with another comment also makes a special shortcut method for group_list settings, adding `_map` to the end of the name to get an array of integers instead of having to do the `split("|")` thing all the time, .e.g `chat_message_flag_allowed_groups_map` and then you can use `user.in_any_groups?` too

lib/chat_review_queue.rb
```diff
@@ -0,0 +1,120 @@
...
+
+    UserSilencer.silence(
+      user,
+      Discourse.system_user,
+      silenced_till: auto_silence_duration.minutes.from_now,
+      reason: I18n.t("chat.errors.auto_silence_from_flags"),
+    )
+  end
+
+  def compoanion_pm_creator(chat_message, flagger, flag_type_id, opts)
```
martin-brennan, line 82:
```suggestion
  def companion_pm_creator(chat_message, flagger, flag_type_id, opts)
```

lib/guardian_extensions.rb
```diff
@@ -95,7 +95,7 @@ def can_see_chat_channel?(chat_channel)
   def can_flag_chat_messages?
     return false if @user.silenced?
 
-    @user.has_trust_level?(TrustLevel[SiteSetting.min_trust_to_flag_posts])
+    @user.has_trust_level?(TrustLevel[SiteSetting.min_trust_to_flag_chat_messages])
```
martin-brennan, line 98: Same comment about group vs trust level setting here

spec/serializer/chat_message_serializer_spec.rb
```diff
@@ -57,4 +57,86 @@
...
+          reviewable_ids: {
+            message_1.id => reviewable.id,
+          },
+        ).as_json
+
+      expect(serialized[:available_flags]).to be_empty
+    end
+
+    it "doesn't include notify_user for self-flags" do
+      a_guardian = Guardian.new(message_1.user)
```
martin-brennan, line 79:
```suggestion
      guardian_1 = Guardian.new(message_1.user)
```

minor nit, just for consistency, this variable name stood out to me but I don't think it needed to

assets/javascripts/discourse/components/chat-message.js
```diff
@@ -678,21 +701,19 @@ export default Component.extend({
...
-          ajax("/chat/flag", {
-            method: "PUT",
-            data: {
-              chat_message_id: this.message.id,
-            },
-          }).catch(popupAjaxError);
-        }
-      }
-    );
+    const targetFlagSupported =
```
martin-brennan, line 704: How come this is required? Won't the core PR https://github.com/discourse/discourse/pull/18251 add this and get merged before this one?

martin-brennan: Oh sorry I saw the above. Hmm I wonder if there is another way we can do this rather than calling `requirejs.entries` directly. Maybe it's not a problem, if it's not don't worry about this comment. I can confirm it worked fine locally since I didn't have the core PR checked out when I tested this at first.

martin-brennan (commented):
I think this looks great, you are obviously very familiar with reviewables so won't really comment on those things too much. Mainly my comments are around how the site settings are working.

app/controllers/chat_controller.rb
```diff
@@ -429,20 +429,36 @@ def move_messages_to_channel
...
+      params.permit(
+        %i[chat_message_id flag_type_id message is_warning take_action queue_for_review],
+      )
 
-    raise Discourse::InvalidParameters unless chat_message
-    set_channel_and_chatable_with_access_check(chat_channel_id: chat_message.chat_channel_id)
-    guardian.ensure_can_flag_chat_message!(chat_message)
+    chat_message =
+      ChatMessage.includes(:chat_channel).find_by(id: permitted_params[:chat_message_id])
+    raise Discourse::InvalidParameters.new(:chat_message_id) if chat_message.nil?
```
jjaffeux, line 439: do we need nil check here? `if !chat_message`

assets/javascripts/discourse/lib/chat-message-flag.js
```diff
@@ -0,0 +1,95 @@
...
+      ? { message: controller.message }
+      : {};
+
+    if (opts) {
+      params = Object.assign(params, opts);
+    }
+
+    controller.send("hideModal");
+
+    ajax("/chat/flag", {
```
jjaffeux, line 65: I really prefer to reurn async calls so they can be thened if needed

spec/lib/chat_review_queue_spec.rb
```diff
@@ -0,0 +1,304 @@
...
+            queue_for_review: true,
+          )
+        end.to raise_error(Discourse::InvalidAccess)
+      end
+    end
+
+    context "when the auto silence threshold is met" do
+      it "silences the user" do
+        SiteSetting.chat_auto_silence_from_flags_duration = 1
+        flagger.update(trust_level: TrustLevel[4]) # Increase Score due to TL Bonus.
```
jjaffeux, line 279: Would probably use update! for all theses flagger.update! calls

spec/lib/chat_review_queue_spec.rb
```diff
@@ -0,0 +1,304 @@
...
+        subject.flag_message(
+          message,
+          admin_guardian,
+          ReviewableScore.types[:notify_user],
+          message: flag_message,
+          is_warning: true,
+        )
+
+        warning = UserWarning.find_by(user: message.user)
+        expect(warning).to be_present
```
jjaffeux, line 131: Should be equivalent?

```suggestion
        expect(UserWarning.exists?(user: message.user)).to eq(true)
```

jjaffeux (approved):

config/settings.yml
jjaffeux: agreed group based is much better

assets/javascripts/discourse/components/chat-message.js
romanrizzi: Yeah, I tested it with and without the core refactor, and it works. I don't think there's a fancier way of doing it other than calling `require(...)` and catching the exception.

Can't you think of any reasons not to use `requirejs.entries`? I know it's hackish, but guaranteeing backward compatibility is tricky, and based on my experience so far, it's hard to come up with a generic solution for this problem. Also, this is much more common on the Ruby side.

martin-brennan: Yeah it's no problem it just stood out to me and it seemed hackish, but if we do it commonly and there's no better way then it's fine by me :+1:

PR #1268, opened 2022-09-30
FIX: Make chat_allow_uploads apply to DM channels

martin-brennan:
This setting should have applied to both public
and DM channels from the beginning, making it
so it works correctly and updating the description accordingly.


assets/javascripts/discourse/components/chat-composer.js
```diff
@@ -44,10 +44,7 @@ export default Component.extend(TextareaTextManipulation, {
   inProgressUploads: null,
   composerEventPrefix: "chat",
   composerFocusSelector: ".chat-composer-input",
-  canAttachUploads: or(
-    "siteSettings.chat_allow_uploads",
-    "chatChannel.isDirectMessageChannel"
-  ),
+  canAttachUploads: reads("siteSettings.chat_allow_uploads"),
```
tgxworld, line 47: Just confirming we have a test that covers this?

tgxworld (approved):

assets/javascripts/discourse/components/chat-composer.js
martin-brennan: Good question, will check!

martin-brennan: Done

PR #1269, opened 2022-09-30
FIX: Tighten flagging restrictions for chat messages.

romanrizzi:
For all messages, make sure the user can see the channel to flag and limit the number of flags a user can create, which is the same we do for post actions.

For each message, prevent multiple users from flagging the same message for the same reason following the post action creator guidelines:

* A reviewable can only have one pending flag of each type, except for notify_moderators.
* Users can only flag a message once until it gets reviewed by staff.
* Enforce a cooldown window after review unless the poster edits the message.

### DIFF
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -429,13 +429,17 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
   end
 
   def flag
+    RateLimiter.new(current_user, "flag_chat_message", 4, 1.minutes).performed!
+
     permitted_params =
       params.permit(
         %i[chat_message_id flag_type_id message is_warning take_action queue_for_review],
       )
 
     chat_message =
-      ChatMessage.includes(:chat_channel).find_by(id: permitted_params[:chat_message_id])
+      ChatMessage.includes(:chat_channel, :revisions).find_by(
+        id: permitted_params[:chat_message_id],
+      )
     raise Discourse::InvalidParameters.new(:chat_message_id) if !chat_message
 
     flag_type_id = permitted_params[:flag_type_id].to_i
--- a/lib/chat_review_queue.rb
+++ b/lib/chat_review_queue.rb
@@ -15,6 +15,13 @@ class DiscourseChat::ChatReviewQueue
     guardian.ensure_can_flag_chat_message!(chat_message)
     guardian.ensure_can_flag_message_as!(chat_message, flag_type_id, opts)
 
+    existing_reviewable = Reviewable.includes(:reviewable_scores).find_by(target: chat_message)
+
+    if !can_flag_again?(existing_reviewable, chat_message, guardian.user, flag_type_id)
+      result[:errors] << I18n.t("reviewables.already_handled")
+      return result
+    end
+
     if opts[:message].present? &&
          ReviewableScore.types.slice(:notify_user, :notify_moderators).values.include?(flag_type_id)
       creator = companion_pm_creator(chat_message, guardian.user, flag_type_id, opts)
@@ -116,5 +123,30 @@ class DiscourseChat::ChatReviewQueue
     end
 
     PostCreator.new(flagger, create_args)
+  end
+
+  def can_flag_again?(reviewable, message, flagger, flag_type_id)
+    return true if reviewable.blank?
+
+    flagger_has_no_pending_flags =
+      reviewable.reviewable_scores.none? { |rs| rs.user == flagger && rs.pending? }
+
+    if flagger_has_no_pending_flags && flag_type_id == PostActionType.types[:notify_moderators]
+      return true
+    end
+
+    flag_not_used =
+      reviewable.reviewable_scores.none? do |rs|
+        rs.reviewable_score_type == flag_type_id && rs.pending?
+      end
+    not_handled_recently =
+      reviewable.pending? ||
+        reviewable.updated_at < SiteSetting.cooldown_hours_until_reflag.to_i.hours.ago
+
+    latest_revision = message.revisions.last
+    edited_since_last_review = latest_revision && latest_revision.updated_at > reviewable.updated_at
+
+    flag_not_used && flagger_has_no_pending_flags &&
+      (not_handled_recently || edited_since_last_review)
   end
 end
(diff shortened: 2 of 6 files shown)


app/controllers/chat_controller.rb
```diff
@@ -429,13 +429,17 @@ def move_messages_to_channel
...
       params.permit(
         %i[chat_message_id flag_type_id message is_warning take_action queue_for_review],
       )
 
     chat_message =
-      ChatMessage.includes(:chat_channel).find_by(id: permitted_params[:chat_message_id])
+      ChatMessage.includes(:chat_channel, :revisions).find_by(
+        id: permitted_params[:chat_message_id],
+      )
     raise Discourse::InvalidParameters.new(:chat_message_id) if !chat_message
```
jjaffeux, line 443: How about we replace all of this by:

```
chat_message = ChatMessage.includes(...).find(id)
```

lib/chat_review_queue.rb
```diff
@@ -117,4 +124,29 @@ def companion_pm_creator(chat_message, flagger, flag_type_id, opts)
 
     PostCreator.new(flagger, create_args)
   end
+
+  def can_flag_again?(reviewable, message, flagger, flag_type_id)
+    return true if reviewable.blank?
+
+    flagger_has_no_pending_flags =
```
jjaffeux, line 131: up to you but I would generally fo with the inverse form:

```
flagger_has_pending_flags =  ...

if !flagger_pending_flags
  ...
end
```

jjaffeux: The same comment is valid for the next lines which are all using "not"

lib/chat_review_queue.rb
```diff
@@ -117,4 +124,29 @@ def companion_pm_creator(chat_message, flagger, flag_type_id, opts)
...
+      reviewable.reviewable_scores.none? { |rs| rs.user == flagger && rs.pending? }
+
+    if flagger_has_no_pending_flags && flag_type_id == PostActionType.types[:notify_moderators]
```
jjaffeux, line 134: I don't see any test using this :notify_moderators

spec/lib/chat_review_queue_spec.rb
```diff
@@ -27,6 +27,56 @@
...
+        subject.flag_message(message, admin_guardian, ReviewableScore.types[:spam])
+
+      expect(second_flag_result[:success]).to eq(false)
+      expect(second_flag_result[:errors]).to contain_exactly(I18n.t("reviewables.already_handled"))
+    end
+
+    it "allow users to re-flag using the same flag type after the cooldown period" do
+      subject.flag_message(message, guardian, ReviewableScore.types[:spam])
+
+      reviewable = ReviewableChatMessage.last
```
jjaffeux, line 56: Might be worth some "update_reviable" helper

spec/lib/chat_review_queue_spec.rb
```diff
@@ -27,6 +27,56 @@
...
+      subject.flag_message(message, guardian, ReviewableScore.types[:spam])
+
+      second_flag_result =
+        subject.flag_message(message, guardian, ReviewableScore.types[:off_topic])
+
+      expect(second_flag_result[:success]).to eq(false)
+      expect(second_flag_result[:errors]).to contain_exactly(I18n.t("reviewables.already_handled"))
+    end
+
+    it "raises an error if a different user uses the same flag but we recently handled it" do
```
jjaffeux, line 40: maybe we should note it's related to cooldown setting as we test both cases

app/controllers/chat_controller.rb
```diff
@@ -429,13 +429,17 @@ def move_messages_to_channel
   end
 
   def flag
+    RateLimiter.new(current_user, "flag_chat_message", 4, 1.minutes).performed!
```
jjaffeux, line 432: Can we test this?

app/controllers/chat_controller.rb
romanrizzi: That would return a different exception if not found. Let me check if it's handled the same way in the `ApplicationController`.

Flink: I guess we could use `rescue_from` in this controller for example. Not sure it would be better though?

lib/chat_review_queue.rb
Flink: From the PR description, it says `“A reviewable can only have one pending flag of each type, except for notify_moderators”` so I guess here we should use `||` instead of `&&`, right?

spec/lib/chat_review_queue_spec.rb
```diff
@@ -27,6 +27,56 @@
       }.to raise_error(Discourse::InvalidAccess)
     end
 
+    it "raises an error if the user already flagged the post" do
```
Flink, line 30: I know there are a lot of things moving at the moment regarding how we write our specs but usually when there’s an `if` in a description it’s a good indicator that a context should be used. This allows for easier setup and grouping of things like related `let`. Not sure where we’re at on this front in the chat project though :slightly_smiling_face: 
Anyway, we could then rewrite this block as:
```ruby
context "when the user already flagged the post" do
  let(:second_flag_result) { queue.flag_message(message, guardian, ReviewableScore.types[:off_topic]) }

  before do
    queue.flag_message(message, guardian, ReviewableScore.types[:spam])
  end

  it "returns an error" do
    expect(second_flag_result).to include success: false, errors: I18n.t("reviewables.already_handled")
  end
end
```
I also named the subject (`queue`) as it’s recommended to do so in the rspec style guide.

romanrizzi: Looks much cleaner. Thanks for the suggestion!

lib/chat_review_queue.rb
romanrizzi: Since `notify_moderators`is always available, we need to make sure the user has no other pending flags. You can only flag a message once.

spec/lib/chat_review_queue_spec.rb
romanrizzi: Re-organizing tests make it better. I don't think we need a helper for the time being

jjaffeux (approved):

Flink (approved):

PR #1270, opened 2022-10-03
FIX: enforces chat_channel_id when present

jjaffeux:
This will deserve an heavier refactoring in the future, and more importantly moving all this logic under something like `/chat/chat_channels/:id/messages?target_message_id=865` to enforce message being a child from the channel.

This commit however prevents the issue for now and ensures this case is tracked for future refactoring with a test.

### DIFF
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -498,7 +498,7 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
   end
 
   def find_chat_message
-    @message = ChatMessage.unscoped.includes(chat_channel: :chatable)
+    @message = preloaded_chat_message_query.with_deleted
     @message = @message.where(chat_channel_id: params[:chat_channel_id]) if params[:chat_channel_id]
     @message = @message.find_by(id: params[:message_id])
     raise Discourse::NotFound unless @message
--- a/app/controllers/chat_controller.rb
+++ b/app/controllers/chat_controller.rb
@@ -498,9 +498,9 @@ class DiscourseChat::ChatController < DiscourseChat::ChatBaseController
   end
 
   def find_chat_message
-    @message =
-      ChatMessage.unscoped.includes(chat_channel: :chatable).find_by(id: params[:message_id])
-
+    @message = ChatMessage.unscoped.includes(chat_channel: :chatable)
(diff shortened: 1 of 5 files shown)


assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1431,9 +1431,11 @@ export default Component.extend({
     });
   },
 
-  _handle429Errors(error) {
+  _handleErrors(error) {
```
Flink, line 1434: Is this function going to handle more errors based on statuses? If you think that’s the case then maybe we could use a `switch` statement instead? Something like that:
```javascript
switch (error?.jqXHR?.status) {
  case 429:
  case 404:
    popupAjaxError(error);
  default:
    throw error;
}
```

Flink (approved):

app/controllers/chat_controller.rb
```diff
@@ -498,9 +498,9 @@ def find_chatable
   end
 
   def find_chat_message
-    @message =
-      ChatMessage.unscoped.includes(chat_channel: :chatable).find_by(id: params[:message_id])
-
+    @message = ChatMessage.unscoped.includes(chat_channel: :chatable)
```
ZogStriP, line 501: Can the `unscoped` be abused in any ways? Usually, only staff/admins can query stuff unscoped. Is the `find_chat_message` properly protected?

assets/javascripts/discourse/components/chat-live-pane.js
jjaffeux: I don't know, but the switch doesn’t hurt :+1:

assets/javascripts/discourse/components/chat-live-pane.js
```diff
@@ -1431,11 +1431,14 @@ export default Component.extend({
...
-    if (error?.jqXHR?.status === 429) {
-      popupAjaxError(error);
-    } else {
-      throw error;
+  _handleErrors(error) {
+    switch (error?.jqXHR?.status) {
+      case 429:
+      case 404:
+        popupAjaxError(error);
+        return;
```
Flink, line 1439: Not sure it will change any behavior but I think we should use a `break` statement here instead.

jjaffeux: In this case that won't change anything but I can do this yes :)

PR #1274, opened 2022-10-04
FIX: correctly censors excerpts

jjaffeux:


jjaffeux:
Making sure all these cases are handled:

Screenshot 2022-10-04 at 15 30 52

spec/serializer/chat_in_reply_to_serializer_spec.rb
```diff
@@ -19,4 +19,14 @@
       end
     end
   end
+
+  describe "#excerpt" do
+    it "censors words" do
```
Flink, line 24: I think we could refactor these specs a bit (not just yours), `fab!(:message_1)` doesn’t seem that useful as it’s only used by the `#user` spec:
```ruby
RSpec.describe ChatInReplyToSerializer do
  subject(:serializer) { described_class.new(message, scope: guardian, root: nil) }

  fab!(:chat_channel) { Fabricate(:chat_channel) }
  let(:guardian) { Guardian.new(Fabricate(:user)) }

  describe "#user" do
    let(:message) { Fabricate(:chat_message, user: Fabricate(:user), chat_channel: chat_channel) }

    context "when user has been destroyed" do
      before do
        message.user.destroy!
        message.reload
      end

      it "returns a placeholder user" do
        expect(serializer.as_json[:user][:username]).to eq(I18n.t("chat.deleted_chat_username"))
      end
    end
  end

  describe "#excerpt" do
    let(:watched_word) { Fabricate(:watched_word, action: WatchedWord.actions[:censor]) }
    let(:message) { Fabricate(:chat_message, message: "ok #{watched_word.word}") }

    it "censors words" do
      expect(serializer.as_json[:excerpt]).to eq("ok ■■■■■")
    end
  end
end
```
Wdyt?

Flink (approved):

spec/serializer/chat_in_reply_to_serializer_spec.rb
jjaffeux: sure works for me :+1:

PR #1277, opened 2022-10-05
FIX: prefers /chat/channel/:id/:title?messageId=x link

jjaffeux:
Prior to this fix we were constructing links using: `/chat/message/id` in chat transcript, which is prone to errors.


Flink (approved):
Just a question/suggestion: we could maybe have a route like `/chat/channel/:id/:title/message/:id`?

jjaffeux:
> Just a question/suggestion: we could maybe have a route like `/chat/channel/:id/:title/message/:id`?

Yes I will do it in the future, but for now, Im just moving from two possibilities to one.

PR #1278, opened 2022-10-05
UX: moves new channel button to browse view

jjaffeux:
Removes the dropdown with browse/create from the sidebar. As a side effect:
- it's now a button using the pencil icon, leading to browse page
- the new channel button is at the top of browse page

It's only changing behavior for staff, regular users were already not seeing the dropdown as they can’t create channels.


jjaffeux:
Screenshot 2022-10-05 at 14 01 20

nlalonde (approved):

PR #1282, opened 2022-10-05
FEATURE: Link general chat channel to general category

oblakeerickson:
Now that the General category is being seeded for new sites we can link
the general chat channel to it.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -101,9 +101,6 @@ en:
         multi_user: "%{users}"
         multi_user_truncated: "%{users} and %{leftover} others"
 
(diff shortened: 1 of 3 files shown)


config/locales/server.en.yml
```diff
@@ -103,6 +103,7 @@ en:
 
       default_titles:
         site_feedback: "General"
```
jjaffeux, line 105: Should we rename this one? "Site feedback" ?

jjaffeux: even better then :p

CvX (approved):

PR #1283, opened 2022-10-06
FEATURE: Inline video player for video uploads

martin-brennan:
This commit makes it so user-uploaded videos in chat are shown inside a video player rather than as a
downloadable attachment. Same width and height rules as youtube videos defined in 811595b8b13bec45abaf18133379e85d5c670ca3 apply.

image


jjaffeux (approved):

jjaffeux:
Looks good to me, thanks :+1: Sharing videos in 2022 is kinda baseline assumption on the internets, good to have it :heart:

PR #1284, opened 2022-10-06
DEV: Refactor the ChatChannel model

Flink:
Since its inception, the `ChatChannel` model evolved quite a bit.
Originally its `chatable` association allowed to link it to any existing model but currently it’s only associated to two models: `Category` and `DirectMessageChannel`.
    
Currently there are a lot of conditionals in the `ChatChannel` model to determine logic based on the associated chatable. It’s a kind of “manual polymorphism”.
This patch addresses this issue by making `ChatChannel` an STI model which allows to create two new models: `CategoryChannel` and `DMChannel`. Those models have dedicated logic allowing us to remove the extra conditionals in the `ChatChannel` model. Then we only have one entry point where we have some logic to choose which model to use based on the provided chatable.


app/models/d_m_channel.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
```
jjaffeux, line 1: is it expected to be d_m_channel.rb and not dm_channel.rb ?

jjaffeux:
@martin-brennan if you work on slugs, take this PR into account please

app/models/d_m_channel.rb
Flink: Yes this way the naming is compatible with what zeitwerk is expecting. `dm_channel` would translate to `DmChannel` instead.
But this is (hopefully) a temporary model name, at some point we should be able to merge the existing `DirectMessageChannel` model in it (and use the proper name).

martin-brennan:
Thanks for pointing this out @jjaffeux  -- At first thought I think based on this the new `DMChannel` and `CategoryChannel` would have their own `generate_slug` or similar methods that call `Slug.for(name/title)`, then the slug itself would be saved on the main channel model. But, will do some experimenting.

jjaffeux:
What's left to be done on this from your point of view @Flink ?

app/models/category_channel.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class CategoryChannel < ChatChannel
+  alias_attribute :category, :chatable
+
+  delegate :read_restricted?, to: :category
+
+  def category_channel?
+    true
+  end
+  alias public_channel? category_channel?
```
jjaffeux, line 11: should we mark some of these calls as deprecated?

app/models/chat_channel.rb
```diff
@@ -65,66 +78,22 @@ def url
...
-    case chatable_type
-    when "Category"
-      chatable.name
-    when "DirectMessageChannel"
-      chatable.chat_channel_title_for_user(self, user)
-    end
+  def allowed_user_ids
+    nil
   end
+  alias allowed_group_ids allowed_user_ids
```
jjaffeux, line 96: same should we mark one as deprecated?

Flink:
> What's left to be done on this from your point of view @Flink ?

I need to write some specs for the new models and the post migration. After that I think this first iteration should be ready :slightly_smiling_face:

app/controllers/chat_controller.rb
```diff
@@ -39,7 +39,7 @@ def enable_chat
     elsif chat_channel
       return render_json_error I18n.t("chat.already_enabled")
     else
-      chat_channel = ChatChannel.new(chatable: @chatable)
+      chat_channel = ChatChannel.for(@chatable).new(chatable: @chatable)
```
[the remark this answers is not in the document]
Flink: Yes I’m not completely satisfied either with this API, I had another version which was instantiating a new object but then it was a bit weird with the `create!` call we have elsewhere. Well I’ll take another look and see if it can be improved :grin:

app/models/category_channel.rb
Flink: Not sure as those methods are used at different places in different contexts. Probably at some point we could (should?) simplify the code which uses them.

app/models/chat_channel.rb
Flink: They are not used in the same contexts, I think one is expected for direct messages and the other one is expected for public channels. (so these methods are redefined in `DMChannel` and `CategoryChannel`)

jjaffeux: Ok sounds great, Im around all day if you need me to push this to the finish line.

app/controllers/chat_channels_controller.rb
```diff
@@ -48,7 +48,7 @@ def create
     auto_join_users = ActiveRecord::Type::Boolean.new.deserialize(params[:auto_join_users]) || false
 
     chat_channel =
-      ChatChannel.create!(
+      ChatChannel.type_for(chatable).create!(
```
jjaffeux, line 51: What about doing this API:


```ruby
chatable.create_chat_channel!(
  name: params[:name],
  description: params[:description]
)
```

?

app/controllers/chat_controller.rb
```diff
@@ -39,7 +39,7 @@ def enable_chat
     elsif chat_channel
       return render_json_error I18n.t("chat.already_enabled")
     else
-      chat_channel = ChatChannel.new(chatable: @chatable)
+      chat_channel = ChatChannel.for(@chatable)
```
jjaffeux, line 42: Same questions here:

```ruby
@chatable.chat_channel
```

db/post_migrate/20221018091412_migrate_chat_channels.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class MigrateChatChannels < ActiveRecord::Migration[7.0]
+  def down
+  end
+
+  def up
+    DB.exec("UPDATE chat_channels SET type='CategoryChannel' WHERE chatable_type = 'Category'")
```
jjaffeux, line 8: We are dropping `chatable_type` in a future PR right?

jjaffeux: If possible in the future PR we should make sure any existing channel with chatable_type  `Tag`or `Site` should be dropped. I don't think we ever did this cleanup.

jjaffeux:
I made few minor notes. I also tried it locally with existing records and it all seemed to work nicely :+1:

LGTM, let's talk of it tomorrow, thanks.

app/models/chat_channel.rb
```diff
@@ -32,6 +32,21 @@ class ChatChannel < ActiveRecord::Base
...
+      else
+        raise "Unknown chatable type #{chatable.class}"
+      end
+    end
+
+    def for(chatable)
+      type_for(chatable).new(chatable: chatable)
+    end
+
     def public_channel_chatable_types
```
martin-brennan, line 50: Will we remove/refactor this at some point?

app/models/chat_channel.rb
```diff
@@ -32,6 +32,21 @@ class ChatChannel < ActiveRecord::Base
         }
```
martin-brennan, line 33: We need to change the scopes for this model (`in_public_channel` and `in_dm_channel`) to point to the new column

app/models/chat_channel.rb
```diff
@@ -65,66 +80,23 @@ def url
...
-
-  def title(user)
-    return chatable.chat_channel_title_for_user(self, user) if direct_message_channel?
-    return name if name.present?
-
-    title_from_chatable
+  def category_channel?
+    false
   end
+  alias direct_message_channel? category_channel?
```
martin-brennan, line 90: I am confused by some of these aliases, they don't seem to line up to me, maybe I am reading it wrong? How is `category_channel?` an alias for `direct_message_channel?`

app/controllers/chat_channels_controller.rb
martin-brennan: I think this and the other suggestion would be more convenient too, even if we have to reopen the class via the plugin API to do it.

martin-brennan (commented):
I think this is good so far, but is this kind of a v1 PR? It seems like there are a lot of places that still look at the `chatable_type` column. The migrations look fine to me

db/post_migrate/20221018091412_migrate_chat_channels.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class MigrateChatChannels < ActiveRecord::Migration[7.0]
+  def down
```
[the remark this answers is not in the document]
jjaffeux: It's post migrate so it's only doing some consistency work, the real migration adding the `type` col is using `change` and would remove the col if reverted.

db/post_migrate/20221018091412_migrate_chat_channels.rb
jjaffeux: They are still being used, `chatable_type`/`chatable_id` represent the link to the chatable object. `type` is only the type of the STI table inheriting from `ChatChannel` (d_m_channel, category_channel). Probably further improvements to be done in the future.

app/models/chat_channel.rb
jjaffeux: It's mostly for defualt state, but we will try to make something simpler/more obvious.

app/models/chat_channel.rb
jjaffeux: `type` is not replacement of `chatable_type`

app/models/chat_channel.rb
jjaffeux: We are clearly not yet on the final target, trying to make this in multiple PRs, chatable is still very much present, but yes we will most likely refactor this at some point too :+1:

Flink:
Everything we discussed about has been addressed I think :grin:

db/post_migrate/20221018091412_migrate_chat_channels.rb
martin-brennan: Ah I see my bad, got turned around a bit from the existence of the other models :+1:

app/models/chat_channel.rb
martin-brennan: Yes sorry this is my bad I confused myself

db/post_migrate/20221018091412_migrate_chat_channels.rb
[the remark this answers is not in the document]
martin-brennan: I still find it odd to have an empty `down` but not super concerned either

martin-brennan (approved):
Thanks @Flink and @jjaffeux I find it much easier to follow after the changes, sorry for mixing up the details around the existing columns :+1:

PR #1286, opened 2022-10-07
FEATURE: Enable flagging for DMs

romanrizzi:
For flagged DMs, we'll include a transcript of the previous messages (up to ten), which we store in a companion topic.

Additionally, we store the cooked content of the message in the reviewable to prevent users from editing it before staff reviews it.

Screen Shot 2022-10-07 at 17 49 51
Screen Shot 2022-10-07 at 17 50 00

### DIFF
--- a/lib/chat_review_queue.rb
+++ b/lib/chat_review_queue.rb
@@ -12,6 +12,12 @@ class DiscourseChat::ChatReviewQueue
   def flag_message(chat_message, guardian, flag_type_id, opts = {})
     result = { success: false, errors: [] }
 
+    is_notify_type =
+      ReviewableScore.types.slice(:notify_user, :notify_moderators).values.include?(flag_type_id)
+    is_dm = chat_message.chat_channel.direct_message_channel?
+
+    raise Discourse::InvalidParameters.new(:flag_type) if is_dm && is_notify_type
+
     guardian.ensure_can_flag_chat_message!(chat_message)
     guardian.ensure_can_flag_message_as!(chat_message, flag_type_id, opts)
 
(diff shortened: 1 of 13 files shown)


martin-brennan (approved):
Looks excellent, nice one! :ok_hand:

lib/chat_review_queue.rb
```diff
@@ -125,6 +136,50 @@ def companion_pm_creator(chat_message, flagger, flag_type_id, opts)
     PostCreator.new(flagger, create_args)
   end
 
+  def find_or_create_transcript(chat_message, flagger, existing_reviewable)
+    previous_message_ids =
```
jjaffeux, line 140: What kind of check do we have to ensure transcript will never surface messages to someone who shouldn't see them?

martin-brennan: IMO any access access control we add for this should be outside the `ChatTranscriptService`, that should be kept as a dumb thing that just spits out the markdown for the provided IDs. In say ing that it does at least make sure the messages all belong to the same channel so you can't just give it arbitrary IDs. When using the quote in the UI we run an access check on the channel before quoting the messages so that is safe.

In this case, we are sending the context to the moderator group, so it should always be safe here too right?

jjaffeux: Yes Im not suggesting we add anything to transcript service. Im mostly asking Roman to take a step back and carefuly think about this.

Its potentially super easy to pass down ids to transcript service, so we should double check it and ensures we have tests for this.

martin-brennan: Yep no problem at all, it's good to have these discussions :+1:

romanrizzi: The transcript is a PM between the system user and the moderator's group, so all PM restrictions should apply. I'll add some tests to assert they're the only users with access to the PM and prevent possible regressions, but not sure if there's anything else to do here.

jjaffeux: Im fine with just extra tests :+1:

jjaffeux (approved):

PR #1287, opened 2022-10-07
DEV: Refactor bootbox dialogs

pmusaraj:
This also includes a small admin UI fix for incoming webhooks. 

Before

image

After

image


jjaffeux:
How does it work for people not on a recent discourse version?

pmusaraj:
@jjaffeux I added an entry to `.discourse-compatibility` so that this change only applies to recent Discourse core versions.

jjaffeux:
> @jjaffeux I added an entry to `.discourse-compatibility` so that this change only applies to recent Discourse core versions. 

Thanks

jjaffeux (approved):

PR #1289, opened 2022-10-10
FIX: Staff should always be able to chat

martin-brennan:
The `chat_allowed_groups` setting is changing in this commit to reflect other group settings, which always allow staff to do the related action regardless of whether the staff user is in one of the designated groups or not.

### DIFF
--- a/app/controllers/chat_channels_controller.rb
+++ b/app/controllers/chat_channels_controller.rb
@@ -105,7 +105,11 @@ class DiscourseChat::ChatChannelsController < DiscourseChat::ChatBaseController
 
     users = User.joins(:user_option).where.not(id: current_user.id)
     if !DiscourseChat.allowed_group_ids.include?(Group::AUTO_GROUPS[:everyone])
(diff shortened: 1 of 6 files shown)


app/controllers/chat_channels_controller.rb
```diff
@@ -105,7 +105,11 @@ def search
 
     users = User.joins(:user_option).where.not(id: current_user.id)
     if !DiscourseChat.allowed_group_ids.include?(Group::AUTO_GROUPS[:everyone])
-      users = users.joins(:groups).where(groups: { id: DiscourseChat.allowed_group_ids })
+      users =
+        users
+          .joins(:groups)
+          .where(groups: { id: DiscourseChat.allowed_group_ids })
+          .or(users.joins(:groups).where("users.admin = true OR users.moderator = true"))
```
tgxworld, line 112:
```suggestion
          .or(users.joins(:groups).staff
```

tgxworld (approved):

PR #1292, opened 2022-10-12
FIX: Use sidebar contentCSSClass for muted channels

martin-brennan:
In https://github.com/discourse/discourse/pull/18555 we are adding a sidebar function, contentCSSClass, for section links, and this commit uses it to make the muted channels low opacity again, since this is not currently working with the new sidebar.

### DIFF
--- a/assets/javascripts/discourse/initializers/chat-sidebar.js
+++ b/assets/javascripts/discourse/initializers/chat-sidebar.js
@@ -108,6 +108,12 @@ export default {
                 ? "urgent"
                 : "unread";
             }
+
+            get contentCSSClass() {
+              return this.channel.current_user_membership.muted
+                ? "--muted"
+                : "";
+            }
           };
 
           const SidebarChatChannelsSection = class extends BaseCustomSidebarSection {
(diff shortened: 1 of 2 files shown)


techAPJ (approved):

assets/javascripts/discourse/initializers/chat-sidebar.js
```diff
@@ -108,6 +108,12 @@ export default {
                 ? "urgent"
                 : "unread";
             }
+
+            get contentCSSClass() {
+              return this.channel.current_user_membership.muted
+                ? "--muted"
```
tgxworld, line 114: This is kind of odd, should there be a prefix for the class?

tgxworld: Screenshot from 2022-10-12 14-59-32

This definitely feels off because we usually don't name our CSS classes like this.

martin-brennan: Good point, will fix

tgxworld: No worries I fixed it in https://github.com/discourse/discourse-chat/commit/aa5f731f1d426fca902a2c5370646f4052cbf0c7v yesterday while looking at something related.

PR #1295, opened 2022-10-13
DEV: Add proper CSS class for sidebar section link content

tgxworld:
* Adds tests to ensure that CSS class is added when channels are muted
as well.


ZogStriP (approved):

martin-brennan:
Thanks for this @tgxworld , much better

PR #1297, opened 2022-10-14
DEV: Add slug column to chat_channel model

martin-brennan:
This commit just adds the column and index,
backfilling and filling it when creating/updating
will come later pending other refactors in
https://github.com/discourse/discourse-chat/pull/1284


martin-brennan:
Test failures unrelated flakys

techAPJ (approved):

PR #1302, opened 2022-10-17
DEV: Use new setupHashtagAutocomplete + registerHashtagSearchParam APIs

martin-brennan:
Include context in the setupHashtagAutocomplete call, and use registerHashtagSearchParam to set up category + tag search for the chat composer. We cannot add channel search yet because we need to be able to look up channels by slugs first and also choose/render selected channels. More core work still to come.

c.f. https://github.com/discourse/discourse/pull/18592

**Do not merge until the core PR has merged!**

### DIFF
--- a/assets/javascripts/discourse/initializers/chat-setup.js
+++ b/assets/javascripts/discourse/initializers/chat-setup.js
@@ -42,6 +42,11 @@ export default {
         });
       }
 
+      if (this.siteSettings.enable_experimental_hashtag_autocomplete) {
+        api.registerHashtagSearchParam("category", "chat-composer", 100);
+        api.registerHashtagSearchParam("tag", "chat-composer", 50);
+      }
+
       // we want to decorate the chat quote dates regardless
       // of whether the current user has chat enabled
       api.decorateCookedElement(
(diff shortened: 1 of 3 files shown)


jjaffeux:
I assume failures are due to core not being merged?

jjaffeux:
I think we should be able to test a test rather easily for this no? No need to go crazy, just a single one to ensure the channel autocomplete is showing.

martin-brennan:
Sure I will add a test @jjaffeux and yes I think it's core but will double check the tests locally.

assets/javascripts/discourse/initializers/chat-setup.js
```diff
@@ -43,6 +43,11 @@ export default {
         });
       }
 
+      if (this.siteSettings.enable_experimental_hashtag_autocomplete) {
+        api.registerHashtagSearchParam("category", "chat-composer", 100);
+        api.registerHashtagSearchParam("tag", "chat-composer", 50);
+      }
```
lis2, line 49: Related to https://github.com/discourse/discourse/pull/18592/files#r998915006

Do you think we should use separate api block like `withPluginApi("0.14.1"`

martin-brennan: I don’t want to bump it yet since it’s experimental I’ll do that once it’s more solid

lis2 (approved):

test/javascripts/acceptance/composer-hashtag-autocomplete-test.js
```diff
@@ -0,0 +1,68 @@
+import { setCaretPosition } from "discourse/lib/utilities";
```
jjaffeux, line 1: great thanks

PR #1303, opened 2022-10-17
FEATURE: improved emoji selection

jjaffeux:
This pull request includes the following changes:
- a fully revamped emoji picker: fast, accessible and clean
- the ability to use an emoji picker when composing a message (accessible from the dropdown actions at the left of the composer)
- various fixes to msgactions which were sometimes not showing when mouse hovering
- the picker is now correctly showing at the right position when opening it from the reactions list of a message
- emoji picker and msg actions are now all out of the live and can only have one instance at a time
- fixes the dreaded blank screen on iOS

### DIFF
--- a/assets/javascripts/discourse/components/chat-emoji-picker.js
+++ b/assets/javascripts/discourse/components/chat-emoji-picker.js
@@ -87,13 +87,13 @@ export default class ChatEmojiPicker extends Component {
   @action
   didPressEscape(event) {
     if (event.key === "Escape") {
-      this.chatComposerEmojiPickerManager.close();
+      this.chatEmojiPickerManager.close();
     }
   }
 
   @action
   didNavigateFitzpatrickScale(event) {
-    if (event.type !== "keydown") {
+    if (event.type !== "keyup") {
       return;
     }
 
@@ -128,7 +128,7 @@ export default class ChatEmojiPicker extends Component {
 
   @action
   didToggleFitzpatrickScale(event) {
-    if (event.type === "keydown") {
+    if (event.type === "keyup") {
       if (event.key === "Escape") {
         event.preventDefault();
         this.isExpandedFitzpatrickScale = false;
@@ -145,9 +145,10 @@ export default class ChatEmojiPicker extends Component {
 
   @action
   didRequestFitzpatrickScale(scale, event) {
-    if (event.type === "keydown") {
+    if (event.type === "keyup") {
       if (event.key === "Escape") {
         event.preventDefault();
+        event.stopPropagation();
         this.isExpandedFitzpatrickScale = false;
         this._focusCurrentFitzpatrickScale();
         return;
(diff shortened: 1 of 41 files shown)


jjaffeux:
Chat composer emoji picker:

Screenshot 2022-10-17 at 17 04 47

Message actions emoji picker:
Screenshot 2022-10-17 at 17 05 27

Mobile:
Screenshot 2022-10-17 at 17 05 52

assets/javascripts/discourse/services/emoji-picker-scroll-observer.js
```diff
@@ -0,0 +1,70 @@
...
+      this.direction = "down";
+    } else {
+      this.direction = "up";
+    }
+
+    this.prevYPosition = target.scrollTop;
+  }
+
+  _isSectionVisibleInPicker(section, picker) {
+    const { bottom, height, top } = section.getBoundingClientRect();
```
romanrizzi, line 63: Just curious, why not use an `IntersectionObserver`?

assets/javascripts/discourse/services/chat-emoji-picker-manager.js
```diff
@@ -0,0 +1,146 @@
...
+    this.callback = null;
+    this.closing = true;
+
+    later(() => {
+      if (this.isDestroyed || this.isDestroying) {
+        return;
+      }
+
+      this.visibleSections = ["favorites", "smileys_&_emotion"];
+      this.lastVisibleSection = "favorites";
```
romanrizzi, line 50: Looks like these two are defaults and also set on `closeExisting`. Should we move them to a constant?

assets/javascripts/discourse/components/chat-message-actions-desktop.js
```diff
@@ -1,11 +1,49 @@
...
+        {
+          placement: "right-start",
+          hide: { enabled: true },
+          modifiers: [
+            {
+              name: "offset",
+              options: {
+                offset: ({ popper, placement }) => {
+                  return [
+                    2,
```
romanrizzi, line 33: Can we move this into a variable to explain what it means?

assets/javascripts/discourse/services/emoji-picker-scroll-observer.js
jjaffeux: I tried, but it has all sorts of limits of not getting triggered when intersection already happened, here I need more precise control during the scroll. Might be wrong, but couldn't make it work as I wanted.

assets/javascripts/discourse/components/chat-emoji-picker.js
```diff
@@ -0,0 +1,339 @@
...
+
+  @action
+  didRequestSection(section) {
+    this.filteredEmojis = null;
+
+    // we disable scroll listener during requesting section
+    // to avoid it from detecting another section during scroll to requested section
+    this.emojiPickerScrollObserver.enabled = false;
+    this.chatEmojiPickerManager.addVisibleSections([section]);
+    this.chatEmojiPickerManager.lastVisibleSection = section;
```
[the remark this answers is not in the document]
jjaffeux: I think this is going to be tricky in some cases, probably better to keep fine control here

PR #1315, opened 2022-10-18
FIX: ensures emojis have a title in the picker

jjaffeux:


assets/javascripts/discourse/helpers/tonable-emoji-title.js
```diff
@@ -0,0 +1,9 @@
+import { registerUnbound } from "discourse-common/lib/helpers";
+
+registerUnbound("tonable-emoji-title", function (emoji, scale) {
```
martin-brennan, line 3: Minor nit but maybe this should be called `diversityScale` or something since we are passing `chatEmojiReactionStore.diversity` as the arg

martin-brennan (approved):
Nice 👍

assets/javascripts/discourse/helpers/tonable-emoji-title.js
jjaffeux: Yep agreed, quickly thought of this and should have done it :+1:

PR #1321, opened 2022-10-20
FIX: Remove hijack on webhook and improve documentation

martin-brennan:
This commit removes the `hijack` on the chat webhooks controller since it is of dubious necessity and because `hijack` is suppressing the correct resposne codes.

Also add some documentation around how the incoming webhooks work, their params, and their limits.

Bump the incoming text param length limit to 2000, 1000 is quite low.


lis2 (approved):

tgxworld:
> because hijack is suppressing the correct resposne codes.

Interesting we actually return 418 here. 

"I'm a teapot client error response code indicates that the server refuses to brew coffee because it is, permanently, a teapot."

PR #1324, opened 2022-10-21
DEV: Rename discourse-chat to chat

romanrizzi:


jjaffeux:
This one can be closed right @romanrizzi ?

PR #1327, opened 2022-10-25
FIX: prevents exception when transcripting multiple messages

jjaffeux:
Only the first message would get the date element and following `dateTimeLinkEl` would be undefined, resulting in an error when trying to assign `innerText`.


jjaffeux:
@romanrizzi I might write a test for this later, but I need to go and would like to get this merged as it's causing issues. Can you review/merge if possible please?

romanrizzi (approved):

PR #1329, opened 2022-10-25
DEV: Lint files to satisfy core standards.

romanrizzi:
Before moving this plugin into core, we must comply with the their linting rules and Ember code standards.

This change applies two Ember codemods to update hbs templates: `ember-no-implicit-this-codemod` and `ember-angle-brackets-codemod`.

### DIFF
--- a/.template-lintrc.js
+++ b/.template-lintrc.js
@@ -1,4 +1,8 @@
 module.exports = {
   plugins: ["ember-template-lint-plugin-discourse"],
   extends: "discourse:recommended",
+
+  rules: {
+    "no-capital-arguments": false, // TODO: we extensively use `args` argument name
+  }
 };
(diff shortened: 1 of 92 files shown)


.template-lintrc.js
```diff
@@ -1,4 +1,8 @@
 module.exports = {
   plugins: ["ember-template-lint-plugin-discourse"],
   extends: "discourse:recommended",
+
+  rules: {
+    "no-capital-arguments": false, // TODO: we extensively use `args` argument name
```
jjaffeux, line 6: what case do we have with capital arg?

jjaffeux:
I tried it locally and it seems to work just fine, let's just merge this.

.template-lintrc.js
romanrizzi: The plugin outlet component uses the `@args` attribute, and this rule complains about it. It's already disabled in core, so I'm doing the same here.

jjaffeux: hmmm, it's not supposed to be the case AFAIK, this rule is preventing this: `<Foo @Bar={1}>`

https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-capital-arguments.md

romanrizzi: It also complains about the `@args` attribute being restricted. See [here](https://github.com/discourse/discourse-chat/actions/runs/3323743368/jobs/5494479287#step:8:11)

PR #1333, opened 2022-10-26
FIX: improve chat emoji reactions

jjaffeux:
- now sort emojis by most used and recent
- ensures we track emojis in every cases where we use them
- adds tests

### DIFF
--- a/assets/javascripts/discourse/components/chat-message.js
+++ b/assets/javascripts/discourse/components/chat-message.js
@@ -52,6 +52,7 @@ export default Component.extend({
   chatMessageActionsMobileAnchor: null,
   chatMessageActionsDesktopAnchor: null,
   chatMessageEmojiPickerAnchor: null,
+  cachedFavoritesReactions: null,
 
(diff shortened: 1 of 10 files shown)


assets/javascripts/discourse/components/chat-message.js
```diff
@@ -563,6 +563,10 @@ export default Component.extend({
     this._loadingReactions.push(emoji);
     this._updateReactionsList(emoji, reactAction, this.currentUser);
 
+    if (reactAction === this.ADD_REACTION) {
+      this.emojiReactionStore.track(`:${emoji}:`);
```
romanrizzi, line 567: Should we rename the service here to match other usages?

romanrizzi (approved):

assets/javascripts/discourse/components/chat-message.js
jjaffeux: oh yes indeed forgot this one :+1: thanks

PR #1336, opened 2022-10-27
FIX: Move reaction emoji name from class to data attribute

davidtaylorhq:
Using it as a class can cause unexpected clashes with 'real' CSS classes. Using a data attribute is safer, and can still be targetted by CSS if desired.


davidtaylorhq (commented):
Test changes LGTM ✅ 

(but you'll have to actually approve it, since I created the PR)

jjaffeux:
> Test changes LGTM ✅
> 
> (but you'll have to actually approve it, since I created the PR)

https://www.youtube.com/watch?v=B4h__cWaWg8

PR #1339, opened 2022-10-27
UX: implements leave button on channel card

jjaffeux:


jjaffeux:
Screenshot 2022-10-27 at 17 14 43

assets/stylesheets/common/chat-browse.scss
```diff
@@ -91,4 +91,20 @@
       font-size: var(--font-up-1);
     }
   }
+
+  .chat-channel-card {
+    .chat-channel-card__leave-btn {
+      color: var(--tertiary-high);
```
chapoi, line 97: I think the colour can come from .btn-link

chapoi (approved):

assets/stylesheets/common/chat-browse.scss
jjaffeux: it was slightly different in my test than the one used for members count

chapoi: and you're sure that isnt due to visited state?

jjaffeux: yes, I might be wrong, but I will merge like this and let you tweak if you see something wrong

PR #1341, opened 2022-10-27
DEV: Move plugin to discourse core

romanrizzi:


jjaffeux:
I will approve this, but I guess we need to sync on this

jjaffeux (approved):

PR #1348, opened 2022-11-01
FIX: Add editing user ids to ChatMessage and ChatMessageRevision

martin-brennan:
This commit adds last_editor_id to ChatMessage for parity with Post in core, as well as adding user_id to the ChatMessageRevision record since we need to know who is making edits and revisions to messages, in case in future we want to allow more than just the current user to edit chat messages. The backfill for data here simply uses the record's creating user ID, but in future if we allow other people to edit the messages it will use their ID.


jjaffeux:
@martin-brennan Im wondering if we need to pass the guardian everywhere given we have `message.user` ?

jjaffeux:
@martin-brennan Please reopen on core repo, sorry for the inconvenience

